PR #41, opened 2021-03-15
Add support for plugin replacement

LukeTowers:


jaxwilko:
Added support for versioning, the now supported methods of replacing a plugin are:

```php
public function pluginDetails()
{
    return [
        'name'        => 'Jax Demo',
        'description' => 'Provides features used by the provided demonstration theme.',
        'author'      => 'Alexey Bobkov, Samuel Georges',
        'icon'        => 'icon-leaf',
        'replaces'    => [
            'name' => 'Winter.Demo',
            'version' => '~1.0'
        ]
    ];
}
 ```
or
```php
public function pluginDetails()
{
    return [
        'name'        => 'Jax Demo',
        'description' => 'Provides features used by the provided demonstration theme.',
        'author'      => 'Alexey Bobkov, Samuel Georges',
        'icon'        => 'icon-leaf',
        'replaces'    => 'Winter.Demo'
    ];
}
```

The version rule is passed through `Semvar::match($rule, $currentPluginVersion)` in `PluginBase::replacesVersion` which means it has support for the same rules as composer.

NOTICE: this can be merged without the PR to add the `Semvar` parser, however using a version'ed replace without that PR will throw an exception

bennothommo:
@jaxwilko I've noticed you've been using the `Semvar` term, wouldn't this be `Semver` (as in **sem**antic **ver**sioning)?

jaxwilko:
@bennothommo yes, that would make a lot more sense 😂. I'll go and rename everything and push updates in the morning. Thanks for pointing it out :)

bennothommo:
@jaxwilko All good :)

LukeTowers:
@jaxwilko I think we should probably be a lot more strict in how the replaces property works:

```php
'replace' => [
    'Original.Plugin' => '1.0.4', // dynamic version ranges are not supported, see https://stackoverflow.com/a/18905069/6652884
    'Other.Plugin' => 'self.version', // support self.version like composer does
],
```

This keeps it more in line with how the composer `replace` property works (with the exception of explicitly denying version ranges since they don't really make sense in this property and forces people to be more strict with what exactly they're advertising drop in replacements for.

jaxwilko:
> @jaxwilko I think we should probably be a lot more strict in how the replaces property works:
> 
> ```
> 'replace' => [
>     'Original.Plugin' => '1.0.4', // dynamic version ranges are not supported, see https://stackoverflow.com/a/18905069/6652884
>     'Other.Plugin' => 'self.version', // support self.version like composer does
> ],
> ```
> 
> This keeps it more in line with how the composer `replace` property works (with the exception of explicitly denying version ranges since they don't really make sense in this property and forces people to be more strict with what exactly they're advertising drop in replacements for.

@LukeTowers you can now supply the replace array in your suggested style, this means that both:

```php
return [
    'name'        => 'Jax Demo',
    'description' => 'Provides features used by the provided demonstration theme.',
    'author'      => 'Alexey Bobkov, Samuel Georges',
    'icon'        => 'icon-leaf',
    'replaces'    => 'Winter.Demo'
];
```
and 
```php
return [
    'name'        => 'Jax Demo',
    'description' => 'Provides features used by the provided demonstration theme.',
    'author'      => 'Alexey Bobkov, Samuel Georges',
    'icon'        => 'icon-leaf',
    'replaces'    => [
        'Winter.Demo' => 'self.version'
    ]
];
```
are valid.

You can also now provide multiple replacements by adding additional key value pairs to the array.

RE: self.version, there is a catch in the plugin to swap that out for the plugin's version.

Version comparison is now handled by composer/semver using `Semver::satisfies($version, $constraints)`.

modules/system/classes/PluginBase.php
```diff
@@ -321,4 +328,104 @@ protected function getConfigurationFromYaml($exceptionMessage = null)
...
+    /**
+     * Passes the check through to composer/semver but allows Winter to perform
+     * checks / changes as required
+     *
+     * @return bool
+     */
+    protected function replacesVersion(string $version, string $constraints): bool
+    {
+        if ($constraints === 'self.version') {
+            $constraints = $this->getVersion();
```
LukeTowers: Shouldn't `self.version` refer to the last advertised version available in version.yaml rather than the current version of the plugin present in the migrations table?

bennothommo:
@jaxwilko @LukeTowers I'm not sure about plugins being able to replace multiple plugins - seems like a whole lot of complexity for not enough gain. I think the vast majority of cases will be a plugin replacing another that is superceded or continuing from an abandoned plugin.

jaxwilko:
@bennothommo the only case I can think of that would be useful is if somebody did something like merge RainLab.User and RainLab.Blog into one to make heavy customisations which this would allow for

LukeTowers:
@bennothommo @jaxwilko the use case I was thinking of was replacing plugins that replace other plugins, so something like RainLab.Users and Lovata.Buddies, but after further thought and factoring in the namespace replacement we could probably be fine with just saying that the use case of forking is the main point of it. Still need to support both specific versions and self.version though, shouldn't be version ranges at all.

LukeTowers:
@jaxwilko @mjauvin @bennothommo what's the current status of this?

modules/system/classes/PluginManager.php
```diff
@@ -680,11 +680,13 @@ public function registerReplacedPlugins()
             if (!isset($this->plugins[$target])) {
                 // register lang namespace alias for bc
                 Lang::registerNamespaceAlias($replacement, $target);
+                Config::getLoader()->registerNamespaceAlias($replacement, $target);
                 continue;
             }
 
             if ($this->plugins[$replacement]->replaces($target, $this->plugins[$target]->getVersion())) {
                 Lang::registerNamespaceAlias($replacement, $target);
+                Config::getLoader()->registerNamespaceAlias($replacement, $target);
```
LukeTowers: @jaxwilko is there a reason why we want to have config namespace aliasing registered in two separate locations?

LukeTowers: I'm more so talking about the two different registerNamespaceAlias methods in https://github.com/wintercms/storm/pull/19

LukeTowers:
@jaxwilko any idea what's up with Error: Class 'Composer\Semver\Semver' not found?

jaxwilko:
@LukeTowers That was added in the system submodule, you may need to require it manually untill it gets installed naturally

https://github.com/wintercms/winter/blob/bd5a83cd85a22d56eab1d0e4aa02055c677eeb05/modules/system/composer.json

LukeTowers:
Waiting for the tests to get fixed and more tests to be added to cover the full range of functionality introduced by these changes.

modules/system/controllers/Updates.php
```diff
@@ -185,19 +184,17 @@ protected function getPluginVersionFile($path, $filename)
...
                 if (!is_array($details)) {
                     $details = (array)$details;
                 }
 
                 //Filter out update scripts
-                $details = array_values(array_filter($details, function ($string) use ($path) {
+                $details = array_filter($details, function ($string) use ($path) {
                     return !preg_match('/^[a-z_\-0-9]*\.php$/i', $string) || !File::exists($path . '/updates/' . $string);
-                }));
+                });
```
LukeTowers, line 197: @jaxwilko @mjauvin what was the reason for this change?

tests/unit/system/classes/UpdatesControllerTest.php
```diff
@@ -14,20 +14,12 @@ public function testGetPluginVersionFile()
...
         $expectedVersions = [
-            '1.3.2' => [
-                'Added support for Translate plugin. Added some new languages.',
-            ],
-            '1.3.1' => [
-                'Minor bug fix Please see changelog',
-            ],
-            '1.3.0' => [
-                '!!! We\'ve refactored major parts of this plugin. Please see the website for more information.',
-            ],
```
LukeTowers, line 25: Why were these test cases removed?

LukeTowers:
@jaxwilko @mjauvin in addition to my comments above, could one of you submit a PR to the docs documenting this new feature; as well as the newly added aliasing in Lang, Config, NavigationManager, & SettingsManager?

modules/system/controllers/Updates.php
jaxwilko: We somehow got the Yaml preprocessor changes in the branch which was causing us all sorts of issues so I cherry picked the commit reverting it across:

`git cherry-pick 0ebe3bc07416e582c40e0e7c4f5737a08329db60`

tests/unit/system/classes/UpdatesControllerTest.php
jaxwilko: Same reason with the as above

jaxwilko:
@LukeTowers I can add the docs this afternoon :)

tests/unit/system/classes/UpdatesControllerTest.php
bennothommo: @jaxwilko is correct - that test was part of the Yaml upgrade which we didn't proceed with.

LukeTowers:
Congrats everyone! Great work from @jaxwilko, @mjauvin, & @bennothommo! 🎊

PR #56, opened 2021-03-21
Update lang.php - small corrections

najbo:
Small corrections in french translations.

### DIFF
--- a/modules/backend/lang/fr/lang.php
+++ b/modules/backend/lang/fr/lang.php
@@ -3,13 +3,13 @@
 return [
     'auth' => [
         'title' => 'Zone d’administration',
-        'invalid_login' => 'L\'utilisateur saisie ne correspond à aucun utilisateur enregistré. Merci de vérifier votre saisie et de réessayer.'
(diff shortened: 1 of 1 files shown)


modules/backend/lang/fr/lang.php
```diff
@@ -246,7 +246,7 @@
         'saving_name' => 'Enregistrement d\'un(e) :name en cours…',
         'delete' => 'Supprimer',
         'deleting' => 'Suppression en cours…',
-        'confirm_delete' => 'Supprimer cet enregistrement?',
+        'confirm_delete' => 'Supprimer cet enregistrement ?',
```
LukeTowers, line 249: Why was the extra space added here?

najbo: Hi Luke,

In french, we put a small space before the ? ! ; (and sometimes before the : )
To be correct, and to avoid that the ? goes on a new line, the space caracter should be replaced by a &nbsp;
I think it's more easy to put a normal space here.

Some informations here (only in french) : 

- https://fr.wikipedia.org/wiki/Espace_fine_ins%C3%A9cable#:~:text=Le%20code%20typographique%20fran%C3%A7ais%20recommande,du%20texte%20mis%20en%20exergue.
- https://www.typofute.com/l_espace_fine_insecable_dans_les_documents_html

PR #65, opened 2021-03-25
Add version.yaml processing for Symfony/Yaml v4 support

### LINKED ISSUE #13: Support newer versions of symfony/yaml
We should add a preprocessor to YAML files to prevent the breaking changes in newer versions of symfony/yaml from affecting the ecosystem (i.e. unquoted string keys that could be interpreted as numbers no longer work in newer versions).

Refs: octobercms/october#5271

bennothommo:
Fixes https://github.com/wintercms/winter/issues/13
Follows on from https://github.com/wintercms/storm/pull/16

Moves the `version.yaml` processing into the System module and updates all version file parsing to use this new processor.

### DIFF
--- a/modules/system/classes/VersionYamlProcessor.php
+++ b/modules/system/classes/VersionYamlProcessor.php
@@ -26,7 +26,7 @@ class VersionYamlProcessor implements YamlProcessor
                 return '"' . trim($matches[1]) . '":';
             }, rtrim($line));
 
-            // Add quotes around any unquoted text proceeding an array key
+            // Add quotes around any unquoted text preceeding an array key
             $line = preg_replace('/^\s*([^\n\r\-:]+)\s*: +(?![\'"\s])(.*)/m', '$1: "$2"', $line);
 
             // If this line is the continuance of a multi-line string, remove the quote from the previous line and
--- a/modules/system/classes/VersionYamlProcessor.php
+++ b/modules/system/classes/VersionYamlProcessor.php
@@ -1,0 +1,56 @@
+<?php namespace System\Classes;
+
+use Winter\Storm\Parse\Processor\Contracts\YamlProcessor;
+
+/**
+ * "version.yaml" pre-processor class.
+ *
+ * Post-v3.x versions of the Symonfy/Yaml package use more recent versions of YAML spec, which breaks common
+ * implementations of our version file format. To maintain compatibility, this class will pre-process YAML
+ * contents from these files to work with Symfony/Yaml 4.0+.
+ *
+ * @author Winter CMS
+ */
+class VersionYamlProcessor implements YamlProcessor
+{
+    /**
+     * @inheritDoc
+     */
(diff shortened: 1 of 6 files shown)


LukeTowers:
As discussed on Discord we should probably actually make a single processor to deal with all of the incompatibilities, and apply it by default in the Storm library.

bennothommo:
Follow up PR here: https://github.com/wintercms/storm/pull/18

modules/system/classes/VersionYamlProcessor.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $lines = preg_split('/[\n\r]+/', $text, -1, PREG_SPLIT_NO_EMPTY);
+
+        foreach ($lines as $num => &$line) {
+            // Surround array keys with quotes if not already
+            $line = preg_replace_callback('/^\s*([\'"]{0}[^\'"\n\r:]+[\'"]{0})\s*:/m', function ($matches) {
+                return '"' . trim($matches[1]) . '":';
+            }, rtrim($line));
+
+            // Add quotes around any unquoted text proceeding an array key
```
mjauvin:
```suggestion
            // Add quotes around any unquoted text preceeding an array key
```

LukeTowers:
Is this PR still required with wintercms/storm#18 in play?

mjauvin:
Pretty sure it is still required @LukeTowers .

It's doing extra stuff special to `version.yaml`

LukeTowers:
What is the extra stuff exactly?

mjauvin:
https://github.com/wintercms/winter/blob/b736df2b88d92f669c7c8b193b8a9d5781f803a7/modules/system/classes/VersionYamlProcessor.php#L29,L43

bennothommo:
@LukeTowers as @mjauvin said, it is still required. Our usage of `!!!` to mark important updates also fails the YAML spec (in the way that we use it) . I didn't worry about it for the global parser because not only was it unlikely to be in use, if it was in use as a parsed value like the spec says, it would be intentional.

modules/system/classes/VersionYamlProcessor.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $lines = preg_split('/[\n\r]+/', $text, -1, PREG_SPLIT_NO_EMPTY);
+
+        foreach ($lines as $num => &$line) {
+            // Surround array keys with quotes if not already
+            $line = preg_replace_callback('/^\s*([\'"]{0}[^\'"\n\r:]+[\'"]{0})\s*:/m', function ($matches) {
+                return '"' . trim($matches[1]) . '":';
+            }, rtrim($line));
+
+            // Add quotes around any unquoted text preceeding an array key
```
LukeTowers:
```suggestion
            // Add quotes around any unquoted text following an array key
            // specifically to ensure usage of !!! in unquoted comments does not fail
```

PR #67, opened 2021-04-04
Mirror sqlite database file

mjauvin:
- added "storage/database.sqlite" to $files array
- mirror file section at the end so that sub-folders have been created at this point

### DIFF
--- a/modules/system/console/WinterMirror.php
+++ b/modules/system/console/WinterMirror.php
@@ -36,6 +36,7 @@ class WinterMirror extends Command
         'robots.txt',
         'humans.txt',
         'sitemap.xml',
+        'storage/database.sqlite',
     ];
 
     protected $directories = [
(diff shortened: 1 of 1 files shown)


bennothommo:
@mjauvin What's the rationale for making the `storage/database.sqlite` file available in the public directory? That would be a huge security issue, I would think?

mjauvin:
@bennothommo do you have another solution to make this work when using a public (winter:mirror) docroot?

If the site is using sqlite, it won't work unless the sqlite file is under public.

bennothommo:
@mjauvin I don't understand the root issue you're trying to solve. Would you mind explaining the issue in more detail?

There is absolutely no reason for the `database.sqlite` file to be publicly accessible. That's essentially allowing anyone to simply download a copy of your database and be able to access all sorts of critical information.

mjauvin:
@bennothommo the file won't be publicly accessible because of the .htaccess file rules, but it is needed by winter cms to render the website (frontend and backend) if your website is using sqlite db backend.

bennothommo:
@mjauvin alright, I've managed to replicate the issue - the issue is that, by default, the `DB_DATABASE` environment variable (or the default value for `database.connnections.sqlite.database`) use a relative path to the database file. With the storage directory being mirrored, this makes the `public/storage` directory the most direct relative path, so it expects the `database.sqlite` file to be in there.

This is a configuration issue though - easily resolved by simply using an absolute path to the database file.

mjauvin:
I confirm this, setting DB_DATABASE to the absolute path does the trick!

LukeTowers:
@mjauvin does https://github.com/wintercms/winter/commit/0061b09246c2ed2a94cf56a81842cd35180ee4e1 solve the problem better by default?

mjauvin:
@LukeTowers yes, that resolves the issue.

Thanks.

PR #70, opened 2021-04-08
trigger a new ajaxRedirect event and handle it in the loading indicator

mjauvin:
Cleaner way to disable loading indicator when an AJAX handler returns a redirect reponse.

ref. https://github.com/octobercms/october/commit/99a08616bb649af7bf442fa1d01b7ce70dc6a939
ref. https://github.com/octobercms/october/issues/5555
ref. https://github.com/octobercms/october/issues/5055

replaces https://github.com/octobercms/october/pull/5321

### DIFF
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -292,7 +292,7 @@ if (window.jQuery.request !== undefined) {
                 // so that the loading indicator for redirects that cause a browser to download 
                 // a file instead of leave the page will properly stop.
                 // @see https://github.com/octobercms/october/issues/5055
-                $el.trigger('ajaxDone')
+                $el.trigger('ajaxRedirect')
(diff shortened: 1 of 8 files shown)


modules/system/assets/js/framework.js
```diff
@@ -292,7 +292,7 @@ if (window.jQuery.request !== undefined) {
                 // so that the loading indicator for redirects that cause a browser to download 
                 // a file instead of leave the page will properly stop.
                 // @see https://github.com/octobercms/october/issues/5055
-                $el.trigger('ajaxDone')
+                $el.trigger('ajaxRedirect')
```
LukeTowers: Could we rename this to ajaxRedirected instead?

mjauvin: done

PR #78, opened 2021-04-16
Replaced hard-coded colors with their corresponding variables names

HitArrowLegend:


bennothommo:
@HitArrowLegend is this work completed on your end?

HitArrowLegend:
there is one thing I need to modify. I'll commit it now

HitArrowLegend:
when these tests are done, this should be good to go

PR #109, opened 2021-04-23
[WIP] Replacement warnings

jaxwilko:
This PR adds warnings to both the backend UI and the CLI to inform the user to remove replaced plugins.

Backend Example:
Backend UI

CLI Example:
CLI


modules/system/lang/en/lang.php
```diff
@@ -336,6 +336,8 @@
         'core_set_build' => 'Setting build number',
         'update_warnings_title' => 'Some issues have been detected and require attention:',
         'update_warnings_plugin_missing' => 'The :parent_code plugin requires :code to be installed before it will work',
+        'update_warnings_plugin_replace' => 'The :plugin plugin replaces :alias, please remove :alias to ensure the are no conflicts',
```
mjauvin:
```suggestion
        'update_warnings_plugin_replace' => 'The :plugin plugin replaces :alias; please remove :alias to ensure there are no conflicts',
```

modules/system/lang/en/lang.php
```diff
@@ -336,6 +336,8 @@
...
+        'update_warnings_plugin_replace_cli' => 'This plugin replaces :alias, please remove :alias to ensure the are no conflicts',
```
mjauvin:
```suggestion
        'update_warnings_plugin_replace_cli' => 'This plugin replaces :alias; please remove :alias to ensure there are no conflicts',
```

mjauvin:
Since both messages are so identical, should we only have one? Simpler to maintain/translate.

jaxwilko:
@mjauvin They're not quite the same as the start different:

`This plugin replaces :alias` (matches the format for the cli update where each class reports back info)
vs
`The :plugin plugin replaces :alias` (matches format of the backend ui)

mjauvin:
I understand Jack, but the message for the UI could be used for both.

mjauvin:
Can we merge this?

PR #114, opened 2021-04-25
Fix to uninstall to include disabled plugins

jaxwilko:
Disabled plugins are not returned by `getPlugins()`, this means that when `winter:down` is ran if plugins are disabled then their tables remain in the DB.

Steps to reporduce are:

- `./artisan winter:up`
- disable any plugin with tables
- `./artisan winter:down`

Result: the disabled plugin's tables are left over.

This patch switches `getPlugins()` for `getAllPlugins()` which includes disabled plugins.

### DIFF
--- a/modules/system/classes/UpdateManager.php
+++ b/modules/system/classes/UpdateManager.php
@@ -347,7 +347,7 @@ class UpdateManager
         /*
          * Rollback plugins
          */
-        $plugins = array_reverse($this->pluginManager->getPlugins());
+        $plugins = array_reverse($this->pluginManager->getAllPlugins());
         foreach ($plugins as $name => $plugin) {
             $this->rollbackPlugin($name);
         }


mjauvin:
That actually work @jaxwilko ?

jaxwilko:
@mjauvin yeah, it's not the fix for replaced plugins, that's something else I'm still working on, this is a fix for a bug I ran into while working on that :).

I have tested it a few different ways and it seems to work fine, very welcome to test it under different conditions though 👍

bennothommo:
@jaxwilko Does this cause any problems if a plugin has been disabled before it had run any migrations? (ie. through the `cms.disablePlugins` config)

jaxwilko:
@bennothommo shouldn't do as it'll just rollback through the plugins history, so if there have been no migrations ran for plugin it won't try and rollback anything for the plugin

PR #115, opened 2021-04-25
fix plugin replace history

mjauvin:
- add replacement plugin history
- remove replaced plugin history

this allow tables to be properly purged when replacement plugin is removed

### DIFF
--- a/modules/system/classes/VersionManager.php
+++ b/modules/system/classes/VersionManager.php
@@ -115,6 +115,33 @@ class VersionManager
             return;
         }
 
+        $code = $plugin->getPluginIdentifier();
+
+        // add history up to $currentVersion
+        if ($versions = $this->getOldFileVersions($code, $currentVersion)) {
+            foreach ($versions as $version => $details) {
+                list($comments, $scripts) = $this->extractScriptsAndComments($details);
+
+                foreach ($scripts as $script) {
+                    Db::table('system_plugin_history')->insert([
+                        'code'       => $code,
+                        'type'       => self::HISTORY_TYPE_SCRIPT,
+                        'version'    => $version,
+                        'detail'     => $script,
+                        'created_at' => new Carbon
+                    ]);
+                }
+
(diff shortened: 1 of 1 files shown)


modules/system/classes/VersionManager.php
```diff
@@ -115,6 +115,33 @@ public function replacePlugin(PluginBase $plugin, string $replace)
...
+            }
+        }
+
+        // delete replaced plugin history
+        Db::table('system_plugin_history')->where('code', $replace)->delete();
+
+        // replace installed version
         Db::table('system_plugin_versions')
             ->where('code', '=', $replace)
             ->update([
```
jaxwilko, line 148: Can't leave a comment where I wanted to, but the line below (line: 148) can use `$code` as well rather than calling the `getPluginIdentifier()` function again

modules/system/classes/VersionManager.php
```diff
@@ -115,6 +115,33 @@ public function replacePlugin(PluginBase $plugin, string $replace)
...
+
+                foreach ($scripts as $script) {
+                    Db::table('system_plugin_history')->insert([
+                        'code'       => $code,
+                        'type'       => self::HISTORY_TYPE_SCRIPT,
+                        'version'    => $version,
+                        'detail'     => $script,
+                        'created_at' => new Carbon
+                    ]);
+                }
```
LukeTowers:
```suggestion
                $now = now()->toDateTimeString();

                foreach ($scripts as $script) {
                    Db::table('system_plugin_history')->insert([
                        'code'       => $code,
                        'type'       => self::HISTORY_TYPE_SCRIPT,
                        'version'    => $version,
                        'detail'     => $script,
                        'created_at' => $now,
                    ]);
                }
```

modules/system/classes/VersionManager.php
mjauvin: yes, thanks!

PR #121, opened 2021-04-26
Added fix to ensure removed plugins show no warnings

### LINKED ISSUE #118: Replaced plugin message
I installed Winter 1.1.3
Then the winter.user installed the plugin:
```
composer require winter/wn-user-plugin
```

Now on the page `/backend/system/updates` there is a message that you need to remove the rainlab.user plugin.

I did not install the rainlab.user plugin.

jaxwilko:
Because the `PluginManager`'s `replacementMap` isn't cleared when an original plugin is not installed the warnings were still being shown.

This patch checks if the original plugin has been removed before displaying warnings.

### DIFF
--- a/modules/system/classes/UpdateManager.php
+++ b/modules/system/classes/UpdateManager.php
@@ -186,9 +186,11 @@ class UpdateManager
 
         // Set replacement warning messages
         foreach ($this->pluginManager->getReplacementMap() as $alias => $plugin) {
-            $this->addMessage($plugin, Lang::get('system::lang.updates.update_warnings_plugin_replace_cli', [
+            if (File::exists(PluginManager::instance()->getPluginPath($alias))) {
+                $this->addMessage($plugin, Lang::get('system::lang.updates.update_warnings_plugin_replace_cli', [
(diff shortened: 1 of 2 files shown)


bennothommo:
@jaxwilko Unless I'm mistaken, this is not going to work - `PluginManager::instance()->getPluginPath($alias)` will return `null` if the plugin has not been loaded, which would be the case here as the plugin does not exist. The following line defines the path mappings which the `getPluginPath` method uses: https://github.com/wintercms/winter/blob/develop/modules/system/classes/PluginManager.php#L175

jaxwilko:
@bennothommo yeah so if the plugin hasn't been loaded it wont be in the `pathMap` which means `getPluginPath()` returns null, null passed to `File::exists()` returns false.

It'll only return true if the original plugin is still in the filesystem, which is the case where we want to display the message

bennothommo:
@jaxwilko Does it return `null` if the plugin is disabled?

jaxwilko:
@bennothommo no it still returns the path even if the plugin is disabled.

image

After the disabled flag is set the `pathMap` is still registered

PR #128, opened 2021-04-29
Updated Latvian lang.php (CMS file)

arvislacis:
Updated Latvian translation file of CMS module:

- Minor fixes and changes in previous translations;
- Updated file with missing translation strings from en/lang.php.


mjauvin:
Thanks, we'll merge soon.

Issue #133, opened 2021-04-30
Image Resizer - Resizing fails for images with a file name extension not all in lowercase


helmutkaufmann:
- Winter CMS Build: 1.1.3
- PHP Version: 7.4

### Description:

When resizing an image, modules/system/classes/ImageResizer.php tests if the file to be resized has en image extension, such as jpg. The function fails if the image's extension is not all lowercase (eg JPG or Jpg). Extension should be converted to all lowercase before testing.

### Steps To Reproduce:
Name a file xyzzy.JPG and send it through the resize.

tschallacka:
Oh yes, this annoyed the hell out of me for years, when trying to upload a windows snip, as thats saved in uppercase filename. Wholehartedly agree.

helmutkaufmann:
Changed code. PR done - I hope I did it correctly, not a git person.

arvislacis:
@helmutkaufmann [PR and fix](https://github.com/wintercms/winter/pull/135) looks good to me. :)

PR #135, opened 2021-04-30
Update ImageResizer.php

helmutkaufmann:
Added strtolower when checking image file extensions.


arvislacis:
Fixes https://github.com/wintercms/winter/issues/133

PR #140, opened 2021-05-01
Fixing styles for custom checkbox

WebVPF:
The custom checkbox has the wrong styles.
Now there is an empty space in front of the custom checkbox:

Instead of full transparency, there is a `visibility` property. Then the element is on the page, but it is not visually visible.
Due to the `height: 0` we remove the space that the element occupies.

### DIFF
--- a/modules/system/assets/ui/less/checkbox.less
+++ b/modules/system/assets/ui/less/checkbox.less
@@ -46,7 +46,9 @@
     input[type=radio],
     input[type=checkbox] {
         display: block;
-        opacity: 0;
+        visibility: hidden;
+        height: 0;
+        margin-top: 0;
     }
 
     label {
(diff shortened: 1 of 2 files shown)


LukeTowers:
@WebVPF the problem with `height: 0` is that it prevents checkboxes from properly having browser validation messages display on them. If you have a form where you make a checkbox, checkboxlist, or radio element required, and then add `data-browser-validate` to the button that submits the form via AJAX, you'll see the difference with what happens to the browser's built in validation popup. 

Do you have any ideas how we can resolve both issues at the same time?

WebVPF:
@LukeTowers problem due to `height: 0`
or because of `display: none` ?

I watched this pull request https://github.com/octobercms/october/pull/5520

LukeTowers:
@WebVPF good point, I believe it was because of display: none. Should be able to merge this.

PR #142, opened 2021-05-01
Improved Latvian translation

arvislacis:
More improvements in Latvian translation files:

- Minor fixes in previous `modules/cms/lang/lv/lang.php`;
- Translated `modules/system/lang/lv/client.php` and re-generated file for client side - `modules/system/assets/js/lang/lang.lv.js`
- Over-viewed and updates `modules/system/lang/lv/validation.php ` and `modules/system/lang/lv/lang.php`.

Although specified changes from my side are reviewed and everything is fine for the pull request, I want to review and update also the last part of Latvian translation for Winter CMS - `modules/backend/lang/lv/lang.php` - that's the reason why currently this pull request is in WIP state.

If anyone has comments then they are welcome.

Also if Winter CMS maintainers are ok that I made separate pull request for `modules/backend/lang/lv/lang.php` then the current (this) pull request can be merged without worries. :)


arvislacis:
This pull request is finished from my side now. :)

modules/backend/lang/lv/lang.php
```diff
@@ -191,272 +287,365 @@
...
+            'iso_8859_8' => 'ISO-8859-8 (Latīņu, Ebreju)',
+            'iso_8859_0' => 'ISO-8859-9 (Latīņu-5, Turku)',
+            'iso_8859_10' => 'ISO-8859-10 (Latīņu-6, Ziemeļvalstu)',
+            'iso_8859_11' => 'ISO-8859-11 (Latīņu, Taju)',
+            'iso_8859_13' => 'ISO-8859-13 (Latīņu-7, Baltijas reģiona)',
+            'iso_8859_14' => 'ISO-8859-14 (Latīņu-8, Ķeltu)',
+            'iso_8859_15' => 'ISO-8859-15 (Latīņu-9, Rietumeiropas revīzija ar eiro zīmi)',
+            'windows_1250' => 'Windows-1250 (CP1250, Centrālās un Austrumeiropas)',
+            'windows_1251' => 'Windows-1251 (CP1251)',
+            'windows_1252' => 'Windows-1252 (CP1252)',
```
LukeTowers, line 572: Are these encoding names typically translated in other references or would end users be used to seeing them in English?

modules/backend/lang/lv/lang.php
```diff
@@ -191,272 +287,365 @@
...
-        'remove' => "Noņemt",
-        'remove_name' => "Noņemt :name",
-        'delete' => "Dzēst",
-        'delete_name' => "Dzēst :name",
-        'delete_confirm' => "Vai esat pārliecināts?",
-        'link' => "Saistīt",
-        'link_name' => "Saistīt :name",
-        'unlink' => "Atsaistīt",
-        'unlink_name' => "Atsaistīt :name",
-        'unlink_confirm' => "Vai esat pārliecināts?",
```
LukeTowers, line 245: Changes that are just swapping quote types around shouldn't be included in this PR, if we decide to normalize them at some point we'll do that for all the files in a single commit to keep the history clean. If you could revert the changes that are only swapping quote types that would be great. Thanks for all your hard work on this!

LukeTowers (changes requested):

modules/backend/lang/lv/lang.php
arvislacis: Yes, this kind of encoding translations can be found in other software (for example, MS Word etc.). I think in general most people still refers to encoding code - ISO-8859-13, Windows-1250 etc.  - and the description part is used mostly as the last option to decide, also, in my opinion, English and Latvian versions doesn't differ so much so users would be able to backtrack translation to English if needed.

modules/backend/lang/lv/lang.php
arvislacis: I agree that quote swapping may be bad for the git history but one of the goals of these Latvian translation updates were also to normalize and unify translations against the English versions. So all quote changes and extra commas for the last array values etc. were inserted on purpose to unify translations with English ones, to reduce potential issues and maintenance work in future. Also the mentioned `modules/backend/lang/lv/lang.php` file has many indention changes before character `=>` because the English version have introduced new parameters/values which didn't existed in Latvian version.

I could try to revert changes but in overall that would make 2-3x more effort:
1) because I would need to compare changed string with previous version (overall I am ok with this);
2) as I mentioned, problem is not only with quotes but also with commas and spaces before `=>` - to sum it up, Latvian translations were just too outdated and didn't have many changes made in English translations. It also means that before there wasn't no normalization process for other languages  when new strings and quote changes were introduced in English files. My changes were not made to complicate or ruin history but to normalize things.
3) If now I revert quote, space, comma changes then in future someone would need to do additional extra work to normalize things back again..., and it also will clutter history.

In my opinion we just accept that translations were too outdated and many changes were needed so in the future there would be less issues when adding new translations or when big changes are made to English language files. Also my proposed changes and versions are easier to use for future translators because if both English and Latvian version contains the same quote style and indentation then it's easier to see what strings has been added or removed from English files.

Short version - I am waiting for additional feedback on this before reverting something. And if we are starting to revert something then it applies only to quotes or also to commas (for the last array values) and spaces before `=>`?

LukeTowers: Fair enough!

LukeTowers (approved):

PR #147, opened 2021-05-04
Maintenance Allowed IP list

jaxwilko:
This PR adds support for whitelisting IP addresses to be allowed to visit the frontend while in maintenance mode. This is really useful when you have a site in maintenance mode to avoid being indexed by crawlers or for staging sites where automated tools need to be ran.

Backend Form

Uses `Request::ip()` under the hood for IP detection.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/147"></a>


modules/cms/models/maintenancesetting/fields.yaml
```diff
@@ -15,4 +15,21 @@ fields:
...
+        form:
+            fields:
+
+                ip:
+                    label: cms::lang.maintenance.ip_whitelist.ip
+                    type: text
+                    span: left
+
+                label:
+                    label: cms::lang.maintenance.ip_whitelist.ip
```
matteotrubini: `cms::lang.maintenance.ip_whitelist.label` ?

jaxwilko: Nice spot, totally missed that :+1:

modules/cms/models/MaintenanceSetting.php
```diff
@@ -89,4 +89,15 @@ public function afterFetch()
             $this->is_enabled = false;
         }
     }
+
+    /**
+     * Returns bool on if IP is in the whitelist
+     * @return bool
+     */
```
LukeTowers:
```suggestion
    /**
     * Check if the provided IP is in the whitelist
     *
     * @param string $ip
     * @return bool
     */
```

modules/cms/models/MaintenanceSetting.php
```diff
@@ -89,4 +89,17 @@ public function afterFetch()
...
+     * Check if the provided IP is in the whitelist
+     *
+     * @param string $ip
+     * @return bool
+     */
+    public static function inWhitelist(string $ip): bool
+    {
+        return in_array($ip, array_map(function ($arr) {
+            return $arr['ip'];
+        }, static::get('ip_whitelist', [])));
```
LukeTowers:
```suggestion
        return \Symfony\Component\HttpFoundation\IpUtils::checkIp($ip, static::get('ip_whitelist', []))
```

modules/cms/models/MaintenanceSetting.php
```diff
@@ -3,6 +3,8 @@
 use Model;
 use Cms\Classes\Page;
 use Cms\Classes\Theme;
+use Illuminate\Support\Arr;
```
LukeTowers: Doesn't October have it's own version?

mjauvin: Yep:
```
<?php namespace Winter\Storm\Support;

use Illuminate\Support\Arr as ArrHelper;

/**
 * Array helper
 *
 * @author Alexey Bobkov, Samuel Georges
 */
class Arr extends ArrHelper
```

jaxwilko: Yeah my bad, it got auto completed and I didn't check the import. Updated to Winter's version :)

damsfx:
Nice!
I Always use this option with a little bit of code in themes to show "pre openning" or waiting pages to visitors, and at the same time let backend users to use the website in real conditions.

modules/cms/lang/en/lang.php
```diff
@@ -100,6 +100,11 @@
         'is_enabled' => 'Enable maintenance mode',
         'is_enabled_comment' => 'Select the page to show when maintenance mode is activated.',
         'hint' => 'Maintenance mode will display the maintenance page to visitors who are not signed in to the back-end area.',
+        'ip_whitelist' => [
+            'name' => 'IP Whitelist',
+            'ip' => 'IP Address',
+            'label' => 'Rule Label'
```
LukeTowers:
```suggestion
            'label' => 'Description'
```

LukeTowers:
Minor point, but as developers in general we should probably move away from white/black list terminology towards include/exclude instead as it's more clear to newcomers and less controversial. I would name the field "Allowed IP addresses"  with the visual comment "IP addresses that are permitted to view the site while maintenance mode is active".

WebVPF:
How about adding the line "Your ip ..." before the list of allowed ip addresses, where to display the user's current ip address and the button "Add ip to the allowed list" if the ip is not in the allowed list?

And is it possible to do this functionality the same way to access the backend? Store only an array with allowed IP addresses in the `config/cms.php` file.

LukeTowers:
> And is it possible to do this functionality the same way to access the backend? Store only an array with allowed IP addresses in the config/cms.php file.

@WebVPF that could easily be achieved via a plugin

modules/cms/models/MaintenanceSetting.php
```diff
@@ -89,4 +91,15 @@ public function afterFetch()
...
         }
     }
+
+    /**
+     * Check if the provided IP is in the whitelist
+     *
+     * @param string $ip
+     * @return bool
+     */
+    public static function inAllowedIps(string $ip): bool
```
LukeTowers:
```suggestion
    public static function isAllowedIp(string $ip): bool
```

modules/cms/classes/Controller.php
```diff
@@ -163,6 +163,7 @@ public function run($url = '/')
         if (
             MaintenanceSetting::isConfigured() &&
             MaintenanceSetting::get('is_enabled', false) &&
+            !MaintenanceSetting::inWhitelist(Request::ip()) &&
```
```suggestion
            !MaintenanceSetting::isAllowedIp(Request::ip()) &&
```

LukeTowers (changes requested):

modules/cms/models/MaintenanceSetting.php
```diff
@@ -89,4 +91,15 @@ public function afterFetch()
             $this->is_enabled = false;
         }
     }
+
+    /**
+     * Check if the provided IP is in the whitelist
```
bennothommo:
```suggestion
     * Check if the provided IP is in the allowed IP list.
```

modules/cms/models/maintenancesetting/fields.yaml
```diff
@@ -15,4 +15,22 @@ fields:
...
+        cssClass: checkbox-align
+
+    allowed_ips:
+        label: cms::lang.maintenance.allowed_ips.name
+        type: repeater
+        commentAbove: cms::lang.maintenance.allowed_ips.description
+        cssClass: checkbox-align
+        form:
+            fields:
+
```
bennothommo:
```suggestion
```

modules/cms/models/maintenancesetting/fields.yaml
```diff
@@ -15,4 +15,22 @@ fields:
...
+                ip:
+                    label: cms::lang.maintenance.allowed_ips.ip
+                    type: text
+                    span: left
+
```
bennothommo:
```suggestion
```

bennothommo:
Just to note, as I'm having a look into it now, but this functionality does not support trusted proxies currently. It's using the proxy's IP in the lookup for allowed IPs.

LukeTowers (approved):

bennothommo:
My previous comment can be ignored. It seems to be something intrinsic to Gitpod in not forwarding on my actual IP - it doesn't happen on other proxies.

PR #45, opened 2021-03-17
Restore full Laravel Mailer behavior

mjauvin:
Related to https://github.com/wintercms/storm/pull/7

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/45"></a>

PR #148, opened 2021-05-04
Support Laravel 9 / PHP 8.1

LukeTowers:
Winter is well on its way to supporting Laravel 9! Thank you for your interest in providing testing for this work, please read the following information on how to get started testing it and report any issues that you encounter!

See https://github.com/wintercms/meta/blob/master/l9-upgrade-notes.md for testing instructions and the upgrade guide.

## Must Have

These issues are crucial for a smooth upgrade process and must be completed before 1.2 can go "stable":

### Internals:

- [x] https://github.com/wintercms/storm/pull/46
- [x] https://github.com/wintercms/winter/issues/578#issuecomment-1165505271
- [x] Revert: Default QueryBuilder methods are defined in the Builder class now preventing the user of `$builder->macro()` to override them in userspace, see https://github.com/laravel/framework/pull/37956#issuecomment-993822397 for details.
- [x] https://github.com/wintercms/storm/pull/64
- [x] https://github.com/wintercms/winter/issues/480
- [x] https://github.com/octobercms/library/commit/29ad91712e5d7ac6e2da70751ea22fe3f2ce9acf (evaluate)
- [x] https://github.com/wintercms/storm/pull/81
- [x] https://github.com/wintercms/winter/issues/503

### v1.1.9 Final changes:
- [x] https://github.com/wintercms/storm/pull/88

### Third Party / Dependencies:

- [x] [barryvdh/laravel-debugbar](https://github.com/barryvdh/laravel-debugbar/tree/master/src/Twig) doesn't support Twig v3.x yet, Winter.DebugBar should replace usage of Twig pieces to add support for v3 or a PR should be submitted upstream.

## Post "stable" release:

The following features were added in Laravel 7-9 and require extra work to "winterize" them for easier consumption / usage in Winter projects. These do not delay "stable" release but should be dealt with soon after:

### New Features:

- [ ] Maintenance mode functionality changed in Laravel 8, see https://github.com/laravel/framework/pull/33560. Explore combining backend maintenance mode setting with the framework maintenance mode, or better document the differences between them and what they're each best used for.
- [ ] Core support for CORS: https://laravel.com/docs/7.x/upgrade#cors-support
- [ ] Review Queue monitoring functionality
- [ ] Review RateLimiter functionality
- [ ] Review Prunable/MassPrunable model traits, especially as it relates to identifying models in the application
- [ ] Review Model Broadcasting
- [ ] Review ImmutableDates 
- [x] Finish implementing `create:model`
- [x] Finish implementing `create:migration`
- [x] Review Laravel commands to identify ones that can be marked as replaced in Winter for aliasing and any that need to be "Winterized".
- [x] Review the difference between Laravel registered default commands and Winter ones and decide which Laravel ones to add / modify for use in Winter.
- [ ] Symfony 5.4 added [support for autocompletion in console commands](https://symfony.com/blog/new-in-symfony-5-4-console-autocompletion), add support to all core commands and document it. Note: Currently waiting on documentation from Symfony
- [ ] Newer Laravel versions don't require manual registration of console commands, investigate how we can implement that for modules and plugins and get rid of our registration code for console commands: https://laravel.com/docs/9.x/artisan#registering-commands
- [ ] Opis/Closure also used in the ExtendableTrait functionality to support serializing classes with dynamic methods (Double check that the usage of the encryption key for serializable closures is being correctly applied)
- [ ] https://github.com/wintercms/winter/pull/45
- [x] https://github.com/wintercms/storm/pull/61
- [x] https://github.com/wintercms/winter/pull/481

### Maintenance:

- [ ] Investigate removing helpers that are now present by default in PHP 8.0
- [ ] https://github.com/wintercms/storm/pull/65
- [ ] Laravel 8.12 introduced support for encryptable casts (https://github.com/laravel/framework/pull/34937 & https://github.com/laravel/framework/commit/1cf55138ac28989a4a8f0f2d6d63784a54dd1061), review and decide how the Encryptable trait should work with this. Technically compatible as of changes in https://github.com/wintercms/storm/pull/26
- [ ] Review the Laravel 6 upgrade (https://github.com/octobercms/october/issues/4381 & https://github.com/octobercms/october/pull/4893) to ensure that nothing was overlooked in the previous upgrade.
- [ ] Review calls to $app->getNamespace() that assume a single application namespace (Usually \App) instead of a collection of modules & plugins that all form the namespace. This is typically used for autoloading files and cleaning up output to remove extra unnecessary paths but those tasks should be handled completely differently in Winter CMS.
- [ ] Laravel v7 introduced changes to getOriginal() in the Model class (HasAttributes trait). The Halcyon model base class does not implement this trait but instead implements similar methods. The Halcyon model base class should be refactored to make use of the HasAttributes trait from Eloquent as much as possible and only provide overrides where necessary.
- [ ] Review YamlProcessor implementation and decide if it needs to be cut down even more.
- [ ] Review Winter\Storm\Foundation\Bootstrap\LoadConfiguration -> Illuminate\Foundation\Bootstrap\LoadConfiguration for conformity with Laravel base class (ties into support for `app.env` configuration variable instead of just `APP_ENV`.
- [ ] Review the helpers that we override from Laravel to make them match up as much as possible (https://github.com/laravel/framework/blob/8b2cdb1dc88af4777ae17dd7279936707f40409e/src/Illuminate/Foundation/helpers.php#L820-L837 vs https://github.com/wintercms/storm/blob/develop/src/Support/helpers.php#L28-L41 (for that one we can possibly go back to the Laravel one but extend e() to turn the parameter into a string first.
- [x] https://github.com/wintercms/storm/pull/72
- [ ] https://github.com/wintercms/winter/pull/521
- [x] Check DebugBar dependency for merging of https://github.com/barryvdh/laravel-debugbar/pull/1318

### Documentation:

- [ ] Document new ArraySource trait
- [ ] Document new tests structure
- [ ] Update 1.2 mail docs to reflect the changes to the configuration files and the introduction of the Driver plugins.
- [ ] Event dispatcher now supports closure event handlers and queued closure handlers, review and document

### API Changes:

- [ ] Revise Cache system (see https://github.com/wintercms/winter/commit/9314a4442e2695b3665e2fa2e0cf1788809e02d8)

## Completed Work:

The following items have all been completed as a part of this Laravel 9 upgrade. Great job to everyone who contributed!

### New Features:
- [x] Add support for `#` as a path symbol that represents the themes directory.
- [x] LazyLoad console commands (https://github.com/laravel/framework/pull/34873 & https://github.com/laravel/framework/pull/34925) (Related: https://github.com/wintercms/storm/pull/60)
- [x] Investigate Anonymous Migrations (https://laravel-news.com/laravel-anonymous-migrations) and alias `artisan migrate` to `winter:up` (fixed in https://github.com/wintercms/storm/commit/b16bb179c4d7f56752b1dd5415fb3d03b6a0581c)
- [x] https://github.com/wintercms/storm/pull/40
- [x] https://github.com/wintercms/winter/pull/404

### Maintenance / Cleanup:
- [x] Reorganize tests out of the project tests folder and into module / plugin specific tests folders.
- [x] Reorganize Console commands into the appropriate modules (https://github.com/wintercms/storm/pull/70, https://github.com/wintercms/winter/pull/471, https://github.com/wintercms/winter/issues/270)
- [x] Fix skipped tests
- [x] Review changes to `artisan serve` (included by the framework now instead of required to existing in the project root)
- [x] https://github.com/wintercms/winter/pull/455
- [x] https://github.com/wintercms/winter/pull/581

### Compatibility Layer:
- [x] Review including a polyfill for the removed `{% spaceless %}` tag in Twig v3.0 (https://github.com/wintercms/winter/pull/456)
- [x] Revise YAML processor system to only engage the processors if the original parsing fails (see https://github.com/wintercms/wn-translate-plugin/commit/a219d7ebf5b2ff6097ce3f7c9f06688b665e19e6 for an example of currently valid YAML that fails if it gets preprocessed).
- [x] ~~`Illuminate\Support\Arr` was moved to `Illuminate\Collections\Arr` (https://github.com/laravel/framework/issues/34292#issuecomment-901657091)~~ (namespace remained the same, file location changed)
- [x] TransportManager renamed to MailManager, check to make sure functionality / public API are unaffected
- [x] ErrorHandler changes to typehint / reference "Throwable" instead of "Exception", check to make sure public API / extensions of this behavior continue to work as expected.
- [x] https://github.com/wintercms/winter/pull/522
- [x] https://github.com/wintercms/storm/pull/69
- [x] https://github.com/wintercms/winter/issues/562
- [x] https://github.com/wintercms/winter/issues/567

### Dependencies:
- [x] https://github.com/wintercms/storm/pull/26
- [x] Switching to Assetic as an external dependency removed the change in FilesystemCache where File::chmod($path) is called, investigate if that actually causes issues for us and we need to implement it in an override or if we're fine.
- [x] Review upgrade to Flysystem v3 (v2 support was added in Winter 1.2 WIP branch, v3 needs to be assessed & tested)
- [x] Revert composer dependencies to stable versions
- [x] Bump minimum version of Laravel to 9.1.0 (see https://github.com/laravel/framework/commit/63ca843643e86fb69efc901051ae079c89a7fd09)
- [x] YAML upgraded from 3 to 5.1 (https://github.com/wintercms/storm/pull/21)
- [x] Remove the embedded implementation of Assetic v1 and utilize [Assetic v2](https://github.com/assetic-php/assetic) instead.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/148"></a>


tschallacka:
Question about the three submodules at https://github.com/wintercms/winter/blob/develop/composer.json#L34 

```json
        "winter/wn-system-module": "dev-develop as 1.1.999",
        "winter/wn-backend-module": "dev-develop as 1.1.999",
        "winter/wn-cms-module": "dev-develop as 1.1.999",
```

They all have a dependency on lLravel 6.0
https://github.com/wintercms/wn-cms-module/blob/develop/composer.json#L28  
https://github.com/wintercms/wn-backend-module/blob/develop/composer.json#L28  
https://github.com/wintercms/wn-system-module/blob/develop/composer.json#L28  

Is it intentional that they are individually dependent on Laravel?   
Or would a winter/storm dependency be better and let the Laravel dependency be decided by Storm?

regardless, can I fork the modules and provide pull requests? As they are marked as read only?

LukeTowers:
@tschallacka the modules are locked to the version dependency to stop people from accidentally upgrading them when they don't have the required laravel version. The modules are split manually from the main wintercms/winter repo, so you can fork the main repo and target the 1.2 branch to change their dependencies in a PR and then we can mirror the wip/1.2 branch across when your PR is merged so that you can pull them from composer.

Does that all make sense? It's a somewhat complicated process to be fair.

bennothommo:
@LukeTowers we could potentially link it to a Storm branch, and let Storm be the arbiter of which Laravel version is required.

tschallacka:
Ah, like laravel does with their system, splitting it up in individual libraries.

jaxwilko:
For reference, this PR https://github.com/laravel/framework/pull/38481 removed `Illuminate\Mail\Transport\Transport` which causes:

> PHP Fatal error:  Uncaught Error: Class "Illuminate\Mail\Transport\Transport" not found in .../vendor/winter/storm/src/Mail/Transport/MandrillTransport.php:7

config/app.php
```diff
@@ -132,7 +134,7 @@
     |   - Amazon ELB users should always use the "HEADER_X_FORWARDED_AWS_ELB" option.
     */
 
-    'trustedProxyHeaders' => Illuminate\Http\Request::HEADER_X_FORWARDED_ALL,
+    'trustedProxyHeaders' => 'HEADER_X_FORWARDED_ALL',
```
bennothommo: Should we still use the constant here? There's no garuntee that the string will remain the same?

modules/backend/controllers/Files.php
```diff
@@ -64,33 +65,33 @@ public function thumb($code = null, $width = 100, $height = 100, $mode = 'auto',
...
-            $url = Cache::get($pathKey);
+        // Check to see if the URL has already been generated
+        $pathKey = 'backend.file:' . $path;
+        $url = Cache::get($pathKey, null);
 
-            // The AWS S3 storage drivers will return a valid temporary URL even if the file does not exist
-            if (is_null($url) && $disk->exists($path)) {
-                $expires = now()->addSeconds(Config::get('cms.storage.uploads.temporaryUrlTTL', 3600));
-                $url = Cache::remember($pathKey, $expires, function () use ($disk, $path, $expires) {
+        if ($url !== false && is_null($url) && $disk->exists($path)) {
```
bennothommo: This condition looks a bit wonky. There's no need for the `$url !== false` if we're checking that `$url` is `null`.

```suggestion
        if (is_null($url) && $disk->exists($path)) {
```

modules/system/classes/PluginBase.php
```diff
@@ -432,7 +432,13 @@ public function getPluginVersion(): string
...
         $versionFile = $this->getPluginPath() . '/updates/version.yaml';
 
-        if (!File::isFile($versionFile) || !($versionInfo = Yaml::parse(file_get_contents($versionFile))) || !is_array($versionInfo)) {
+        if (
+            !File::isFile($versionFile) ||
+            !($versionInfo = Yaml::withProcessor(new VersionYamlProcessor, function ($yaml) use ($versionFile) {
+                return $yaml->parseFile($versionFile);
+            })) ||
+            !is_array($versionInfo)
+        ) {
```
bennothommo: I think the PSR guidelines recommend boolean operators being at the start of the line if conditions are multi-line.

```suggestion
            !File::isFile($versionFile)
            || !($versionInfo = Yaml::withProcessor(new VersionYamlProcessor, function ($yaml) use ($versionFile) {
                return $yaml->parseFile($versionFile);
            }))
            || !is_array($versionInfo)
        ) {
```

bennothommo (dismissed):

config/app.php
LukeTowers: The constant has been completely removed.

bennothommo: Looks like the recommendation is to use the full definition of constants instead: https://github.com/symfony/symfony/blob/4537f85963fba75f9ec2aa9a39c048140260f1f9/src/Symfony/Component/HttpFoundation/Request.php#L50-L51

So I think the value would need to be `'trustedProxyHeaders' => Illuminate\Http\Request::HEADER_X_FORWARDED_FOR | Illuminate\Http\Request::HEADER_X_FORWARDED_HOST | Illuminate\Http\Request::HEADER_X_FORWARDED_PORT | Illuminate\Http\Request::HEADER_X_FORWARDED_PROTO`

LukeTowers: That's what it's mapped to in the middleware in the storm repo, I'm switching to a default of the string constant instead because we've already been burnt by them removing a constant which we have no way to polyfill back in, so all default configs from 1.1.4 to 1.1.8 will be forced to make a configuration file change when upgrading to 1.2 because of it or else their entire site will break because the Symfony / Laravel controlled constant no longer exists.

modules/system/classes/PluginBase.php
LukeTowers: Got a link to that? I've always done it the above way so that's probably what's present in the core already. I don't think I've seen it done the other way in the core before.

bennothommo: It's defined in PSR-12 (https://www.php-fig.org/psr/psr-12/) although we've only declared to follow PSR-2 at the moment which doesn't define it.

However, it does say they can be at the beginning or end of a line, but not both:

> Boolean operators between conditions MUST always be at the beginning or at the end of the line, not a mix of both.

Their examples show them at the start of the line, but that's probably the preference of the author.

Personally, I like them at the start because it reads to me like:

- Condition
- OR Condition
- AND Condition

I've also deployed code with this format - eg: [here](https://github.com/wintercms/winter/blob/develop/modules/backend/classes/BackendController.php#L80-L84) and [here](https://github.com/wintercms/winter/blob/develop/modules/backend/widgets/Filter.php#L164-L168). Others have [done so](https://github.com/wintercms/winter/blob/develop/modules/backend/widgets/MediaManager.php#L1414-L1417) as well.

From what I can tell, I think Sam and Alexey used your style mostly.

I don't mind either way - I think we just need to make a ruling, adjust everything to suit and then update the code checks to suit.

LukeTowers: Fair points, I like your reasoning for going with that approach, lets standardize on that and add it to the styling standards.

bennothommo:
I had a look at the following:

> Review changes to artisan serve (included by the framework now instead of required to existing in the project root)

Because I use `artisan serve` all the time.

My understanding is that we're already covered here, as `serve` is provided by Laravel, and it looks for an `.env` flie which we create anyway. Is there anything else we would need to do here?

LukeTowers:
@bennothommo things to look for are what happens when we get rid of the server.php file and if we need to explicitly register the artisan serve command. Then if that's all fine we need to remove the server.php from the repo and include that in the upgrade guide.

jaxwilko:
There is probably going to be an issue with `str_contains`.

php8.0 spec:
image

Winter's implmentation:
image

The issue here is in `< php8.0` the `str_contains()` method will allow for an array of strings to be passed as the second option, but upon updating to `> php8.0` the user will get an error along the lines of `TypeError: str_contains(): Argument #2 ($needle) must be of type string, array given`.

Not sure we can do much apart from checking the codebase to make sure we're not using `str_contains()` with arrays and documenting it for users.

`Str::contains($haystack, $needles)` can be used as a replacement.

bennothommo:
@jaxwilko I think we have to, as a rule, encourage the use of the facades, so we can ensure that they always remain compatible and can account for the discrepancy.

tschallacka:
In this case I would suggest adding a special check to the error handler to make the exception message useful for resolving the issue.

Something like, upgrade notice! Replace str_contains() with Str::contains() to use an array for needles.

LukeTowers:
@tschallacka That might be a good candidate for integrating the Ignition error handler into the core so that we can write those sorts of custom suggestions using their suggestions/solutions framework (plus the error page is just nicer overall)

config/session.php
```diff
@@ -170,43 +175,44 @@
     |
     | By setting this option to true, session cookies will only be sent back
     | to the server if the browser has a HTTPS connection. This will keep
-    | the cookie from being sent to you if it can not be done securely.
+    | the cookie from being sent to you when it can't be done securely.
     |
     */
 
-    'secure' => false,
+    'secure' => env('SESSION_SECURE_COOKIE'),
```
bennothommo:
```suggestion
    'secure' => env('SESSION_SECURE_COOKIE', false),
```

modules/system/classes/PluginBase.php
```diff
@@ -432,7 +432,13 @@ public function getPluginVersion(): string
 
         $versionFile = $this->getPluginPath() . '/updates/version.yaml';
 
-        if (!File::isFile($versionFile) || !($versionInfo = Yaml::parse(file_get_contents($versionFile))) || !is_array($versionInfo)) {
+        if (
+            !File::isFile($versionFile)
+            || !($versionInfo = Yaml::withProcessor(new VersionYamlProcessor, function ($yaml) use ($versionFile) {
+                return $yaml->parseFile($versionFile);
```
LukeTowers, line 436:
```suggestion
                return $yaml->parse(file_get_contents($versionFile));
```

modules/system/classes/UpdateManager.php
```diff
@@ -890,7 +890,7 @@ public function setNotesOutput($output)
      */
     public function requestServerData($uri, $postData = [])
     {
-        $result = Http::post($this->createServerUrl($uri), function ($http) use ($postData) {
+        $result = \Winter\Storm\Network\Http::post($this->createServerUrl($uri), function ($http) use ($postData) {
```
LukeTowers:
```suggestion
        $result = Http::post($this->createServerUrl($uri), function ($http) use ($postData) {
```

modules/system/classes/VersionManager.php
```diff
@@ -345,7 +345,9 @@ protected function getFileVersions($code)
         }
 
         $versionFile = $this->getVersionFile($code);
-        $versionInfo = Yaml::parseFile($versionFile);
+        $versionInfo = Yaml::withProcessor(new VersionYamlProcessor, function ($yaml) use ($versionFile) {
+            return $yaml->parseFile($versionFile);
```
LukeTowers, line 349:
```suggestion
            return $yaml->parse(file_get_contents($versionFile));
```

GrCOTE7:
Hi, all !

I try to apply the documentation about Relationships with this version (Part ManyToMany).

Under the version with Laravel 8, i completly succeed that:

(
Note I have just 2 main tables:
- **grcote7_marriage_guests** (With has only a 1 to 1 relation with **users**) and 4 guests
- **grcote7_marriage_groups** (3 goups: AMI - FAMILLE - TRAVAIL)

and a pivot table named: **grcote7_marriage_group_guest**... as described in the documentation
)

and so, ll is perfect with Laravel 8 :

http://prntscr.com/VXZ1Rs4fvC6h

But, with Laravel 9  (Exactly the same code): 

http://prntscr.com/xf0X0wq6306m

Note:
All the data of this example can easily be reproduced: see the **README.md** of : 

https://github.com/GrCOTE7/prepa

*Hope this can help you to... Help me ! ;-)*

Thanks.

mjauvin:
@GrCOTE7 What do you mean Laravel 8 ? The previous version for winter was Laravel 6.

GrCOTE7:
Hi, mjauvin...

You certainly are right...

I mean the version which has got this requirement : 

http://prntscr.com/qgLBiNhBbOr8

And I 'd imagined 'More than 6' and perhaps quick concluded '8' ;-) !

;-)

bennothommo:
@LukeTowers 

> https://github.com/octobercms/library/commit/29ad91712e5d7ac6e2da70751ea22fe3f2ce9acf (evaluate)

FYI, I've made that same change in the code analysis branch. Symfony's `UploadedFile` class no longer allows the size to be passed through. Symfony 4 had a BC layer in place, but 5 dropped it.

https://github.com/symfony/symfony/pull/25471

LukeTowers:
@bennothommo can you resolve the conflict with the ViewMaker test?

PR #171, opened 2021-05-13
Update queue connection wording.

ericp-mrel:
Update wording to be more inline with what is actually being set, since these are actually a list of connections and are not actually queue drivers. The env variable is also defined as `QUEUE_CONNECTION` so it makes more sense to title this as a queue connection.

### DIFF
--- a/config/queue.php
+++ b/config/queue.php
@@ -4,12 +4,12 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Default Queue Driver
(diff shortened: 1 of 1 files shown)


matteotrubini:
Maybe this file should be aligned to https://github.com/laravel/laravel/blame/6.x/config/queue.php

bennothommo:
Replaced with #179, which aligns most Winter config files with the latest updates to the L6 config files.

PR #178, opened 2021-05-15
Force resized images to be regenerated when the source image changes

### LINKED ISSUE #177: resize - Cached resized image provided even if it the original changed
- Winter CMS Build: 1.1.3
- PHP Version: 8 (and 7.4)
- Database Engine: sqlite
- Plugins Installed: 

### Description:

When an image that has been resized using Winter's built-in resize Twig function, is replaced, the resize does not identify that and provides the resized version of the ORIGINAL image instead of the NEW image - probably as the modification date of the replaced image is not checked against the creation date of the resized image.

Note that https://github.com/toughdeveloper/oc-imageresizer-plugin provides a resize that actually checks for image file modification time and re-generates the images as needed. Perhaps this would be a solution.

### Steps To Reproduce:
1) Place an image in the media directory and resize it, "image.jpg" | media | resize(200, 200)
2) Replace image.jpg with another image
3) Re-run  "image.jpg" | media | resize(200, 200) --> the old image is served.

LukeTowers:
This should fix https://github.com/wintercms/winter/issues/177, but it will cause all existing installations to resize all of their already resized images. 

Another alternative could have been to include the hash of the source image; but it would have been at the expense of more work every time this function is called, which could potentially include generating hashes of multiple extremely large (10+ MB) images on every page load if the page includes such images that need to be resized since the filter needs to identify the configuration in order to identify the path to the resized image to check if it needs to return a resizer/ URL or a /resized one.

Further improvements could potentially be made in the isResized() / getPathToResized() methods to cache the results to avoid doing all the work involved with the getConfig() method while still allowing people to run a command to trigger a refresh.

### DIFF
--- a/modules/system/classes/ImageResizer.php
+++ b/modules/system/classes/ImageResizer.php
@@ -230,6 +230,9 @@ class ImageResizer
                 $disk->getAdapter()->setPathPrefix($realPath);
             }
         }
+        
+        // Include last modified time to tie generated images to the source image
+        $mtime = $disk->lastModified($this->image['path']);
 
         // Handle disks that can't be serialized by referencing them by their
         // filesystems.php config name
@@ -243,6 +246,7 @@ class ImageResizer
             'image' => [
                 'disk' => $disk,
                 'path' => $this->image['path'],
+                'mtime' => $mtime,
                 'source' => $this->image['source'],
             ],
             'width' => $this->width,


bennothommo (approved):

LukeTowers:
@bennothommo did you have any ideas for a better experience for this:

> it will cause all existing installations to resize all of their already resized images.

bennothommo:
@LukeTowers I personally don't see a problem with that. The resized images should be considered temporary and treated as a "cache" so to speak.

LukeTowers:
@bennothommo fair enough, but if we treat it that way then we need to provide a way to clear the cache. Will wait to merge this until we can add a util command to purge the resized images cache.

LukeTowers:
With that as well, should we provide a migration that clears that cache as a part of the update to avoid unnecessarily cluttering up people's storage directory with now-invalid images or is the added server load of regenerating all existing images not worth it (especially if they're caching the generated HTML which could include references to final files that would no longer exist).

bennothommo:
I would envisage that the server load would be staggered, as the resizing only happens on an "as-needed" basis, when people load up a particular page with the resized image(s), so I don't think that's too much of a concern.

We definitely will need some form of command though to clear the cached resized images, if `cache:clear` doesn't already do that.

I'm not sure what we can do about people caching the generated HTML (and ostensibly, the resized image's direct URL). I would hope that doing a `cache:clear` as part of this proposed migration to clear the resized images would also reset the page cache as well. The same problem would've occurred with the previous resize functionality as well.

helmutkaufmann:
> @bennothommo fair enough, but if we treat it that way then we need to provide a way to clear the cache. Will wait to merge this until we can add a util command to purge the resized images cache.

I think it would not be about clearing the actual Cache, it would also be about removing all the resized images. But perhaps that is what you meant anyway.

PR #185, opened 2021-05-23
Updated RO language for modules/system/lang RO files

luciandex:
+Small fix for  modules/system/lang/EN/lang.php file.


bennothommo (changes requested):
@luciandex Would you mind removing the changes made to the `en/lang.php` file? I believe the casings are that way for a reason.

bennothommo (approved):

PR #191, opened 2021-05-26
Update .htaccess

WebVPF:
Correction in comments to allowed and unresolved folders.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -31,7 +31,7 @@
     RewriteRule ^storage/app/uploads/protected/.* index.php [L,NC]
 
     ##
-    ## Allowed listed folders
+    ## Paths explicitly handled by the server
     ##
     RewriteCond %{REQUEST_FILENAME} -f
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -31,7 +31,7 @@
     RewriteRule ^storage/app/uploads/protected/.* index.php [L,NC]
 
     ##
-    ## White listed folders
+    ## Allowed listed folders
```
LukeTowers:
```suggestion
    ## Paths explicitly handled by the server
```

.htaccess
```diff
@@ -19,7 +19,7 @@
     # RewriteRule (.*) https://%{SERVER_NAME}/$1 [L,R=301]
 
     ##
-    ## Black listed folders
+    ## Unresolved listed folders
```
LukeTowers:
```suggestion
    ## Paths explicitly blocked from being handled by the server
```

LukeTowers (changes requested):

LukeTowers (approved):

PR #192, opened 2021-05-27
Optimise CMS page loading by simplifying model query

jaxwilko:
This PR adds the method `CmsObject::listInThemeArray` which can be used to return an `Collection` of arrays containing filenames and url patterns.

Each element of the collection will look something like this:
```php
[
    'fileName' => 'example.htm',
    'pattern' => '/example/url'
]
```

The router has been changed to call this method instead of `listInTheme`. 

In short this means that instead of returning a collection of page objects and storing only filename & url, we can just retrieve the filename and url and directly store them as the url map in the router.

The speed results with these changes are as follows:

Test results

Tests have been done isolating the the time it takes for `$page = $this->router->findByUrl($url);` within `Cms\Classes\Controller` to execute. All tests have been done with 1000 CMS pages.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/192"></a>


modules/cms/classes/CmsObject.php
```diff
@@ -174,10 +176,91 @@ public static function listInTheme($theme, $skipCache = false)
...
-
         return $result;
     }
 
+    /**
+     * Returns the list of files and url patterns in the specified theme.
+     * This method is used internally by the system.
+     * @param \Cms\Classes\Theme $theme Specifies a parent theme.
+     * @param boolean $skipCache Indicates if objects should be reloaded from the disk bypassing the cache.
+     * @param array $extras Request extra fields to be appended to the return
```
bennothommo:
```suggestion
     * @param array $extras Request extra fields to be appended to the return
     * @param boolean $skipCache Indicates if objects should be reloaded from the disk, bypassing the cache.
```

bennothommo:
@jaxwilko when this is ready for review, feel free to take it out of draft :)

jaxwilko:
@bennothommo my bad, didn't realise it was still in draft lol

bennothommo:
@jaxwilko it wasn't - I put it in draft. It still had [WIP] in the title though :P

jaxwilko:
Bump

bennothommo:
I've had a test of this - it does work, but I only see a very small performance benefit. I imagine it's more pronounced when there's more pages to read, but when there's only a handful of pages, it's minuscule.

I just had an idea though - given that for this (and our routing in general), we only want the filename and URL to be retrieved, could we be clever and format the URL to be the first line and just read the first line of pages to determine the URL? It would save us having to parse the sections on each page in order to retrieve the URL, and might make a larger performance boost.

We could maintain backwards compatibility by storing the URL as a "hashbang" of sorts - if the first line doesn't match the format, then assume it's still the old format and would need to be parsed for each of the sections, but we could perhaps save it in this new format on save, going forward?

Eg.

```
#</test-page
title = "Test Page"
layout = "home"
==
Content
```

As opposed to the current:

```
title = "Test Page"
url = "/test-page"
layout = "home"
==
Content
```

bennothommo:
Or even better, could we create a URL cache file and have all CMS objects register a URL in this cache along with a reference to the file name (and perhaps object class) and just use this for routing? Saves us reading all the files altogether.

jaxwilko:
While I was testing I had 1000 CMS pages which each had a img tag with a base64 encoded image to simulate medium sized html content (in terms of bytes). With less pages this issue isn't massively apparent.

It was a while ago i looked at this, but the parser is actually super efficent and just opening the file and reading doing something like:

```php
$fp = fopen($file, 'r');
while ($line = fgets($fp) && !feof($fp) && str_starts_with($line, 'url =')) {
    fclose($fp);
    return trim(explode('=', $line)[1]);
}
```

Was not really even quicker than just calling the parser.

The issue comes from creating a Halcyon model for each page (`Cms\Classes\CmsObject::listInTheme`), then discarding the model straight away (`Cms\Classes\Router::getUrlMap`) which means there was no point to wasting time and memory initing the models.

mjauvin:
@bennothommo for multilingual sites, the translated url in the viewBag needs to be read though...

LukeTowers:
How does this compare to just using the route cache on the next couple of lines? If we have problems with cache invalidation then that's a simple fix, just trigger cache clears whenever the cache would be invalidated

PR #197, opened 2021-05-31
cPanel AutoSSL / Let's Encrypt webroot support

matteotrubini:
Ref: https://github.com/h5bp/server-configs-apache/blob/93460e7ae3ca296089a44b4a4eaab235e473e10d/h5bp/rewrites/rewrite_http_to_https.conf

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -15,8 +15,18 @@
     ##
     ## Uncomment following lines to force HTTPS.
     ##
+    ## (1) If you're using cPanel AutoSSL or the Let's Encrypt webroot method it
+    ##     will fail to validate the certificate if validation requests are
+    ##     redirected to HTTPS. Turn on the condition(s) you need.
+    ##
+    ##     https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml
+    ##     https://tools.ietf.org/html/draft-ietf-acme-acme-12
     # RewriteCond %{HTTPS} off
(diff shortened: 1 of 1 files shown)


mjauvin:
@matteotrubini cPanel has been doing this at server-level for the past few years AFAIK

bennothommo:
@matteotrubini would you be willing to put this in a PR to the docs somewhere - perhaps within [this page?](https://github.com/wintercms/docs/blob/main/setup-configuration.md)

@mjauvin is right, cPanel will auto-inject this workaround for you in most cases, and I'd rather we don't clutter the `.htaccess` file with workarounds for specific hosting environments, if we can avoid it.

matteotrubini:
I understand your view.
If you are unwill to merge this, I can simply make a PR to docs to "remember" users to check H5BP if they have troubles in certificate validation.

More generally, what about to include some H5BP like directives management in `winter:install` or into a dedicated command? It make sense to me for automated deployment.

bennothommo:
@matteotrubini I don't think we'd be against any common-sense changes that would work across all hosting environments, but for configurations for specific hosting environments, such changes could certainly be made available through a plugin which could create its own commands to run and implement those changes for users.

Where possible, Winter CMS should remain environment-agnostic so people who know what they are doing don't have to remove or disable any extraneous boilerplate code.

PR #198, opened 2021-06-01
Updated all RO translations for Winter CMS modules.

luciandex:
- added all missing keys.
- minor changes to system/lang
- some corrections to already present texts
- written all texts with diacritics


luciandex:
The modules/system/lang/ro/lang,php file was saved with another program, so don't realize that was added extra new line at the end of file and added a non wanted char at the beginning before "<?php" tag.
Now it is ok. :)

bennothommo:
Thanks for this, @luciandex :tada:

PR #202, opened 2021-06-08
Add winter:test command

RomainMazB:
First of the console command discussed here https://github.com/wintercms/winter/discussions/201

Accepts any review from anyone out there.

I did not found how it would be possible to display the phpunit running test. As of today it runs the tests "silently" and then output the results.

https://user-images.githubusercontent.com/53976837/121252668-93859c00-c8a8-11eb-961e-869d3c1f135e.mp4


bennothommo:
@RomainMazB it's definitely possible to get the output happening in real-time - see the Dusk command class here, which does the same thing for Dusk tests: https://github.com/wintercms/wn-dusk-plugin/blob/main/console/Dusk.php. I would prefer that the output be in real-time if possible.

The key is to use the `Process` component from Symfony.

RomainMazB:
@bennothommo sure, I'd definitely prefer too. Thanks for pointed out how to do this, I'll definitely implement this

LukeTowers:
Might be nice to have it as `winter:test ?target` where target could be blank for the application's test suite or a specific plugin, and also forward any arguments / options / filters to the phpunit command (example: I use the `--filter` option a lot to run a specific single test)

RomainMazB:
Ok, improvements made.
@LukeTowers to proxy all the phpunit filter we are forced to ignore validation (or create a 50 lines signature to describe all the available options). The same is done into the Laravel's core test command: https://github.com/nunomaduro/collision/blob/stable/src/Adapters/Laravel/Commands/TestCommand.php#L48-L53

A quick view of the final results.
https://user-images.githubusercontent.com/53976837/124030370-68e5b980-d9f6-11eb-893f-c6d57cc45b58.mp4

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+     */
+    public function __construct()
+    {
+        parent::__construct();
+
+        // Because we need to proxy some options that are not defined
+        // We need to ignore validation errors
+        // Same is done in the native Laravel test commands
+        // @link https://github.com/nunomaduro/collision/blob/stable/src/Adapters/Laravel/Commands/TestCommand.php
+        $this->ignoreValidationErrors();
```
LukeTowers:
```suggestion
        /**
         * Ignore validation errors as option proxying is used by this command
         * @see https://github.com/nunomaduro/collision/blob/stable/src/Adapters/Laravel/Commands/TestCommand.php
         */
        $this->ignoreValidationErrors();
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Execute the console command.
+     * @return int
+     */
+    public function handle()
+    {
+        // If the plugin argument is set, search for it and run its tests
+        if ($pluginArgument = $this->argument('plugin')) {
+            $pluginManager = PluginManager::instance();
+            $pluginName = $pluginManager->normalizeIdentifier($pluginArgument);
```
bennothommo: @RomainMazB the `exists()` method will already normalise the given plugin name, so no need to call this.
```suggestion
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Execute the console command.
+     * @return int
+     */
+    public function handle()
+    {
+        // If the plugin argument is set, search for it and run its tests
+        if ($pluginArgument = $this->argument('plugin')) {
```
bennothommo: @RomainMazB where possible, I would like us to avoid assigning variables in a condition. It trips up Xdebug sometimes, and makes it difficult to breakpoint.
```suggestion
        $pluginName = $this->argument('plugin');
        if (!is_null($pluginName)) {
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+            $pluginManager = PluginManager::instance();
+            $pluginName = $pluginManager->normalizeIdentifier($pluginArgument);
+
+            if (!$pluginManager->exists($pluginName)) {
+                throw new \InvalidArgumentException(sprintf('Plugin "%s" not found.', $pluginName));
```
bennothommo: @RomainMazB would it be better to just simply output an error as opposed to throwing an exception? A missing plugin is a fixable error.
```suggestion
                $this->error(sprintf('Plugin "%s" not found.', $pluginName));
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            $pluginDir = $pluginManager->getPluginPath($pluginName);
+
+            $pluginPHPUnitXMLFile = $pluginDir . '/phpunit.xml';
+            if (!file_exists($pluginPHPUnitXMLFile)) {
+                throw new \InvalidArgumentException(sprintf('phpunit.xml file not found in "%s".', $pluginDir));
+            }
```
bennothommo: @RomainMazB same comment about outputting an error as above.

Also, note that distributed packages (like Winter plugins), can also have a `phpunit.xml.dist` file as well, allowing someone to locally use their own `phpunit.xml` if required, so it might be worth checking for the existence of this too.

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+
+            /*
+             * Test plugin
+             */
+            $this->exec[] = "--configuration=$pluginPHPUnitXMLFile";
+            $this->output->writeln(sprintf('<info>Running  plugin\'s tests: %s.</info>', $pluginName));
```
bennothommo:
```suggestion
            $this->info(sprintf('Running  plugin\'s tests: %s.', $pluginName));
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        // Else run winter's tests
```
bennothommo:
```suggestion
        // Run the core Winter CMS tests
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+        else {
+            $this->output->writeln('<info>Running  Winter\'s core tests.</info>');
```
bennothommo:
```suggestion
            $this->info('Running  Winter\'s core tests.');
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
+<?php namespace System\Console;
+
+use Illuminate\Console\Command;
+use Symfony\Component\Process\Exception\ProcessSignaledException;
+use Symfony\Component\Process\Process;
+use System\Classes\PluginManager;
+use Symfony\Component\Console\Input\InputArgument;
+
+/**
+ * Console command to test a plugin or winter's core.
```
bennothommo:
```suggestion
 * Console command to test a plugin or the Winter CMS core.
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+            $pluginManager = PluginManager::instance();
+            $pluginName = $pluginManager->normalizeIdentifier($pluginArgument);
+
+            if (!$pluginManager->exists($pluginName)) {
+                throw new \InvalidArgumentException(sprintf('Plugin "%s" not found.', $pluginName));
+            }
+
+            $pluginDir = $pluginManager->getPluginPath($pluginName);
+
+            $pluginPHPUnitXMLFile = $pluginDir . '/phpunit.xml';
```
LukeTowers: Might also have to look for the phpunit.xml.dist or whatever it is that we're using on the winter storm library

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+use Illuminate\Console\Command;
+use Symfony\Component\Process\Exception\ProcessSignaledException;
+use Symfony\Component\Process\Process;
+use System\Classes\PluginManager;
+use Symfony\Component\Console\Input\InputArgument;
+
+/**
+ * Console command to test a plugin or winter's core.
+ *
+ * This searches for a phpunit.xml file into the plugin's directory and run its tests
```
bennothommo:
```suggestion
 * If a plugin is provided, this command will search for a `phpunit.xml` file inside the plugin's directory and run its tests.
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,112 @@
...
+     */
+    protected $name = 'winter:test';
+
+    protected $signature = 'winter:test {plugin?}';
+
+    /**
+     * The console command description.
+     * @var string
+     */
+    protected $description = "Run tests of Winter's core or an existing plugin.";
```
bennothommo:
```suggestion
    protected $description = "Run tests for the Winter CMS core or an existing plugin.";
```

bennothommo (dismissed):

LukeTowers:
@RomainMazB does the core tests differentiate between the CMS tests and the library tests? @bennothommo should we have a way to trigger them separately or together?

bennothommo:
@LukeTowers we could define all the tests as separate suites in the `phpunit.xml` file and it would allow us to run all tests together, but there's no way to separate the test results - not sure if that's necessary a problem for people.

We used to combine the library and core tests, but I separated them during the L6 update because tests were failing when core PRs were relying on unmerged library PRs.

modules/system/console/WinterTest.php
RomainMazB: I'm not sure to understand this one. `exists()` method returns a boolean, not the plugin name 🤔

RomainMazB:
In the latest commit, we can now pass a custom configuration file using `--configuration` option.

If none is defined, it looks for `phpunit.xml` file and fallback to `phpunit.xml.dist`.

Need precision about my comment on @bennothommo review about `exists()` and the final thoughts on the library and core tests to continue.

modules/system/console/WinterTest.php
bennothommo: @RomainMazB what I mean is that you do not need to normalize the plugin name before calling `exists` below - the `exists` method already normalizes the plugin name.

jaxwilko:
I've made a few changes to improve the arguments passthrough to phpunit, there is an issue where if phpunit arguments are passed before the command arguments then the command args aren't parsed.

Seems to be an issue from `Symfony\Component\Console\Input\ArgvInput`. I shall look into it and see what I can find :)

RomainMazB:
@LukeTowers @bennothommo.
I've been toying with the `ArgvInput` symfony component to find the cause of the issue mentioned by @jaxwilko.

The issue is not directly the fact that we pass a PHPUnit argument before `--plugin` or `--core`, but the fact the `winter:test` command doesn't define those options, which raise a `RuntimeException`, the exception is silently hidden by the `ignoreValidationErrors` rule we are using and present in the Symfony Command class here: https://github.com/symfony/symfony/blob/e34cd7dd2c6d0b30d24cad443b8f964daa841d71/src/Symfony/Component/Console/Command/Command.php#L248-L252

This makes the options to not be correctly set and runs the command without any options at all. I'm afraid that we won't be able to allow phpunit options before the `winter:test` command's ones.

I've added a section in the documentation emphasizing the fact that [phpunit options should be passed **after** the command's options.](https://github.com/wintercms/docs/pull/34/files#diff-8fe50b3d6c3c6248eeb1df04812987226c7bafc5f5305b48277e5ee6f79eaf7aR335-R337) which works well:
image

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+
+        if (($config = $this->option('configuration')) && file_exists($config)) {
+            return $this->execPhpUnit($config, $arguments);
+        }
+
+        $configs = $this->getPhpUnitConfigs();
+
+        if ($this->option('core')) {
+            if (!$configs['core']) {
+                throw new ApplicationException('unable to find core `phpunit.xml`. Try downloading it from github.');
```
LukeTowers:
```suggestion
                throw new ApplicationException("Unable to find the core's phpunit.xml file. Try downloading it from GitHub.");
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+            }
+            $this->info('Running tests for: Winter CMS core');
+
+            return $this->execPhpUnit($configs['core'], $arguments);
+        }
+
+        if ($plugin = $this->option('plugin')) {
+            if (!isset($configs['plugins'][strtolower($plugin)])) {
+                throw new ApplicationException(sprintf('unable to find %s\'s `phpunit.xml`', $plugin));
```
LukeTowers:
```suggestion
                throw new ApplicationException(sprintf("Unable to find %s\'s phpunit.xml file', $plugin));
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+            if (is_array($configs[$type])) {
+                foreach ($configs[$type] as $plugin => $config) {
+                    $this->info('Running tests for plugin: ' . PluginManager::instance()->normalizeIdentifier($plugin));
+                    $exit = $this->execPhpUnit($config, $arguments);
+                    $exitCode = $exitCode === 0 ? $exit : $exitCode;
+                }
+                continue;
+            }
+
+            $this->info('Running tests for: Winter CMS ' . $type);
```
LukeTowers:
```suggestion
            $this->info('Running tests for Winter CMS: ' . $type);
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * Get the console command options.
+     *
+     * @return array
+     */
+    protected function getOptions(): array
+    {
+        return [
+            ['plugin', 'p', InputOption::VALUE_OPTIONAL, 'The name of the plugin. Eg: AuthorName.PluginName', null],
```
LukeTowers:
```suggestion
            ['plugin', 'p', InputOption::VALUE_OPTIONAL, 'The name of the plugin. Ex: AuthorName.PluginName', null],
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+            ['configuration', 'c', InputOption::VALUE_OPTIONAL, 'The path to a phpunit xml config', null],
```
LukeTowers:
```suggestion
            ['configuration', 'c', InputOption::VALUE_OPTIONAL, 'The path to a PHPUnit XML config file', null],
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+            ['core', 'o', InputOption::VALUE_NONE, 'Test the core'],
```
LukeTowers:
```suggestion
            ['core', 'o', InputOption::VALUE_NONE, 'Run the Winter CMS core tests'],
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+    /**
+     * Execute a phpunit test
+     * @param string $config path to configuration file
+     * @param array $args array of params for phpunit
+     *
+     * @return int exit code from process
+     */
+    protected function execPhpUnit(string $config, array $args): int
+    {
+        // find and bind the phpunit executable
```
LukeTowers:
```suggestion
        // Find and bind the phpunit executable
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        $process = new Process(
+            array_merge([$this->phpUnitExec, '--configuration=' . $config], $args),
+            dirname($config),
+            null,
+            null
+        );
+
+        // attempt to set tty mode, if unsupported catch and warn the exception message
```
LukeTowers:
```suggestion
        // Attempt to set tty mode, catch and warn with the exception message if unsupported
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+     * The console command description.
+     * @var string
+     */
+    protected $description = 'Run tests for the Winter CMS core or an existing plugin.';
+
+    /**
+     * Path to phpunit binary
+     * @var ?string
+     */
+    protected $phpUnitExec = null;
```
LukeTowers:
```suggestion
    /**
     * The console command name.
     * @var string
     */
    protected $name = 'winter:test';

    /**
     * @var string The console command signature as ignoreValidationErrors causes options not to be registered.
     */
    protected $signature = 'winter:test {?--p|plugin=} {?--c|configuration=} {?--o|core}';

    /**
     * @var string The console command description.
     */
    protected $description = 'Run tests for the Winter CMS core or an existing plugin.';

    /**
     * @var ?string Path to phpunit binary
     */
    protected $phpUnitExec = null;
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+         */
+        $this->ignoreValidationErrors();
+    }
+
+    /**
+     * Execute the console command.
+     * @throws ApplicationException
+     *
+     * @return int|void
+     */
```
LukeTowers:
```suggestion
    /**
     * Execute the console command.
     *
     * @throws ApplicationException
     * @return int|void
     */
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+        return [];
+    }
+
+    /**
+     * Execute a phpunit test
+     * @param string $config path to configuration file
+     * @param array $args array of params for phpunit
+     *
+     * @return int exit code from process
+     */
```
LukeTowers:
```suggestion
    /**
     * Execute a phpunit test
     *
     * @param string $config Path to configuration file
     * @param array $args Array of params for PHPUnit
     * @return int Exit code from process
     */
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+        return $configs;
+    }
+
+    /**
+     * Search for the config file to use.
+     * Priority order is: phpunit.xml, phpunit.xml.dist
+     * @param string $path
+     *
+     * @return ?string
+     */
```
LukeTowers:
```suggestion
    /**
     * Search for the config file to use.
     * Priority order is: phpunit.xml, phpunit.xml.dist
     */
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+            return $configFilePath;
+        }
+
+        return null;
+    }
+
+    /**
+     * Strips out commands arguments and options in order to return arguments/options for PHPUnit.
+     *
+     * @return array
```
LukeTowers:
```suggestion
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+        return $arguments;
+    }
+
+    /**
+     * Removes flags from argument list and their value if present
+     *
+     * @param array $arguments
+     * @param string $remove
+     *
+     * @return array
```
LukeTowers:
```suggestion
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+
+            return 1;
+        }
+    }
+
+    /**
+     * Find all PHPUnit config files (core, lib, plugins)
+     *
+     * @return array
+     */
```
LukeTowers:
```suggestion
    /**
     * Find all PHPUnit config files (core, lib, plugins)
     */
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+            ['configuration', 'c', InputOption::VALUE_OPTIONAL, 'The path to a phpunit xml config', null],
+            ['core', 'o', InputOption::VALUE_NONE, 'Test the core'],
+        ];
+    }
+
+    /**
+     * Get the console command arguments.
+     *
+     * @return array
+     */
```
LukeTowers:
```suggestion
     * Get the console command arguments.
     */
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        return $exitCode;
+    }
+
+    /**
+     * Get the console command options.
+     *
+     * @return array
+     */
```
LukeTowers:
```suggestion
    /**
     * Get the console command options.
     */
```

LukeTowers (changes requested):

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,263 @@
...
+                throw new ApplicationException("Unable to find the core's phpunit.xml file. Try downloading it from GitHub.");
+            }
+            $this->info('Running tests for: Winter CMS core');
+
+            return $this->execPhpUnit($configs['core'], $arguments);
+        }
+
+        if ($plugin = $this->option('plugin')) {
+            if (!isset($configs['plugins'][strtolower($plugin)])) {
+                throw new ApplicationException(sprintf("Unable to find %s\'s phpunit.xml file', $plugin));
```
LukeTowers:
```suggestion
                throw new ApplicationException(sprintf("Unable to find %s\'s phpunit.xml file", $plugin));
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /**
+     * Get the console command arguments.
+     */
+    protected function getArguments(): array
+    {
+        return [];
+    }
+
```
LukeTowers:
```suggestion
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,255 @@
...
+use Illuminate\Console\Command;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Process\Exception\ProcessSignaledException;
+use Symfony\Component\Process\ExecutableFinder;
+use Symfony\Component\Process\Process;
+use System\Classes\PluginManager;
+use Winter\Storm\Exception\ApplicationException;
+
+/**
+ * Console command to test a plugin or the Winter CMS core.
```
LukeTowers:
```suggestion
 * Console command to run tests for plugins or the Winter CMS core.
```

modules/system/console/WinterTest.php
```diff
@@ -0,0 +1,255 @@
...
+ * If a plugin is provided, this command will search for a `phpunit.xml` file inside the plugin's directory and run its tests.
+ *
+ * @package winter\wn-system-module
+ */
+class WinterTest extends Command
+{
+    /**
+     * The console command name.
+     * @var string
+     */
```
LukeTowers:
```suggestion
    /**
     * @var string The console command name.
     */
```

LukeTowers (approved):

PR #216, opened 2021-06-25
Optimize system_files table performance

### LINKED ISSUE #234: Unable to migrate OctoberCMS on Postgres database
- Winter CMS Build: 1.1.3 (currently latest)
- PHP Version:   7.3.18
- Database Engine: PostgreSQL server version 11.6

### Description:
I performed the migration from OctoberCMS build 472 using `composer update`. On the development system with MySQL everything was ok. On the stage system with Posgres database the upgrade process was ok (no errors), however, I cannot open any page. The exception I get is

```
SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: character varying = integer LINE 1: ...ystem_files" where "system_files"."attachment_id" in (1, 2) ... ^ HINT: No operator matches the given name and argument types. You might need to add explicit type casts. (SQL: select * from "system_files" where "system_files"."attachment_id" in (1, 2) and "system_files"."attachment_type" = RainLab\Blog\Models\Post and "field" = featured_images order by "sort_order" asc)
```

Looking at the table the reason is:  on the database the attribute `system_files.attachment_id` is of type varchar and this type cannot be directly compared to integer:

```
select * from "system_files" 
 where "system_files"."attachment_id" in (1, 2) 
       and "system_files"."attachment_type" = "RainLab\Blog\Models\Post" 
       and "field" = featured_images 
order by "sort_order" asc
```

returns on the plain Postgres database

```
ERROR:  operator does not exist: character varying = integer
LINE 2:  where "system_files"."attachment_id" in (1, 2)
```

Probably Winter issue because with October everything worked fine.

ericp-mrel:
This is in reference to this discussion: https://github.com/wintercms/winter/discussions/208

This PR will drop the single column indexes and convert them into a composite index for improved performance and convert the `attachment_id` column to bigint if the database column only contains numeric values.

Feel free to suggest any changes/improvements because this is my first attempt at this type of a change and want this change to be as safe as possible.

### DIFF
--- a/modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
+++ b/modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
@@ -30,7 +30,7 @@ class DbAttachmentIdToInt extends Migration
         // Convert attachment_id to bigint if the attachment_id column only contains integer values.
         $convert = true;
         DB::table('system_files')
-            ->select(['attachment_id'])
+            ->select(['id', 'attachment_id'])
             ->chunkById(1000, function ($attachmentIds) use (&$convert) {
                 $files = $attachmentIds->filter(function ($file) {
                     return ! is_numeric($file->attachment_id);
--- a/modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
+++ b/modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
@@ -33,7 +33,7 @@ class DbAttachmentIdToInt extends Migration
             ->select(['id', 'attachment_id'])
             ->chunkById(1000, function ($attachmentIds) use (&$convert) {
                 $files = $attachmentIds->filter(function ($file) {
-                    return ! is_numeric($file->attachment_id);
+                    return ((string) $file->attachment_id === (string) intval($file->attachment_id));
                 });
 
                 if ($files->isNotEmpty()) {
--- a/modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
+++ b/modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
@@ -1,0 +1,64 @@
+<?php
+
+use Illuminate\Support\Facades\DB;
+use Winter\Storm\Database\Schema\Blueprint;
+use Winter\Storm\Database\Updates\Migration;
+use Winter\Storm\Support\Facades\Schema;
+
+class DbAttachmentIdToInt extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('system_files', function (Blueprint $table) {
+            // Drop old indexes
+            $table->dropIndex(['attachment_id']);
+            $table->dropIndex(['attachment_type']);
+            $table->dropIndex(['field']);
+
+            $table->index([
+                'field',
+                'attachment_type',
+                'attachment_id',
+            ]);
+        });
+
+        // Convert attachment_id to bigint if the attachment_id column only contains integer values.
+        $convert = true;
+        DB::table('system_files')
+            ->select(['attachment_id'])
+            ->chunkById(1000, function ($attachmentIds) use (&$convert) {
+                $files = $attachmentIds->filter(function ($file) {
+                    return ! is_numeric($file->attachment_id);
+                });
+
+                if ($files->isNotEmpty()) {
+                    return $convert = false;
+                }
+            });
+
+        if ($convert) {
+            Schema::table('system_files', function (Blueprint $table) {
+                $table->unsignedBigInteger('attachment_id')->change();
+            });
+        }
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('system_files', function (Blueprint $table) {
+            $table->index('attachment_id');
+            $table->index('attachment_type');
+            $table->index('field');
+        });
+    }
+}
--- a/modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
+++ b/modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
@@ -56,6 +56,14 @@ class DbAttachmentIdToInt extends Migration
     public function down()
     {
         Schema::table('system_files', function (Blueprint $table) {
+            $table->string('attachment_id')->change();
+
+            $table->dropIndex([
+                'field',
+                'attachment_type',
+                'attachment_id',
+            ]);
+
             $table->index('attachment_id');
             $table->index('attachment_type');
             $table->index('field');
--- a/tests/unit/system/database/migrations/TestConvertingSystemFilesAttachmentIdToInt.php
+++ b/tests/unit/system/database/migrations/TestConvertingSystemFilesAttachmentIdToInt.php
@@ -1,0 +1,173 @@
+<?php
+
+use Illuminate\Support\Facades\DB;
+use Winter\Storm\Database\Schema\Blueprint;
+use Winter\Storm\Support\Facades\Schema;
+
+class TestConvertingSystemFilesAttachmentIdToInt extends TestCase
+{
+    public static $tempMigrationsPath = '';
+
+    protected $migrationFileName = '2021_06_25_143510_Db_Attachment_ID_To_int.php';
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->setupTestFixtures();
+
+        if (Schema::hasTable('migrations')) {
+            // Clean database starting point
+            $this->artisan('winter:down --force');
+        }
+
+        $this->artisan('winter:up');
+    }
+
+    protected function tearDown(): void
+    {
+        $this->tearDownTestFixtures();
+
+        $this->artisan('winter:down --force');
+
+        parent::tearDown();
+    }
+
+    public function test_it_can_change_attachment_id_column_to_int()
+    {
+        $faker = \Faker\Factory::create();
+
+        $data = [];
+
+        foreach (range(1, 100) as $i) {
+            $data[] = [
+                'disk_name' => str_random().'.jpg',
+                'file_name' => str_random().'.jpg',
+                'file_size' => mt_rand(),
+                'content_type' => 'image/jpeg',
+                'title' => $faker->title,
+                'description' => $faker->text,
+                'field' => 'files',
+                'attachment_id' => mt_rand(1, 10000),
+                'attachment_type' => 'System\Models\File',
+                'sort_order' => mt_rand(0, 100),
+                'created_at' => now()->toDateTimeString(),
+                'updated_at' => now()->toDateTimeString(),
+            ];
+        }
+
+        DB::table('system_files')->insert($data);
+
+        $this->enableMigration();
+
+        $this->artisan('winter:up');
+
+        $this->assertEquals('bigint', DB::getSchemaBuilder()->getColumnType('system_files', 'attachment_id'));
+
+        $files = DB::table('system_files')->get();
+
+        foreach ($files as $file) {
+            $this->assertIsInt($file->attachment_id);
+            $this->assertNotEmpty($file->attachment_id);
+            $this->assertTrue($file->attachment_id !== 0);
+        }
+    }
+
+    public function test_it_will_not_change_attachment_id_column_to_int()
+    {
+        $faker = \Faker\Factory::create();
+
+        $data = [];
+
+        foreach (range(1, 100) as $i) {
+            $data[] = [
+                'disk_name' => str_random().'.jpg',
+                'file_name' => str_random().'.jpg',
+                'file_size' => mt_rand(),
+                'content_type' => 'image/jpeg',
+                'title' => $faker->title,
+                'description' => $faker->text,
+                'field' => 'files',
+                'attachment_id' => str_random(10),
+                'attachment_type' => 'System\Models\File',
+                'sort_order' => mt_rand(0, 100),
+                'created_at' => now()->toDateTimeString(),
+                'updated_at' => now()->toDateTimeString(),
+            ];
+        }
+
+        DB::table('system_files')->insert($data);
+
+        $this->enableMigration();
+
+        $this->artisan('winter:up');
+
+        $this->assertEquals('string', DB::getSchemaBuilder()->getColumnType('system_files', 'attachment_id'));
+    }
+
+    public function test_it_will_not_change_attachment_id_column_to_int_when_column_contains_both_strings_and_integers()
+    {
+        $faker = \Faker\Factory::create();
+
+        $data = [];
+
+        foreach (range(1, 100) as $i) {
+            $data[] = [
+                'disk_name' => str_random().'.jpg',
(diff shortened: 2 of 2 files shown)


ericp-mrel:
Wasn't sure if there was really a way to write tests to be able to seed some test data before running the migration.
If there is, I'd be happy to try and implement some tests to make sure this goes as smoothly as possible.

mjauvin:
https://wintercms.com/docs/database/structure#seeder-structure

bennothommo:
@ericp-mrel does the composite index make that much difference? I was under the impression that there wouldn't be much difference between a composite index, or the separate indexes. I figured that only changing the VARCHAR to an integer column would suffice.

ericp-mrel:
@bennothommo In my limited testing it didn't make a ton of difference, but changing VARCHAR to int obviously made the biggest difference. I can remove the index changes if you'd rather not switch to a composite one.

bennothommo:
@ericp-mrel let's do that, if you don't mind. Just keeps the potential points of failure to a minimum.

LukeTowers:
@bennothommo I'm fine with changing the index as well, it did result in a performance increase prior to the id column type change so theoretically it will still help even with the varchar to bigint change. @ericp-mrel perhaps you could do a performance test for the difference between the index changed and unchanged (both using bigint attachment_id columns)?

LukeTowers:
Also, it looks like my other comment on the subject never posted, but is there a way we can use a DB query to determine if the columns can be changed rather than doing that in PHP, especially with the is_numeric() function which would return true for some forms of data that can't be reliably converted

modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
```diff
@@ -0,0 +1,72 @@
...
+            ]);
+        });
+
+        // Convert attachment_id to bigint if the attachment_id column only contains integer values.
+        $convert = true;
+        DB::table('system_files')
+            ->select(['id', 'attachment_id'])
+            ->chunkById(1000, function ($attachmentIds) use (&$convert) {
+                $files = $attachmentIds->filter(function ($file) {
+                    return ! is_numeric($file->attachment_id);
```
bennothommo: @ericp-mrel this might be a more reliable way of determining if an attachment ID is an integer. A conversion of the attachment ID to an integer then to a string should result in the same value as a conversion straight to a string if the ID truly is an integer. It would also help resolve @LukeTowers' correct comment about some numerics passing this check even though they are not an integer.

```suggestion
                    return ((string) $file->attachment_id === (string) intval($file->attachment_id));
```

bennothommo:
@LukeTowers I had this problem a couple of years ago with a project - my understanding at the time was that the different database server types had different ways of handling this, so there might not be a query that can be accurately run across all our supported types.

That being said, things might've changed in the last couple of years since. Perhaps a regular expression?

ericp-mrel:
I tried to find out if there was a way to run a query to determine if there are non-int values in a column, but I couldn't really find anything. I then thought about if a single query could even work for all database engines, because I know some databases handle types differently and I don't have any experience with Postgres or MSSQL and I wanted to make this as reliable as possible which is why I settled on handling this in PHP.

I am open to handling this with a SQL query if it is possible and if anyone knows of a reliable way to do this across all database engines.

RomainMazB:
> Sidenote: I didn't noticed that Sam rollbacked my previous PR that fixed the `keyType` on PostgreSQL installations: https://github.com/octobercms/library/pull/502 (before that we saved the keys as `integer`s into a `varchar` column without typecasting them as `string`s, PostgreSQL is more strict about types)
>
> I didn't tested yet but this PR should fix the bug Sam re-introduced as we now will be more consistent about the `keyType` and value which will be both as `integer`s from the model to the database.

modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('system_files', function (Blueprint $table) {
+            // Drop old indexes
+            $table->dropIndex(['attachment_id']);
+            $table->dropIndex(['attachment_type']);
+            $table->dropIndex(['field']);
```
RomainMazB, line 21: bracket is unnecessary here

modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
```diff
@@ -0,0 +1,72 @@
...
+                if ($files->isNotEmpty()) {
+                    return $convert = false;
+                }
+            });
+
+        if ($convert) {
+            Schema::table('system_files', function (Blueprint $table) {
+                $table->unsignedBigInteger('attachment_id')->change();
+            });
+        }
```
RomainMazB, line 48: Suggestion on how it would be possible to know if conversion is possible directly from the database query:
The function `CAST AS` is valid in: [MySQL](https://dev.mysql.com/doc/refman/5.7/en/cast-functions.html#function_cast), [SQLite](https://sqlite.org/lang_expr.html#castexpr), [PosgreSQL](https://www.postgresql.org/docs/9.2/typeconv-oper.html) and [MSSQL](https://docs.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver15) and probably many other DBMS.

The same way you are doing the double type-casting comparison using PHP, this should work **(testing needed)**:
```php
try {
    $possible =
        DB::table('system_files')
            ->selectRaw('
                SELECT (varchar_ids.total = numeric_ids.total) AS possible
                FROM (
                    SELECT COUNT(*) AS total
                    FROM system_files
                    ) AS varchar_ids
                JOIN (
                    SELECT COUNT(*) AS total
                    FROM system_files
                    WHERE CAST(CAST(attachment_id AS INTEGER) AS VARCHAR) = attachment_id
                    ) AS numeric_ids
                ON true
            ')->value('possible');
} catch (\Exception $e) {
    // Some DB raise an exception when typecast is not possible
    // Abort as this mean conversion is not possible
    return;
}

if ($possible) {
    Schema::table('system_files', function (Blueprint $table) {
        $table->unsignedBigInteger('attachment_id')->change();
    });

    Schema::table('system_files', function (Blueprint $table) {
        // Drop old indexes
        $table->dropIndex('attachment_id');
        $table->dropIndex('attachment_type');
        $table->dropIndex('field');

        $table->index([
            'field',
            'attachment_type',
            'attachment_id',
        ]);
    });
}
```

Also note that with this solution we need to know if the conversion is possible **before** modifying the indexes due to the fact that some DBMS will raise an exception when typecasting is not possible (as PostgreSQL does), so this code represent the whole code of the `up` method.

ericp-mrel: Hey, thank you for attempting to solve this via SQL but this query does not execute for me when running directly on the database (without going through Laravel/PHP).

Here's the error I get back (MariaDB 10.3):
```
[42000][1064] (conn=30) You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ') = attachment_id [42000][1064] You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ') = attachment_id ) AS numeric_ids ON true' at line 9
```

EDIT: Looks like according to the MySQL docs that it only accepts these types: https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#function_cast and for me changing `VARCHAR` to `CHAR` seems to allow the query to execute correctly.

RomainMazB: I thought the same after your answer but using [a syntax checker](https://en.rakko.tools/tools/36/), even `CHAR(n)` doesn't valid, the problem seems to be the double typecasting

EDIT: `INTEGER` also need to be changed to `SIGNED INTEGER`, this is valid: `WHERE CAST(CAST(attachment_id AS SIGNED INTEGER) AS CHAR(300)) = attachment_id`

We now need to verify if these types are valid on other DBMS

EDIT²: after a quick search, `SIGNED INTEGER` is only valid on MySQL/MariaDB systems, and CHAR is not valid on SQLite.
This can be tackled by making a basic test before the request to define the correct type to be used into the query
```php
$connection = DB::getDefaultConnection();

$driver = config("database.connections.{$connection}.driver");

$textType = 'CHAR';
$integerType = 'INTEGER';

if ($driver === 'mssql') { // Not sure we need to support this one?
  $integerType = 'INT';
} elseif ($driver === 'mysql') {
  $integerType = 'SIGNED INTEGER';
} elseif ($driver === 'sqlite') {
  $textType = 'VARCHAR';
}
```
And then use the `$integerType` and `$textType` variables into the query.

EDIT³: The more we dig into this, the more I feel sad about choosing SQL queries over PHP. This will need a lot of various systems testing.

ericp-mrel: I believe the length of the CHAR cast is optional based on my understanding when reading the MySQL docs.

Here's some working code that executes a completely custom/raw query inside of Laravel:
```php
$possible = DB::selectOne(DB::raw('
                SELECT (varchar_ids.total = numeric_ids.total) AS possible
                    FROM (
                        SELECT COUNT(*) AS total
                        FROM system_files
                        ) AS varchar_ids
                    JOIN (
                        SELECT COUNT(*) AS total
                        FROM system_files
                        WHERE CAST(CAST(attachment_id AS INTEGER) AS CHAR) = attachment_id
                        ) AS numeric_ids
                    ON true
            '))->possible;
```

I'll try and do some testing with Postgres and SQLite. I just don't know how to setup a MSSQL database since I've never used it before, so I'm not able to test this there.

modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
ericp-mrel: Supplying the index name as an array actually allows us to just specify the index to drop by field name instead of needing to include the prefix and suffix that are automatically generated by Laravel when creating the index. Basically, it just allows us to omit the `system_files_` prefix and the `_index` suffix from the index name, because the full index name is actually `system_files_field_index`

Reference: https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Schema/Blueprint.php#L1319

modules/system/database/migrations/2021_06_25_143510_Db_Attachment_ID_To_int.php
RomainMazB: I'm not sure we need to support MSSQL as I can't find any mention of it into the documentation (ping @LukeTowers).

Also `CHAR` in your last comment will return only the first character on some DBMS (just tested on PostgreSQL), you need to specify a length inside parenthesis. System default is 191 since laravel uses utf8mb4 character set but a higher value won't break anything so we are safe with: `CHAR(300)`

LukeTowers: @RomainMazB SQLServer is supported by Laravel so yes, we need to support MSSQL.

RomainMazB: Ok, so we should use the conditions I've made into [my comment](https://github.com/wintercms/winter/pull/216#discussion_r671536613).
Equivalent types for type casting are:
- INT and CHAR on MSSQL ([Basic documentation example similar to us using both INT and CHAR](https://docs.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver15#e-using-cast-with-the-like-clause))
- SIGNED INTEGER and CHAR on MySQL/MariaDB ([Types cast MySQL documentation](https://dev.mysql.com/doc/refman/5.7/en/cast-functions.html#function_cast) / [Similar on MariaDB](https://mariadb.com/kb/en/cast/))
- INTEGER and VARCHAR on SQLite ([Types cast SQLite documentation](https://sqlite.org/datatype3.html#affinity_name_examples))
- INTEGER and CHAR on PostgreSQL ([PostgreSQL datatypes documentation](https://www.postgresql.org/docs/9.5/datatype-numeric.html))
```php
$connection = DB::getDefaultConnection();

$driver = config("database.connections.$connection.driver");

$textType = 'CHAR';
$integerType = 'INTEGER';

if ($driver === 'mssql') {
  $integerType = 'INT';
} elseif ($driver === 'mysql') {
  $integerType = 'SIGNED INTEGER';
} elseif ($driver === 'sqlite') {
  $textType = 'VARCHAR';
}

$possible =
  DB::selectOne(DB::raw("
      SELECT (varchar_ids.total = numeric_ids.total) AS possible
          FROM (
              SELECT COUNT(*) AS total
              FROM system_files
              ) AS varchar_ids
          JOIN (
              SELECT COUNT(*) AS total
              FROM system_files
              WHERE CAST(CAST(attachment_id AS $integerType) AS $textType(300)) = attachment_id
              ) AS numeric_ids
          ON true
  "))->possible;
```

bennothommo: @RomainMazB @ericp-mrel I'm still leaning towards us keeping the type check in PHP using the method already in this PR. This query is getting too complex for a query that's going to be run once (on migration) and that's it. I don't think there's any expectation by users that migrations have to be "fast", so I don't mind if this conversion takes a couple of seconds.

RomainMazB:
That's my feeling too, as said in one on my previous comment's edit [here](https://github.com/wintercms/winter/pull/216#discussion_r671536613), the more we was digging this, the more complicated the query was, the more I felt sad to use SQL to solve this.

But @luketowers was worried about is_numeric method which could generate false truthy [here]( https://github.com/wintercms/winter/pull/216#issuecomment-869927792).

Now that we have both solution you are free to chose.

bennothommo:
@RomainMazB I'd suggested a fix here (https://github.com/wintercms/winter/pull/216#discussion_r660277003) which should resolve @LukeTowers concerns. Everything would need to be an integer to pass that check.

mjauvin:
I would definitely delay this to 1.2.1 personally.

LukeTowers:
@bennothommo what was left on this to get it merged?

bennothommo:
@LukeTowers sorry for the delay on getting back to you on this one.

On reviewing it, I think this PR has kinda gone in two directions. Originally, it was to change the indexes in the `system_files` table to a composite index for apparently better performance. I'm not too keen on this as I'm not sure of its effect on other database types and I haven't seen any hard numbers to determine its performance benefit.

However, there is also a fix to change the attachment IDs to a `bigint` column if the table only contains numbers. I think this is needed because some strict SQL servers (like PostgreSQL or MySQL with some SQL modes in play) will type the column as a string column and will reject integer values outright, as opposed to coercing them to strings. This part of the PR I believe is needed to fix the linked issues.

LukeTowers:
@bennothommo are you interested in resubmitting the relevant part of the PR that you want to keep?

As an aside, @ericp-mrel did provide hard performance numbers here: https://github.com/wintercms/winter/discussions/208#discussioncomment-889912 and they were fairly convincing in favour of this PR in its entirety.

bennothommo:
Fair enough, so it does have a performance benefit (for MySQL), but we'd need to see if that's the same with the other database types too.

I can certainly make a PR for the `bigint` change.

LukeTowers:
Ping @bennothommo 😉

PR #217, opened 2021-06-25
Reset column order for lists when using "Reset to Default" action

mjauvin:

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -841,7 +841,7 @@ class Lists extends WidgetBase
         /*
          * Use a supplied column order
          */
-        if ($columnOrder = $this->getSession('order', null)) {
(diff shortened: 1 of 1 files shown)


modules/backend/widgets/Lists.php
```diff
@@ -1685,6 +1685,7 @@ public function onApplySetup()
      */
     public function onResetSetup()
     {
+        $this->putSession('order', null);
```
LukeTowers: Can we just make the order a userPreference same as the other ones? Any reason why this is managed a different way than the other ones?

mjauvin: No idea why, I can convert to a user preference if you want.

bennothommo (approved):
Looks good to me. Happy to merge if it's been tested.

mjauvin:
Yes, it's working great!

PR #222, opened 2021-07-01
Return 404 for asset combiner resources that don't exist

RickAcb:
Returning a 500 error (server error) for a error which should be a 404.

### DIFF
--- a/modules/system/classes/SystemController.php
+++ b/modules/system/classes/SystemController.php
@@ -27,7 +27,7 @@ class SystemController extends ControllerBase
     {
         try {
             if (!strpos($name, '-')) {
-                throw new ApplicationException(Lang::get('system::lang.combiner.not_found', ['name' => $name]));
+                return Response::make('/* '.e(Lang::get('system::lang.combiner.not_found', ['name' => $name])).' */', 404);
             }
 
             $parts = explode('-', $name);
@@ -38,7 +38,7 @@ class SystemController extends ControllerBase
 
             return $combiner->getContents($cacheId);
         } catch (Exception $ex) {
-            return Response::make('/* '.e($ex->getMessage()).' */', 404);
+            return Response::make('/* '.e($ex->getMessage()).' */', 500);
         }
     }
 
--- a/modules/system/classes/SystemController.php
+++ b/modules/system/classes/SystemController.php
@@ -38,7 +38,7 @@ class SystemController extends ControllerBase
 
             return $combiner->getContents($cacheId);
         } catch (Exception $ex) {
-            return Response::make('/* '.e($ex->getMessage()).' */', 500);
+            return Response::make('/* '.e($ex->getMessage()).' */', 404);
         }
     }
 
--- a/modules/system/classes/SystemController.php
+++ b/modules/system/classes/SystemController.php
@@ -21,7 +21,7 @@ class SystemController extends ControllerBase
     /**
      * Combines JavaScript and StyleSheet assets.
      * @param string $name Combined file code
-     * @return string Combined content.
+     * @return Response Combined content.
      */
     public function combine($name)
     {
--- a/modules/system/classes/CombineAssets.php
+++ b/modules/system/classes/CombineAssets.php
@@ -234,7 +234,7 @@ class CombineAssets
     /**
      * Returns the combined contents from a prepared cache identifier.
      * @param string $cacheKey Cache identifier.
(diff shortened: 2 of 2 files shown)


LukeTowers:
Most of the time I encounter this, a 500 error is valid because it's usually caused by a parsing error, could you provide more information on your exact situation where a 404 was more applicable?

RickAcb:
The response 500 gives the browser the indication that the server encountered an unexpected condition that prevented it from fulfilling the request.

The 5xx codes are really severed for errors on the server side and not for when a file is not found. A 400 Bad Request can also be send as it is clearly a user error and not a server error.

The reason I found this is because get errors from scanning tools that the A6 of OWASP is in danger. Giving back 500 server errors can give a great point for DDOS attacks as DDOS protectors like CloudFlare don't "cache" 500 errors but 4xx errors are, depending on the settings of the protectors.

LukeTowers:
Can you provide me specific URLS that are causing the problem? Sounds like it actually needs to return both 404 and 500 depending on where the issue happened, so probably replace the throw applicationexception with the return 404 instead of replacing the try catch handler with 404

RickAcb:
1t is totally true that there is a chance a server error would occur which should not report as 404 Not Found. I have changed the Exceptions code to a response.

I didn't catch all the ApplicationException to a 404 as the ApplicationException is a global Exception and in later usage it might been thrown inside the combiner code. (atm it was only used for the not found as far I could find)

bennothommo:
I don't necessarily think it is up to us to determine whether a compilation error is the result of a missing file or parse error. At the end of the day, if compilation fails, it is due to something to do with the asset files on server, and thus the 500 error is applicable as it is a developer concern. It's not a failure on the client or visitor.

modules/system/classes/SystemController.php
```diff
@@ -27,7 +27,7 @@ public function combine($name)
     {
         try {
             if (!strpos($name, '-')) {
-                throw new ApplicationException(Lang::get('system::lang.combiner.not_found', ['name' => $name]));
+                return Response::make('/* '.e(Lang::get('system::lang.combiner.not_found', ['name' => $name])).' */', 404);
```
LukeTowers, line 30: Can you update the docblock `@return string` to say `@return Response` for both changed methods please?

RickAcb:
> 
> 
> I don't necessarily think it is up to us to determine whether a compilation error is the result of a missing file or parse error. At the end of the day, if compilation fails, it is due to something to do with the asset files on server, and thus the 500 error is applicable as it is a developer concern. It's not a failure on the client or visitor.

The problem is that there is no compilation error it is an error that something is not found.

When you open a page as example https://wintercms.com/i-am-an-evil-monster the website will return 404, page not found as the page doesn't exist.
The same goes for non existing assests for example: https://wintercms.com/themes/frost/assets/images/i-am-a-monster-truck.jpg

But if you change the combine URL for example: https://wintercms.com/combine/winter.css it is suddenly a 500 server error but the file doesn't exists and also shouldn't exist.
There is a change that the system generates a valid combine URL and then it would now return a 404 however between the generation of the link and the retrieving of the link is a very and very small change of failing.

The issue there is with granting 500 errors is that any caching server or (D)DOS service will respect the 500 error and not cache the request. If you send a 404 error back with the cache time of 10 minutes for example CloudFlare will only fetch this page once instead of all the time.

Sending different URLs after the page will trigger the (D)DOS system in CloudFlare but sending the same URL when there is a 500 error code is more allowed and all the traffic is forwarded to the server.

We also log all 5xx errors (almost all big sites do/should) we have tons of records just because the combine gives not found errors on the most strange URL combinations.

I can understand you want to have the 500 error for the very small change there is a valid error. I searched to code and I only found 2 places that could trigger errors (unless functions don't exists).

The first error I found is that the cache system doesn't work at all this will send a 404 because the Config::get() will return no data however it could also generate a 500 code depending on how broken the cache system is.

The second error is that storage_path() generates an error but that will still send a 500 error code back.

LukeTowers:
> I can understand you want to have the 500 error for the very small change there is a valid error. I searched to code and I only found 2 places that could trigger errors (unless functions don't exists).

There's actually more. Any actual error in the compilation of the asset files themselves (whether that's less, sass, or a failing path in the JS combiner, or any custom asset compiling logic) will cause an error; and that's the key 500 error that we wish to maintain and was the original intent behind the existing logic.

PR #230, opened 2021-07-04
Fix response code when not found assets

Omikorin:
Previously Winter returns 500 error code in response when not found assets during combining JavaScript and CSS. It would make sense but caught exception uses `system::lang.combiner.not_found` text suggesting it should returns 404 code.


bennothommo:
Duplicate of https://github.com/wintercms/winter/pull/222.

PR #231, opened 2021-07-04
Sort order type casting for navigation menu, quick action and side menu items

### LINKED ISSUE #229: Back-end navigation menu throws exception when menu order value is empty in plugin.yaml
- Winter CMS Build: dc0d6aa (currently latest dev)
- PHP Version: 8.04
- Database Engine: [not related]
- Plugins Installed: [not related]

### Description:

`order: ''` in `plugin.yaml` for navigation menu items cause Exception and almost all back-end is inaccessible until `plugin.yaml` is manually edited.

Exception:

This issue is related to https://github.com/wintercms/wn-builder-plugin/issues/7 - although the Builder plugin is the easiest way to cause it, it is also possible to manually specify this value into `plugin.yaml`.

From one point of view I understand that specified value should be only valid integer but I also thing that there should be proper data type handling in sorting functions in `NavigationManager.php`:

- https://github.com/wintercms/winter/blob/develop/modules/backend/classes/NavigationManager.php#L121
- https://github.com/wintercms/winter/blob/develop/modules/backend/classes/NavigationManager.php#L124
- https://github.com/wintercms/winter/blob/develop/modules/backend/classes/NavigationManager.php#L147
- https://github.com/wintercms/winter/blob/develop/modules/backend/classes/NavigationManager.php#L154

^ For example, `return (int) $a->order - (int) $b->order;` - although then another issue may arise that these empty or invalid/non-numeric menu items would be shown at the beginning of the Back-end menu.

arvislacis:
Fixes https://github.com/wintercms/winter/issues/229 and https://github.com/wintercms/wn-builder-plugin/issues/7

### DIFF
--- a/modules/backend/classes/NavigationManager.php
+++ b/modules/backend/classes/NavigationManager.php
@@ -118,10 +118,10 @@ class NavigationManager
          * Sort menu items and quick actions
          */
         uasort($this->items, static function ($a, $b) {
-            return $a->order - $b->order;
+            return (int) $a->order - (int) $b->order;
         });
         uasort($this->quickActions, static function ($a, $b) {
-            return $a->order - $b->order;
+            return (int) $a->order - (int) $b->order;
         });
 
         /*
@@ -151,7 +151,7 @@ class NavigationManager
              * Sort side menu items
              */
             uasort($item->sideMenu, static function ($a, $b) {
-                return $a->order - $b->order;
+                return (int) $a->order - (int) $b->order;
             });
 
             /*


LukeTowers:
Thanks for reporting this and proposing a fix @arvislacis, this should be fixed in https://github.com/wintercms/winter/commit/763db7eb8097fbd2c8ca19cd92935fcd1f5abe5f

arvislacis:
@LukeTowers No problem :) Yeah, I see that your changes are a lot better - covers the main cause of the problem.

PR #242, opened 2021-07-19
check overrides for parent locale when compiling language files

mjauvin:
- fallback to parent locale for client localization strings overrides

### DIFF
--- a/modules/system/console/WinterUtil.php
+++ b/modules/system/console/WinterUtil.php
@@ -191,7 +191,6 @@ class WinterUtil extends Command
              * Generate messages
              */
             $fallbackPath = base_path() . '/modules/system/lang/en/client.php';
-
             $srcPath = base_path() . '/modules/system/lang/'.$locale.'/client.php';
 
             $messages = require $fallbackPath;
--- a/modules/system/console/WinterUtil.php
+++ b/modules/system/console/WinterUtil.php
@@ -191,9 +191,11 @@ class WinterUtil extends Command
              * Generate messages
              */
             $fallbackPath = base_path() . '/modules/system/lang/en/client.php';
+
             $srcPath = base_path() . '/modules/system/lang/'.$locale.'/client.php';
 
             $messages = require $fallbackPath;
+
             if (File::isFile($srcPath) && $fallbackPath != $srcPath) {
                 $messages = array_replace_recursive($messages, require $srcPath);
             }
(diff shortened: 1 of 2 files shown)


modules/system/console/WinterUtil.php
```diff
@@ -204,6 +204,12 @@ protected function utilCompileLang()
             $overridePath = base_path() . '/lang/'.$locale.'/system/client.php';
             if (File::isFile($overridePath)) {
                 $messages = array_replace_recursive($messages, require $overridePath);
+            } else if (str_contains($locale, '-')) {
```
LukeTowers:
```suggestion
            } elseif (str_contains($locale, '-')) {
```

modules/system/console/WinterUtil.php
```diff
@@ -204,6 +204,12 @@ protected function utilCompileLang()
             $overridePath = base_path() . '/lang/'.$locale.'/system/client.php';
             if (File::isFile($overridePath)) {
                 $messages = array_replace_recursive($messages, require $overridePath);
+            } elseif (str_contains($locale, '-')) {
+                list ($parentLocale, $country) = explode('-', $locale);
```
LukeTowers:
```suggestion
                list($parentLocale, $country) = explode('-', $locale);
```

modules/system/console/WinterUtil.php
```diff
@@ -204,6 +204,12 @@ protected function utilCompileLang()
             $overridePath = base_path() . '/lang/'.$locale.'/system/client.php';
             if (File::isFile($overridePath)) {
                 $messages = array_replace_recursive($messages, require $overridePath);
+            } elseif (str_contains($locale, '-')) {
+                list($parentLocale, $country) = explode('-', $locale);
+                $overridePath = base_path() . '/lang/'.$parentLocale.'/system/client.php';
+                if (File::isFile($overridePath)) {
+                    $messages = array_replace_recursive($messages, require $overridePath);
+                }
```
LukeTowers, line 219: This probably shouldn't be an elseif because I would assume that client overrides behave like any other override by respecting the fallback language. 

I.e. if I have `/lang/en/system/client.php` and `/lang/en-ca/system/client.php`, I would expect it to merge all of the messages such that `array_replace_recursive($original, $parentLocale, $childLocale)` with appropriate handling for if a child locale exists but the parent doesn't and vice versa.

mjauvin: Good point @LukeTowers, I'll make the change!

LukeTowers:
@mjauvin @bennothommo do you guys think it's worth unit testing this or no?

mjauvin:
I guess so, yeah.

mjauvin:
Any idea on how to approach this @LukeTowers ?

LukeTowers:
Probably would need to break it into a separate method to be more testable? Isn't there anything preexisting in the lang class that can handle getting and merging the messages appropriately?

bennothommo:
@LukeTowers:

> @mjauvin @bennothommo do you guys think it's worth unit testing this or no?

Yes! All the test cases!

@mjauvin:

> Any idea on how to approach this @LukeTowers ?

You should be able to test this method direct, if you use the [`callProtectedMethod`](https://github.com/wintercms/winter/blob/develop/tests/TestCase.php#L31) helper function in the `TestCase` class. Since the test cases will have a full copy of Winter CMS when running the tests, you could simply make some modifications to the necessary lang files to test different scenarios. Ultimately, you would be looking at the resulting JS for the test, I would imagine.

jaxwilko:
@mjauvin @bennothommo had a bit of time free this morning so I've added a test for lang compiling. Let me know what you guys think :)

LukeTowers:
LGTM, thanks @mjauvin and @jaxwilko!

PR #244, opened 2021-07-19
Fix null saving in relation widgets when hidden/disabled from model

### LINKED ISSUE #15: BelongsTo relationship fields lose data when saved if hidden in filterFields()
Refs octobercms/october#3568

RomainMazB:
This fix https://github.com/wintercms/winter/issues/15 :
`filterFields` method is called through the rendering process to hide/disable the form fields, but is never called during the post saving process.

This does not cause any issue into basic form fields because hiding/disabling a basic input field simply remove the data from the post() and so is not saved, but concerning the relation: the field keep "being here" (because it's not hidden into the fields.yaml) and so is saved with null (since it's not present into the post data).

### DIFF
--- a/modules/backend/formwidgets/RecordFinder.php
+++ b/modules/backend/formwidgets/RecordFinder.php
@@ -1,5 +1,6 @@
 <?php namespace Backend\FormWidgets;
 
+use Backend\Classes\FormField;
 use Lang;
 use ApplicationException;
 use Backend\Classes\FormWidgetBase;
@@ -241,6 +242,10 @@ class RecordFinder extends FormWidgetBase
      */
     public function getSaveValue($value)
     {
+        if ($this->formField->disabled || $this->formField->hidden) {
+            return FormField::NO_SAVE_DATA;
+        }
+
         return strlen($value) ? $value : null;
     }
 
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -173,6 +173,8 @@ class FormController extends ControllerBehavior
 
         $this->formWidget->bindToController();
 
+        $this->formWidget->applyFiltersFromModel();
+
         /*
          * Detected Relation controller behavior
          */
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1235,7 +1235,7 @@ class Form extends WidgetBase
     /*
      * Allow the model to filter fields.
      */
-    protected function applyFiltersFromModel()
+    public function applyFiltersFromModel()
     {
         /*
          * Standard usage


RomainMazB:
This fix is easily testable through the test plugin but a PR won't be efficient as the problem is mostly viewable inside the database and only four lines of code is needed to reproduce.

To reproduce, just add this code to [the end of Person.php](https://github.com/wintercms/wn-test-plugin/blob/main/models/Person.php#L58):
```php
public function filterFields($fields)
{
    // $fields->phone->hidden = true;
}
```
- Go to the people update action: `/backend/winter/test/people/update/6`
- Set a phone type and save: all is ok
- Uncomment the code above to make the phone field hidden
- Refresh the page: phone field disappear, just save
- Comment the code to make the field viewable again
- Refresh > The phone as been set to null

https://user-images.githubusercontent.com/53976837/126238269-acef6d9e-e1ef-4b21-b376-324fa921f438.mp4

modules/backend/formwidgets/RecordFinder.php
```diff
@@ -241,6 +242,10 @@ protected function loadAssets()
      */
     public function getSaveValue($value)
     {
+        if ($this->formField->disabled || $this->formField->hidden) {
+            return FormField::NO_SAVE_DATA;
+        }
```
LukeTowers, line 247: Why doesn't the Form widget handle this for all custom widgets already? Is there an example of other widgets using this pattern already?

RomainMazB: I'm not sure why the **Form** widget doesn't handle this at its level or even if it should, I mean that we may should keep this logic at the Widget level (and not the Form) to ease the build of customized FormWidget.

Some FormWidget use it like the [DatePicker](https://github.com/wintercms/winter/blob/develop/modules/backend/formwidgets/DatePicker.php#L151-L153), [MediaFinder](https://github.com/wintercms/winter/blob/develop/modules/backend/formwidgets/MediaFinder.php#L111-L113), [Relation](https://github.com/wintercms/winter/blob/develop/modules/backend/formwidgets/Relation.php#L174-L176), some don't like [the ColorPicker doesn't](https://github.com/wintercms/winter/blob/develop/modules/backend/formwidgets/ColorPicker.php#L145-L148). 

And from what I discovered, this issue **may** appear with some other FormWidget that I'll test like the ColorPicker (returning null as default), the DataTable ([which return `[]` as default](https://github.com/wintercms/winter/blob/develop/modules/backend/formwidgets/DataTable.php#L116) and may re-init the database json field), the Sensitive and some others.

I'll write some failing tests to see and show the range of this issue.

LukeTowers: @RomainMazB are you still interested in writing some tests for this and getting it pushed through?

PR #245, opened 2021-07-20
Add client-side localization strings for german locale

datune:
Finished translating all strings to german.


LukeTowers:
@mjauvin we'll need the `compile lang` command run to commit the updated JS file

PR #253, opened 2021-07-23
avoid translation error

PrestaSafe:
If we use Renatio.Builder and Offline.Mall plugin, an error can show up because template `code` is an array

### DIFF
--- a/modules/backend/widgets/form/partials/_field_dropdown.htm
+++ b/modules/backend/widgets/form/partials/_field_dropdown.htm
@@ -27,13 +27,17 @@
         <?php endif ?>
         <?php foreach ($fieldOptions as $value => $option): ?>
             <?php
-                if (!is_array($option)) $option = [$option];
+                if (!is_array($option) )  $option = [$option];
             ?>
-            <option
-                <?= $field->isSelected($value) ? 'selected="selected"' : '' ?>
-                <?php if (isset($option[1])): ?>data-<?=strpos($option[1],'.')?'image':'icon'?>="<?= $option[1] ?>"<?php endif ?>
-                value="<?= e($value) ?>"
-            ><?= e(trans($option[0])) ?></option>
+            <?php if (!is_array( trans($option[0]) )): ?>
+                <option
+                    <?= $field->isSelected($value) ? 'selected="selected"' : '' ?>
+                    <?php if (isset($option[1])): ?>data-<?=strpos($option[1],'.')?'image':'icon'?>="<?= $option[1] ?>"<?php endif ?>
+                    value="<?= e($value) ?>"
+                ><?= e(trans($option[0])) ?>
+            
+                </option>
+            <?php endif; ?>
         <?php endforeach ?>
     </select>
 <?php endif ?>


LukeTowers:
@PrestaSafe please provide the code sample that causes an issue. Also, I've never heard of Renatio.Builder, got a link?

RomainMazB:
I guess @PrestaSafe talks about [renation.formbuilder](https://octobercms.com/plugin/renatio-formbuilder).

As a paid plugin I can't investigate further into the plugin's code but after [some research](https://octobercms.com/plugin/support/renatio-formbuilder?search=array), it seems that the plugin has trouble with the translate plugin.

Are you using Winter.Translate and which version of the renation.formbuilder plugin are you using?

Also what is the initial error that is raised?

LukeTowers:
Sounds more like an issue with the third party plugin, closing it for now unless more details can be provided.

PrestaSafe:
Hi, thanks for the reply, 

i don't khnow if the problems came to the plugin, in my old october version, it works fine. 
I've this issue without this patch: 
Capture d’écran 2021-07-25 à 09 08 46

The relation detail: 
```yaml
template:
        label: renatio.formbuilder::lang.field.template
        type: relation
        nameFrom: code
        emptyOption: renatio.formbuilder::lang.template.empty
        span: right
```

```php
// use System\Models\MailTemplate;
 public $belongsTo = [
        'template' => MailTemplate::class,
]
```


Ok if you don't accept the patch, but, there is a way to override this view from a plugin ? Because without this, i can't use this plugin. 

Thanks

PrestaSafe:
@RomainMazB i've the latest version of this plugin. 
Thanks for your answers !

LukeTowers:
@PrestaSafe seems like the relation formwidget is returning a list of values that aren't expected by the rest of the system. Are you able to wrap that htmlspecialchars call in a try catch and dd() on fail so we can see what the invalid values it's being passed are?

PrestaSafe:
@LukeTowers you can see the `$option` who can't work. 
```php 
$errors[] = [
                    '$option' => $option,
                    '$option[0]' => $option[0],
                    'trans[option[0]]' => trans($option[0])
                ]; 
```

Capture d’écran 2021-07-26 à 15 05 50

LukeTowers:
I don't understand what I'm looking at, can you provide more clarification?

PrestaSafe:
i put the code for a better understanding. 
`$errors` contains the fields who are in "errors" 
```php
<?php
    $fieldOptions = $field->options();
    $useSearch = $field->getConfig('showSearch', true);
    $emptyOption = $field->getConfig('emptyOption', $field->placeholder);
?>
<!-- Dropdown -->
<?php if ($this->previewMode || $field->readOnly): ?>
    <div class="form-control" <?= $field->readOnly ? 'disabled="disabled"' : ''; ?>>
        <?= (isset($fieldOptions[$field->value])) ? e(trans($fieldOptions[$field->value])) : '' ?>
    </div>
    <?php if ($field->readOnly): ?>
        <input
             type="hidden"
             name="<?= $field->getName() ?>"
             value="<?= $field->value ?>">
    <?php endif; ?>
<?php else: 
    $errors = []; ?>
    <select
        id="<?= $field->getId() ?>"
        name="<?= $field->getName() ?>"
        class="form-control custom-select <?= $useSearch ? '' : 'select-no-search' ?>"
        <?= $field->getAttributes() ?>
        <?= $field->placeholder ? 'data-placeholder="'.e(trans($field->placeholder)).'"' : '' ?>
        >
        <?php if ($emptyOption): ?>
            <option value=""><?= e(trans($emptyOption)) ?></option>
        <?php endif ?>
        <?php foreach ($fieldOptions as $value => $option): ?>
            <?php
                if (!is_array($option) )  $option = [$option];
            ?>
            <?php if (!is_array( trans($option[0]) )): ?>
                <option
                    <?= $field->isSelected($value) ? 'selected="selected"' : '' ?>
                    <?php if (isset($option[1])): ?>data-<?=strpos($option[1],'.')?'image':'icon'?>="<?= $option[1] ?>"<?php endif ?>
                    value="<?= e($value) ?>"
                ><?= e(trans($option[0])) ?>
            
                </option>
            <?php 
        else: 
            $errors[] = [
                '$option' => $option,
                '$option[0]' => $option[0],
                'trans[option[0]]' => trans($option[0])
            ]; 

        endif; ?>
        <?php endforeach ?>
    </select>
    <?php dd($errors); ?>
<?php endif ?>

```

PR #273, opened 2021-08-21
Impersonation improvements

LukeTowers:
See wintercms/storm#47 for more details.

Todo:
- [x] Make it more obvious in the backend UX when a user is being impersonated
- [x] Validate any logging systems in place (including first party plugins) to ensure they correctly identify and log impersonators
- [ ] Add integration tests for user authentication, permissions, impersonation systems

### DIFF
--- a/modules/backend/models/User.php
+++ b/modules/backend/models/User.php
@@ -246,7 +246,7 @@ class User extends UserBase
      * present and the impersonator has access to `backend.impersonate_users`, and the impersonator is not the
      * user being impersonated
      *
-     * @param static|false $impersonator The user attempting to impersonate this user, false when not available
+     * @param \Winter\Storm\Auth\Models\User|false $impersonator The user attempting to impersonate this user, false when not available
      * @return boolean
      */
     public function canBeImpersonated($impersonator = false)
@@ -254,11 +254,18 @@ class User extends UserBase
         if (
             $this->isSuperUser() ||
             !$impersonator ||
+            !($impersonator instanceof static) ||
             !$impersonator->hasAccess('backend.impersonate_users') ||
             $impersonator === $this
         ) {
             return false;
         }
+
+        // Clear the merged permissions before the impersonation starts
+        // so that they are correct even if they had been loaded prior
+        // to the impersonation starting
+        $this->mergedPermissions = null;
+
         return true;
     }
 }
--- a/modules/backend/models/User.php
+++ b/modules/backend/models/User.php
@@ -243,14 +243,20 @@ class User extends UserBase
     /**
      * Check if this user can be impersonated by the provided impersonator
      * Super users cannot be impersonated and all users cannot be impersonated unless there is an impersonator
-     * present and the impersonator has access to `backend.impersonate_users`.
+     * present and the impersonator has access to `backend.impersonate_users`, and the impersonator is not the
+     * user being impersonated
      *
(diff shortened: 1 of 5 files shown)


modules/backend/models/User.php
```diff
@@ -209,4 +209,56 @@ public function unsuspend()
...
+     * user being impersonated
+     *
+     * @param static|false $impersonator The user attempting to impersonate this user, false when not available
+     * @return boolean
+     */
+    public function canBeImpersonated($impersonator = false)
+    {
+        if (
+            $this->isSuperUser() ||
+            !$impersonator ||
```
bennothommo: Instead of `false`, should this be `null`? Better describes an absence of an impersonator.

```suggestion
     * @param static|null $impersonator The user attempting to impersonate this user, `null` when not available
     * @return boolean
     */
    public function canBeImpersonated($impersonator = null)
    {
        if (
            $this->isSuperUser() ||
            is_null($impersonator) ||
```

bennothommo: Might also be worth having a method existence check for the `hasAccess` method if the impersonator is provided.

LukeTowers: No because null will screw with Session::has()

LukeTowers: hasAccess() is present on the base user model instance already

bennothommo: Fair enough, then the property should be type-hinted to the base user model instead. Given impersonation is supposed to be between two "user" models, we should be strict on what is allowed to be compared for impersonation.

LukeTowers: You mean instead of static?

bennothommo: I mean `Winter\Storm\Auth\Models\User`, which every user model should be extended from.

LukeTowers: Can you submit the your idea as a suggestion?

bennothommo: Sorry, I should be more clear. So I think the base `canBeImpersonated` method in `Winter\Storm\Auth\Models\User` should have that model as the type hint, which would enforce the same type-hint on the User models in the Backend module and the User plugin. If the intention is interoperability between models, it should be that the method allows any type of user.

bennothommo: We'd then need to restrict the checks on the final models whether they will allow interoperability or not. In the case of the Backend user model, I would imagine that it can only be impersonated by other Backend users of a specific level of access, so you would need to do a check within its `canBeImpersonated` method that the given impersonatee is a Backend user too.

modules/backend/models/User.php
```diff
@@ -209,4 +209,56 @@ public function unsuspend()
...
+    /**
+     * Returns an array of merged permissions based on the user's individual permissions
+     * and their group permissions filtering out any permissions the impersonator doesn't
+     * have access to (if the current user is being impersonated)
+     *
+     * @return array
+     */
+    public function getMergedPermissions()
+    {
+        if (!$this->mergedPermissions) {
```
bennothommo, line 226: @LukeTowers we'll need to make sure that the merged permissions of an impersonatee are cleared immediately on impersonation, on the off-chance that some functionality is fired immediately after impersonation, before the model has had a chance to refresh on, say, a page refresh.

LukeTowers: Good catch!

modules/backend/models/User.php
LukeTowers: Makes sense

PR #295, opened 2021-08-30
Allow quick development environment with Gitpod

bennothommo:
This PR introduces the capability of creating a quick development environment of Winter CMS using [Gitpod](https://www.gitpod.io) for testing and further development.

Gitpod is a cloud-based service that creates development environments direct from a Git repository. Each development environment, at its base, contains the VSCode editor direct in the browser, with support for multiple languages. It can then be extended to create a custom environment with Docker and initialisation callbacks.

I have implemented such an environment so that Winter CMS is completely installed, migrated and operational when starting a new instance in Gitpod. It does the following:

- Installs the PHP Xdebug extension and Composer 2.
- Installs the [MailHog](https://github.com/mailhog/MailHog/) service for capturing and reviewing emails.
- Creates a MySQL database.
- Installs all Composer dependencies.
- Rewrites the environment configuration to work within Gitpod.
- Runs all database migrations.
- Sets admin account to `admin / admin` for brevity.
- Serves the site on port 8000, and the Mailhog interface on port 8025. The site will be publicly accessible.

Gitpod will work for pull requests and other branches, as well as the main `develop` branch. You simply need to create an account on Gitpod - this is easy as you can simply use your GitHub account to register and login. For PRs, it will provide a link in the comments to open this PR in GitHub immediately. You can also install extensions in Chrome and Firefox to provide a button direct on the GitHub interface to quickly open up a branch or PR in Gitpod.

**Note:** Gitpod offers 50 hours per month free. Any more will require a paid subscription. This should be enough for most testing and small development purposes, but best to keep it in mind when considering this PR.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/295"></a>


LukeTowers:
Cool! Few questions, is xdebug required and or actually helpful or would it be better to just include the debugbar plugin? Second how / are we able to use this to provide one click demos of a fully functional winter experience? Third, can we also use this on our plugin repositories? Fourth, is mysql fine for this or would sqlite perform better / require less setup and server resources?

bennothommo:
@LukeTowers 

> is xdebug required and or actually helpful or would it be better to just include the debugbar plugin?

Personally, I think it's necessary. If we need to find out why something breaks, it's much easier to use Xdebug. In saying that, DebugBar would probably be good to include too in the default build.

> how / are we able to use this to provide one click demos of a fully functional winter experience?

We can provide a button like the one in my initial comment which serves as a "one-click" approach, but everyone who uses it would need a Gitpod account - we can't offer a public one that works without authentication. After registration however, it'd be essentially one-click.

The Workshop theme I demonstrated to you the other day is working towards a full Winter CMS experience. I would imagine we could install it and other common plugins to provide a better base for people to jump start a project.

> Third, can we also use this on our plugin repositories?

Yep, we'd just need to change the initialization steps slightly to install Winter CMS itself first, but certainly very doable.

> Fourth, is mysql fine for this or would sqlite perform better / require less setup and server resources?

I figured MySQL would be best as its probably our most common setup. I'm not overly concerned on resources - this is using Gitpod's cloud service, so they can worry about the resources :)

PR #303, opened 2021-09-08
Add missing filter translations

datune:
<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/303"></a>

### DIFF
--- a/modules/system/lang/de/client.php
+++ b/modules/system/lang/de/client.php
@@ -61,7 +61,7 @@ return [
 
     'filter' => [
         'group' => [
-            'all' => 'all',
+            'all' => 'Alle',
         ],
         'scopes' => [
             'apply_button_text' => 'Anwenden',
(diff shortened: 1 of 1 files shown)


modules/system/lang/de/client.php
```diff
@@ -61,7 +61,11 @@
 
     'filter' => [
         'group' => [
-            'all' => 'Alle'
+            'all' => 'all',
```
bennothommo: @datune Is this meant to be "alle"? I'm probably wrong, but just wanted to check :)

Would you also mind changing all "Alle" references in other `all` definitions, as these are all supposed to be lower-case?

datune: You‘re indeed wrong 😉 It is supposed to be like that.

bennothommo: I sometimes am 😂 

Do you mind doing this for us?:
> Would you also mind changing all "Alle" references in other all definitions, as these are all supposed to be lower-case?

datune: Thanks for catching that one, I've updated the PR!

bennothommo:
Thanks @datune :)

PR #304, opened 2021-09-09
Added Latvian translations for Allowed IP messages

arvislacis:
<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/304"></a>

### DIFF
--- a/modules/cms/lang/lv/lang.php
+++ b/modules/cms/lang/lv/lang.php
@@ -105,8 +105,8 @@ return [
             'description' => 'IP adreses, kurām ir atļauts skatīt vietni, kamēr uzturēšanas režīms ir aktīvs',
             'prompt' => 'Pievienot IP adresi',
             'ip' => 'IP adrese',
-            'label' => 'Apraksts'
-        ]
+            'label' => 'Apraksts',
+        ],
     ],
     'page' => [
         'not_found_name' => "Lapa ':name' netika atrasta",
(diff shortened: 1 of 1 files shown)


modules/cms/lang/lv/lang.php
```diff
@@ -100,6 +100,13 @@
         'is_enabled' => 'Ieslēgt uzturēšanas režīmu',
         'is_enabled_comment' => 'Izvēlieties lapu, kura tiks attēlota, kad uzturēšanas režīms ir aktīvs.',
         'hint' => 'Uzturēšanas režīms attēlos izvēlēto uzturēšanas lapu apmeklētājiem, kuri nav pierakstījušies back-end sistēmā.',
+        'allowed_ips' => [
+            'name' => 'Atļautās IP adreses',
+            'description' => 'IP adreses, kurām ir atļauts skatīt vietni, kamēr uzturēšanas režīms ir aktīvs',
+            'prompt' => 'Pievienot IP adresi',
+            'ip' => 'IP adrese',
+            'label' => 'Apraksts'
+        ]
```
bennothommo: @arvislacis Just for future reference, we put commas at the end of arrays too. :)
```suggestion
            'label' => 'Apraksts',
        ],
```

arvislacis: @bennothommo Thanks. Yeah, I was thinking and know this..., the problem is the English version has the same issue: https://github.com/wintercms/winter/blob/b1fdf0745a65d91224acb100f4c2b64ce27fe356/modules/cms/lang/en/lang.php#L108

bennothommo: If you spot those issues in the English translation too, feel free to include a fix for them. We generally fix them as we spot them.

PR #105, opened 2021-04-22
Added format option to allow color picker widget to return alternate color formats

webbati:
Added format option to allow color picker widget to return alternate color formats.
Furthermore, now it is also possible to use the slider for the alpha channel for transparency.

https://github.com/wintercms/winter/issues/94

PR #324, opened 2021-09-27
Redesigned color picker widget

### LINKED ISSUE #94: Allow color picker widget to return alternate color formats
Per https://github.com/wintercms/winter/discussions/90,

The Spectrum color picker widget in use with Winter CMS supports exporting of various types of color formats (HSL, RGB, hex, etc.). We could potentially have an option with the widget to allow an alternate color format to be returned and used for certain fields - perhaps `format` - with it defaulting to `hex`.

### LINKED ISSUE #321: Pressing Enter after manually entering a colour in the Color Picker widget saves the entire form
### Winter CMS Build

1.1

### PHP Version

7.4

### Database engine

SQLite

### Plugins installed

All of them!

### Issue description

To manually enter and select a color in the color picker, one must select the text field at the bottom of the widget and enter in the value, but to actually select it, you have to hit the Enter key. Doing this, however, saves the form.

It would appear that the ENTER key is not captured by the color picker in this context.

### Steps to replicate

See issue description.

### Workaround

_No response_

### Any further details?

_No response_

bennothommo:
This PR introduces a newly redesigned color picker widget, built on top of the [Pickr UI widget](https://github.com/Simonwep/pickr) as opposed to Spectrum, which has not seen a release [since 2016](https://github.com/bgrins/spectrum/releases/tag/1.8.0).

image

### Benefits

- More clean and streamlined UX - available colors are now shown within the popover as opposed to spanning across the entire row at all times. The UI is very similar to VSCode's color picker, which should hopefully encourage familiarity.
- Text input next to the color picker allows full copy+paste capabilities, as well as simply typing out the color code.
- Supports touch screen devices.
- Full accessibility and internationalisation support.
- Full support for other color formats (HEX / RGB / HSL / CMYK) - this PR would replace https://github.com/wintercms/winter/pull/105 and fix https://github.com/wintercms/winter/issues/94
- Will also fix https://github.com/wintercms/winter/issues/321
- Will properly clean up and dispose of the widget as needed.

### New config options

Option | Description
------------- | -------------
**formats** | Specifies the color format(s) to store. Can be a string or an array of values `cmyk`, `hex`, `hsl` and `rgb`. Specifying `all` as a string will allow all formats. Defaults to `hex`.
**allowCustom** | If `false`, only the colors specified in `availableColors` will be available for selection. The color picker palette will be disabled. Default `true`.

### Notes

- Given that we've dropped support for IE11 in the Backend, I've used some ES6 code (just arrow functions, really) to resolve some scoping issues in the JS, and the picker itself uses CSS Grid Layout. Both features, AFAICT, have been supported by all browser versions within the last 4-5 years, so I don't think we're leaving anyone behind with this.
- This should be fully compatible with values from the old color picker, since they would've been stored as hex values. I also believe that it should be able to convert any value to another format if someone opts to change the format in the config, but I have not tested this yet.
- I'll post up a video of this tomorrow, but feel free to have a play around with it on Gitpod in the mean time. Go to **Settings > Customize back-end > Colors** to get a couple of color picker widgets to play with.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/324"></a>


LukeTowers:
Something seems off about the font, is that the default font? Maybe we should use a mono space font instead

bennothommo:
@LukeTowers yeah it's the default body font. I'll change it to monospace.

LukeTowers:
Could we also provide `format: any` to allow the user to pick which format they want to use?

bennothommo:
@LukeTowers multiple format support has been added.

LukeTowers:
Is there a docs PR for the changes?

bennothommo:
@LukeTowers not yet - I'll put one through shortly. Any further thoughts on this, as it is ready to merge?

modules/backend/formwidgets/colorpicker/partials/_colorpicker.htm
```diff
@@ -1,42 +1,33 @@
...
+    class="field-colorpicker <?php if($readOnly || $disabled || $this->previewMode): ?>disabled<?php endif; ?>"
+    data-control="colorpicker"
+    data-formats="<?= e(json_encode($formats)) ?>"
+    data-available-colors="<?= e(json_encode($availableColors)) ?>"
+    <?php if ($showAlpha): ?>data-show-alpha="<?= $showAlpha ?>"<?php endif ?>
+    <?php if ($allowEmpty): ?>data-allow-empty="<?= $allowEmpty ?>"<?php endif ?>
+    <?php if ($allowCustom): ?>data-allow-custom="<?= $allowCustom ?>"<?php endif ?>
+    data-data-locker="#<?= $this->getId('input') ?>"
+    <?php if ($readOnly || $disabled || $this->previewMode): ?>data-disabled="true"<?php endif; ?>
+    <?= $this->formField->getAttributes() ?>>
```
LukeTowers:
```suggestion
    <?= $this->formField->getAttributes() ?>
>
```

modules/backend/formwidgets/colorpicker/assets/js/colorpicker.js
```diff
@@ -9,102 +9,399 @@
...
+        $.wn.foundation.controlUtils.markDisposable(element)
+        Base.call(this)
+
         // Init
         this.init()
     }
 
     ColorPicker.DEFAULTS = {
         showAlpha: false,
+        format: 'hexa',
```
LukeTowers: Shouldn't this default to the same default as the PHP?

modules/backend/models/BrandSetting.php
```diff
@@ -79,6 +80,68 @@ public function initSettingsData()
...
+            [
+                'color' => '#bdc3c7',
+            ],
+            [
+                'color' => '#95a5a6',
+            ],
+            [
+                'color' => '#7f8c8d',
+            ],
+        ]);
```
LukeTowers, line 144: Why does this format have to be an array of arrays instead of just an array of values?

modules/backend/formwidgets/colorpicker/assets/js/colorpicker.js
bennothommo: Good spot.

modules/backend/models/BrandSetting.php
bennothommo: So it works in the repeater widget :)

modules/backend/formwidgets/ColorPicker.php
```diff
@@ -91,13 +99,14 @@ public function render()
...
-        $this->vars['disabled'] = $this->disabled;
-        $this->vars['isCustomColor'] = !in_array($value, $availableColors);
+        $this->vars['value'] = $this->getLoadValue();
+        $this->vars['availableColors'] = $this->getAvailableColors();
+        $this->vars['formats'] = $this->getFormats();
+        $this->vars['allowEmpty'] = (bool) $this->allowEmpty;
+        $this->vars['allowCustom'] = (bool) $this->allowCustom;
+        $this->vars['showAlpha'] = (bool) $this->showAlpha;
+        $this->vars['readOnly'] = (bool) $this->readOnly;
+        $this->vars['disabled'] = (bool) $this->disabled;
```
Mrkbingham, line 109: I've noticed when setting $field->disabled on the old colorpicker in a model's filterFields function it doesn't inherit the disabled state.  Is that something that has been / could be fixed in this update?

bennothommo: I believe this is working, but will need to double check.

PR #328, opened 2021-10-06
Get parent model from its child relation

xyz1123581321:
<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/328"></a>

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -360,21 +360,22 @@ class RelationController extends ControllerBehavior
          * Relationship Parent
          */
 
-        $parent_classname = "";
+        $parentClass = get_class($this->model);
+        $allRelations = $this->relationModel->getRelationDefinitions();
 
-        foreach($all_relations as $relations)
+        $parentClassName = "";
+
+        foreach($allRelations as $relations)
         {
             foreach($relations as $key => $value)
             {
-                if (strtolower($value[0]) == strtolower($parent_class))
-                {
-                    $parent_classname = $key;
-                }
+                if (strtolower($value[0]) == strtolower($parentClass))
+                    $parentClassName = $key;
             }
         }
 
-        if ($parent_classname != "")
-              $this->relationModel->{$parent_classname} = $this->model;
+        if ($parentClassName != "")
+              $this->relationModel->{$parentClassName} = $this->model;
 
 
         $this->manageId = post('manage_id');
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -356,6 +356,27 @@ class RelationController extends ControllerBehavior
         $this->relationObject = $this->model->{$field}();
         $this->relationModel = $this->relationObject->getRelated();
 
+        /*
+         * Relationship Parent
+         */
+
+        $parent_classname = "";
+
+        foreach($all_relations as $relations)
+        {
+            foreach($relations as $key => $value)
+            {
+                if (strtolower($value[0]) == strtolower($parent_class))
+                {
+                    $parent_classname = $key;
+                }
+            }
+        }
+
+        if ($parent_classname != "")
+              $this->relationModel->{$parent_classname} = $this->model;
+
+
         $this->manageId = post('manage_id');
         $this->foreignId = post('foreign_id');
         $this->readOnly = $this->getConfig('readOnly');
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -215,6 +215,8 @@ class RelationController extends ControllerBehavior
      */
     public $deferredBinding = false;
 
+
+
     /**
      * Behavior constructor
      * @param Backend\Classes\Controller $controller
@@ -317,6 +319,7 @@ class RelationController extends ControllerBehavior
 
         $this->config = $this->originalConfig;
         $this->model = $model;
+
         $this->field = $field;
 
         if ($field == null) {
@@ -357,20 +360,21 @@ class RelationController extends ControllerBehavior
         $this->relationModel = $this->relationObject->getRelated();
 
         /*
-         * Relationship Parent
+         * Parent relationship
          */
 
         $parentClass = get_class($this->model);
-        $allRelations = $this->relationModel->getRelationDefinitions();
-
+        $allRelationDefinitions = $this->relationModel->getRelationDefinitions();
         $parentClassName = "";
 
-        foreach($allRelations as $relations){
-            foreach($relations as $key => $value){
+        foreach ($allRelationDefinitions as $relations){
+            foreach ($relations as $key => $value){
                 if (strtolower($value[0]) == strtolower($parentClass)){
                     $parentClassName = $key;
                 }
+
             }
+
         }
 
         if ($parentClassName != ""){
@@ -678,7 +682,6 @@ class RelationController extends ControllerBehavior
     protected function makeViewWidget()
     {
         $widget = null;
-
         /*
          * Multiple (has many, belongs to many)
          */
@@ -1112,6 +1115,7 @@ class RelationController extends ControllerBehavior
         $this->beforeAjax();
         $saveData = $this->manageWidget->getSaveData();
         $sessionKey = $this->deferredBinding ? $this->relationGetSessionKey(true) : null;
+
 
         if ($this->viewMode == 'multi') {
             $newModel = $this->relationModel;
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -365,18 +365,19 @@ class RelationController extends ControllerBehavior
 
         $parentClassName = "";
 
-        foreach($allRelations as $relations)
-        {
-            foreach($relations as $key => $value)
-            {
-                if (strtolower($value[0]) == strtolower($parentClass))
+        foreach($allRelations as $relations){
+
+            foreach($relations as $key => $value){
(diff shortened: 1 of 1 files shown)


RomainMazB:
Is it related to any issue?
Which problem or situation are you trying to solve or feature are you trying to implement?

modules/backend/behaviors/RelationController.php
```diff
@@ -356,6 +356,28 @@ public function initRelation($model, $field = null)
         $this->relationObject = $this->model->{$field}();
         $this->relationModel = $this->relationObject->getRelated();
 
+        /*
+         * Relationship Parent
```
RomainMazB:
```suggestion
         * Parent relationship
```

modules/backend/behaviors/RelationController.php
```diff
@@ -356,6 +356,28 @@ public function initRelation($model, $field = null)
...
+         */
+
+        $parentClass = get_class($this->model);
+        $allRelations = $this->relationModel->getRelationDefinitions();
+
+        $parentClassName = "";
+
+        foreach($allRelations as $relations)
```
RomainMazB, line 368: Some codestyle issues in this block: 
- Opening bracket should be on the same line on the `foreach` loops
- Missing opening and closing brackets on the single line `if` conditions

modules/backend/behaviors/RelationController.php
```diff
@@ -356,6 +356,28 @@ public function initRelation($model, $field = null)
...
+        {
+            foreach($relations as $key => $value)
+            {
+                if (strtolower($value[0]) == strtolower($parentClass))
+                    $parentClassName = $key;
+            }
+        }
+
+        if ($parentClassName != "")
+              $this->relationModel->{$parentClassName} = $this->model;
```
RomainMazB: Some codestyle issues in this block:
- Opening bracket should be on the same line on the foreach loops
- Missing opening and closing brackets on the single line if conditions

modules/backend/behaviors/RelationController.php
xyz1123581321: Fixed

modules/backend/behaviors/RelationController.php
xyz1123581321: Fixed

xyz1123581321:
We talked about this small improvement on discord with Luke.

RomainMazB:
Ok, just for prosperity I post here the issue from the OC forum you are trying to solve. Just for others who don't know like me what's going on here :)
> I have 2 models: Item and ItemPrice
> Item hasMany ItemPrices
> ItemPrice belongsTo Item
> 
> When I create new ItemPrice from Item update form (using relation controller), I need to access Item model (i.e parent model) when generation dropdown options in related model form.
> Something like $this->item->...
> But relation controller doesn't set relation to parent model before related model is added.

xyz1123581321:
Did you review it?

modules/backend/behaviors/RelationController.php
```diff
@@ -215,6 +215,8 @@ class RelationController extends ControllerBehavior
      */
     public $deferredBinding = false;
 
+
+
```
LukeTowers:
```suggestion
```

modules/backend/behaviors/RelationController.php
```diff
@@ -317,6 +319,7 @@ public function initRelation($model, $field = null)
 
         $this->config = $this->originalConfig;
         $this->model = $model;
+
```
LukeTowers:
```suggestion
```

modules/backend/behaviors/RelationController.php
```diff
@@ -356,6 +359,28 @@ public function initRelation($model, $field = null)
...
+                    $parentClassName = $key;
+                }
+
+            }
+
+        }
+
+        if ($parentClassName != ""){
+              $this->relationModel->{$parentClassName} = $this->model;
+        }
```
LukeTowers:
```suggestion
        // Ensure that the other side of the relationship is set on the related model
        $parentClass = get_class($this->model);
        $allRelationDefinitions = $this->relationModel->getRelationDefinitions();
        $inverseRelation = null;
        foreach ($allRelationDefinitions as $relations) {
            foreach ($relations as $key => $value) {
                if (strtolower($value[0]) === strtolower($parentClass)) {
                    $inverseRelation = $key;
                    break;
                }
            }
        }

        if (!empty($inverseRelation)){
            $this->relationModel->{$inverseRelation} = $this->model;
        }
```

modules/backend/behaviors/RelationController.php
```diff
@@ -657,7 +682,6 @@ protected function makeSearchWidget()
     protected function makeViewWidget()
     {
         $widget = null;
-
```
LukeTowers, line 660: Needs to have whitespace added back in

modules/backend/behaviors/RelationController.php
```diff
@@ -1092,6 +1116,7 @@ public function onRelationManageCreate()
         $saveData = $this->manageWidget->getSaveData();
         $sessionKey = $this->deferredBinding ? $this->relationGetSessionKey(true) : null;
 
+
```
LukeTowers:
```suggestion
```

LukeTowers (changes requested):

modules/backend/behaviors/RelationController.php
```diff
@@ -356,6 +356,23 @@ public function initRelation($model, $field = null)
...
+        $allRelationDefinitions = $this->relationModel->getRelationDefinitions();
+        $inverseRelation = null;
+        foreach ($allRelationDefinitions as $relations) {
+            foreach ($relations as $key => $value) {
+                if (strtolower($value[0]) === strtolower($parentClass)) {
+                    $inverseRelation = $key;
+                    break;
+                }
+            }
+        }
```
LukeTowers, line 370: This seems a bit fragile to me, I hadn't realized the complexity of setting the inverse relationship. @bennothommo do you have any ideas for this?

Right now I'm concerned about how well it supports all of the relationship modes supported by the RelationController and also if the related model has multiple relationship definitions tying back to the original parent model, how would we support that / should we try to support that? I'm thinking our best bet for that issue is to explicitly detect any more than one possible inverse relationship and don't do anything if that's encountered. It's already a bit presumptuous to be setting the inverse relationship directly when it might have scopes or conditions limiting the current parent model from being set in the first place.

Or perhaps I'm just overthinking this all, @bennothommo any thoughts on the above?

modules/backend/behaviors/RelationController.php
xyz1123581321: I didn't change anything at 660...
Only 362 to 382.

```
         /*
         * Parent relationship
         */

        $parentClass = get_class($this->model);
        $allRelationDefinitions = $this->relationModel->getRelationDefinitions();
        $parentClassName = "";

        foreach ($allRelationDefinitions as $relations) {
            foreach ($relations as $key => $value) {
                if (strtolower($value[0]) == strtolower($parentClass)) {
                    $parentClassName = $key;
                } 

            } 

        } 

        if ($parentClassName != "") {
              $this->relationModel->{$parentClassName} = $this->model;
        }
```

modules/backend/behaviors/RelationController.php
LukeTowers: @bennothommo ping

bennothommo: Sorry for the extreme delay in responding to this one!

It is somewhat flimsy, but at the same time - conventionally speaking - we should be alright. I can't think of a better way right now, and even if I could, there's no telling whether it would work or not given our relation definition is completely different to Laravel's.

bennothommo:
@LukeTowers anything else required here?

mjauvin:
The way I do this personally is this in the related model constructor:
```php
public function __construct(array $attributes = [])
{
    parent::__construct($attributes);
    $this->bindEvent('model.form.filterFields', function ($formWidget, $fields, $context) {                                              
        if ($context === 'create') {
            if ($model = $formWidget->model) {
                $model->parent = $formWidget->getController()->relationObject->getParent();
            }
        }
    });
}
```

mjauvin:
I do this when I need the parent:
```php
public function __construct(array $attributes = [])
    {                                              parent::__construct($attributes);          $this->bindEvent('model.form.filterFields', function ($formWidget, $fields, $context) {                                              if ($context === 'create') {                   if ($model = $formWidget->model) {
                    $model->parent = $formWidget->getController()->relationObject->getParent();                                          }                                      }
```

LukeTowers:
I'm concerned about knock-on effects from making this sort of change, is it possible to simplify the problem down to a small demo of the issue that can have test cases documenting the expectations and then verify the solution from there or is it too complicated with how it integrates with the relationcontroller functionality?

Either way I'll need to spend some time on this to refresh my memory on what the original problem was as well as how this solved it and my original concerns. Is there still interest in getting this merged or is @mjauvin's proposed workaround decent enough for anyone else encountering this sort of bump in the road?

mjauvin:
I just use the following:
```
$formwidget->getController()->relationObject->getParent();
```

bennothommo:
Let's close this for now, given there are available workarounds. If someone wishes to take charge of this and get this over the line, feel free to drop a line here and we'll re-open.

LukeTowers:
Posting here as I just ran into this myself:

```php
public function getOptionsOptions(): array
{
    return $this->tournament->options()->pluck('name', 'id')->all();
}
```

causes the creation of related records through the relation controller to break as the parent relationship (`tournament`) is not automatically set by the RelationController in the create context.

@mjauvin's workaround worked great though.

PR #354, opened 2021-11-25
Implement an opt-in improved JS framework (Snowboard)

bennothommo:
**REPLACED BY https://github.com/wintercms/winter/pull/401**

Documentation: https://github.com/wintercms/docs/pull/45

This PR moves towards introducing a new JavaScript framework for Winter, that uses modern JavaScript and removes the hard-dependency to jQuery that has traditionally been the case with Winter and October before it.

If accepted, we intend to make this **opt-in**, and still provide the option of using the original framework for legacy apps and older browser support.

### Goals

- Rewrite the framework to be modular and give the user full control of what features are to be enabled in the framework, and not impose technical opinions on the developer.
- Use a vanilla JavaScript approach to remove the need for jQuery and decrease the asset bandwidth considerably.
- Leverage the power of ES6 classes to allow for inherited code, extensions and bring the code style more in-line with our PHP code.
- Take advantage of the Node.js ecosystem to provide minification and compatibility with a wide variety of browsers using Babel.
- Still provide exactly (or as close to as possible) the same functionality that the original framework provides, with a minimum need for rewriting established projects.

### Concepts

The four cornerstones of this framework are as follows:

#### The `Snowboard` class
Represents the global container for the application. This stores all activated plugins and provides instances of these modules as required. It is synonymous with the `Application` class of Laravel, or a `Vue` instance.

#### The `PluginLoader` class
The main conduit between a plugin and the Snowboard app. This provides the mechanism for delivering instances of each module (or a single instance of Singleton plugin) and resolves dependencies.

#### The `PluginBase` class
A plugin is a specific extension to the Snowboard application. It can contain code that provides new functionality or augments other functionality already registered in the application. A plugin can be reused - in essence, each call to the plugin will create a new instance and will be independent from other instances. This would be used for scenarios such as Flash messages, AJAX requests and the like.

#### The `Singleton` class
An extension of the `PluginBase` class. It signifies to the application that this plugin should exist once, and all uses of this plugin should use the same instance. This would be useful for things like event handlers, extending functionality of another plugin or global functionality.

A singleton is initialized automatically when the DOM is ready, if it is called directly or if it listens to an event that is fired. The singleton instance is then used no matter how many times it is called from the Snowboard class.

### What's included

By default, this framework only includes the framework architecture and a couple of utilities. This is an intentional choice to allow the greatest flexibility in how the framework is used. The utilities include:

- A JSON parser that is customized to handle the established data structure of Winter (and October before it).
- An HTML sanitizer that provides sanitization for HTML code.
- A debouncer, that acts similar to the `Vue.newTick` function in allowing code to run after the previous frame is rendered - useful for ensuring DOM changes are made and events are correctly fired.

In addition, some opt-in features include a rewritten `Request` class that represents a fresh take on the JavaScript AJAX framework available in Winter. This new `Request` class leverages the `fetch` method in all modern browsers to conduct AJAX calls. It supports all options and configuration that were previously available in the original framework.

A second extension - `AttributeRequest` - provides an augmentation to this functionality and allow the AJAX framework to also handle AJAX calls that use the `data-` attributes. By separating these features, the developer has the ability to turn off support for the `data-` attribute AJAX framework should they not need it.

### What's not included

- The HTML data attribute framework allowed for arbitrary JavaScript to be called from within `data-` attribute callbacks (eg. `data-request-success`, `data-request-error` etc.). This feature, at a minimum, prevented sites using content security policies because of its use of the `eval()` method in JavaScript - at worst, it presented a potential vector for naughtiness. We will not be including this in the framework.

### Further notes

- Due to the usage of Promises, ES6 classes and ES6 syntax, this framework *does not* support any version of Internet Explorer. It however is supported by all browsers available since 2017 (except for, I suppose, Opera Mini or the Samsung Browser).
- Node.js will only be required for development of the framework. There will be no need for Node.js usage in themes, plugins, components, etc. (of course, you'll be free to use it in your own development pipeline if you want).
- More details will be forthcoming as the framework solidifies.

### To-do

- [ ] Documentation ([wintercms/docs](https://github.com/wintercms/docs/pull/45))
- [x] Unit tests
- [x] Finish off all abilities of the HTML data attributes framework
- [x] Full testing ([test cases added to Workshop theme](https://github.com/wintercms/wn-workshop-theme))

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/354"></a>


LukeTowers:
A couple immediate quick thoughts, `EnableDataRequests` should probably just be `AttributeRequests` or something similar, `Module` doesn't seem like the best term to use, especially within the context of Winter as a whole where `Modules` are already a distinct documented thing that's completely different than this. Additionally, I'm not 100% sure I'm sold the idea of people mixing and matching what "modules" they want to use between the requests, jsonparsing, serialization, and data requests, could you expand on the rationale for that?

Configuration is good in general but too much configuration unnecessarily can have negative impacts on usability and maintainability.

bennothommo:
@LukeTowers thanks for the feedback.

Happy to change it to `AttributeRequests`.

Do you have another term you would use instead of "module"? Extensions, perhaps?

The JSON parser and Sanitizer functionality will always be included, they aren't optional. It's only the Request and AttributeRequests classes that are optional. The rationale for this is that there are people who don't use the data attribute framework at all. I've had several themes where I use the JS framework and that's it. I believe this would also be common for people who use other JS frameworks, such as Vue or React etc.

LukeTowers:
I'm not sure on the terminology to use just yet, what do other frameworks / libraries call similar features (i.e. Vue, React, Angular, Livewire, Alpine, Svelte, etc)? Maybe plugins? I'm not really sure just yet, I might need to play around with it more personally to get a feel for how it works.

As an aside, while someone might not want to use the data attributes I feel like the core of the winter.js (side note I think we should go with that and rename the existing backend winter.js to backend.js) is to use the AJAX requests functionality, so I'm not sure the Requests module itself should be able to be toggled off either. How does framework.extras fit into this new framework as well?

bennothommo:
The common terminology in most of the frameworks seems to be "component", so perhaps that might be the best term?

> As an aside, while someone might not want to use the data attributes I feel like the core of the winter.js (side note I think we should go with that and rename the existing backend winter.js to backend.js) is to use the AJAX requests functionality, so I'm not sure the Requests module itself should be able to be toggled off either. How does framework.extras fit into this new framework as well?

My intention is to set up the mix to do three builds - one with no request and no data attribute component, one with just the request component and one with both. I still need to figure out the Twig tag, but I intend to make it as simple as the `{% framework %}` tag is now. But it is a goal of this to not *force* either component. I feel someone should have the ability to extend the Request class and implement their own - perhaps to replace `fetch()` with the Axios library, or implement their own front-end library to make the partial updates.

I'm going to put up some examples soon so you can see what I mean. I can understand it's probably difficult to see how it works from code alone, so hopefully they'll more thoroughly demonstrate what I'm intending to allow.

bennothommo:
### Notes on the new `Request` class

- Will be called with `winter.request()` as opposed to `$.request` or `$(element).request()`.
- Parameters for `winter.request()`
  - `element`: The element that this request is targeting. Should be a UI element - ie. a form, input, button or link.
  - `handler`: The AJAX handler to call. As previously, this should have the format "on[Event]".
  - `options`: The configuration of the request.
- Available options remain the same as specified in the old framework (https://wintercms.com/docs/ajax/javascript-api#javascript-api)
- Events are now [normal DOM events](https://developer.mozilla.org/en-US/docs/Web/API/Event), which can be targeted by both vanilla JS and jQuery.  This means that event data is now stored in an Event class, and is not available as callback arguments like the old system. This will be detailed in an "Upgrade Guide" that will form part of the docs. Global events are also changed to use the `Winter` class event dispatcher, and will be extended/listened with modules added to the Winter JS framework. This is probably the biggest breaking change, and cannot be avoided due to jQuery's unique events system.
- Normal [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) objects are being used for all deferred/asynchronous functionality, as opposed to jQuery's `deferred` system. This is a smaller breaking change, as deferred functionality was only available through the `ajaxPromise` event previously, but I've noted it down to be documented.

modules/system/assets/js/framework-next/abstracts/Module.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Module abstract.
+ *
+ * This class provides the base functionality for all modules.
+ *
+ * @copyright 2021 Winter.
+ * @author Ben Thomson <git@alfreido.com>
+ */
+export default class Module {
```
LukeTowers: Let's make it PluginBase to match the PHP side

modules/cms/twig/Extension.php
```diff
@@ -77,6 +77,7 @@ public function getTokenParsers()
             new FlashTokenParser,
             new ScriptsTokenParser,
             new StylesTokenParser,
+            new WinterJsTokenParser,
```
LukeTowers:
```suggestion
            new SnowboardTokenParser,
```

modules/system/assets/less/framework.extras.less
```diff
@@ -71,68 +71,68 @@ body.oc-loading, body.oc-loading * {
...
-        outline: none;
-
-        &:hover,
-        &:focus {
-            color: white;
-            text-decoration: none;
-            cursor: pointer;
-            .opacity(.5);
-        }
-    }
```
LukeTowers: doesn't this break the existing framework extras CSS? We should probably have a separate CSS file for Snowboard extras

modules/cms/twig/WinterJsNode.php
```diff
@@ -0,0 +1,54 @@
...
+        $compiler
+            ->addDebugInfo($this)
+            ->write("\$_minify = ".CombineAssets::class."::instance()->useMinify;" . PHP_EOL);
+
+        $moduleMap = [
+            'request' => 'framework-js-request',
+            'attr' => 'framework-attr-request',
+            'extras' => 'framework.extras',
+        ];
+        $basePath = Request::getBasePath() . '/modules/system/assets/js/framework-next/build/';
```
LukeTowers:
```suggestion
        $basePath = Request::getBasePath() . '/modules/system/assets/js/snowboard/build/';
```

LukeTowers: Also re: our conversation this should support switching between a "debug" and "prod" build based on app.debug server config.

modules/cms/twig/WinterJsTokenParser.php
```diff
@@ -0,0 +1,60 @@
...
+use Twig\Token as TwigToken;
+use Twig\TokenParser\AbstractTokenParser as TwigTokenParser;
+
+/**
+ * Parser for the `{% winterjs %}` Twig tag.
+ *
+ * @package winter\wn-cms-module
+ * @author Winter CMS
+ */
+class WinterJsTokenParser extends TwigTokenParser
```
LukeTowers:
```suggestion
class SnowboardTokenParser extends TwigTokenParser
```

modules/cms/twig/WinterJsTokenParser.php
```diff
@@ -0,0 +1,60 @@
+<?php namespace Cms\Twig;
+
+use Twig\Token as TwigToken;
+use Twig\TokenParser\AbstractTokenParser as TwigTokenParser;
+
+/**
+ * Parser for the `{% winterjs %}` Twig tag.
```
LukeTowers:
```suggestion
 * Parser for the `{% snowboard %}` Twig tag.
```

modules/cms/twig/WinterJsTokenParser.php
```diff
@@ -0,0 +1,60 @@
...
+
+        return new WinterJsNode($modules, $lineno, $this->getTag());
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getTag()
+    {
+        return 'winterjs';
```
LukeTowers:
```suggestion
        return 'snowboard';
```

modules/cms/twig/WinterJsTokenParser.php
```diff
@@ -0,0 +1,60 @@
...
+
+        if (in_array('all', $modules)) {
+            $modules = [
+                'request',
+                'attr',
+                'extras',
+            ];
+        }
+
+        return new WinterJsNode($modules, $lineno, $this->getTag());
```
LukeTowers:
```suggestion
        return new SnowboardNode($modules, $lineno, $this->getTag());
```

modules/cms/twig/WinterJsNode.php
```diff
@@ -0,0 +1,54 @@
...
+use Twig\Node\Node as TwigNode;
+use Twig\Compiler as TwigCompiler;
+
+/**
+ * Represents a "winterjs" node
+ *
+ * @package winter\wn-cms-module
+ * @author Winter CMS
+ */
+class WinterJsNode extends TwigNode
```
LukeTowers:
```suggestion
/**
 * Represents a "snowboard" node
 *
 * @package winter\wn-cms-module
 * @author Winter CMS
 */
class SnowboardNode extends TwigNode
```

modules/cms/twig/Extension.php
LukeTowers: Also could be moved next to the FrameworkTokenParser

modules/system/ServiceProvider.php
```diff
@@ -562,6 +568,10 @@ protected function registerAssetBundles()
             $combiner->registerBundle('~/modules/system/assets/js/framework.combined.js');
             $combiner->registerBundle('~/modules/system/assets/less/framework.extras.less');
         });
+
+        MixAssets::registerCallback(function ($mix) {
+            $mix->registerPackage('framework', '~/modules/system/assets/js/framework-next');
```
LukeTowers:
```suggestion
            $mix->registerPackage('framework', '~/modules/system/assets/js/snowboard');
```

modules/system/assets/js/snowboard/ajax/Request.js
```diff
@@ -0,0 +1,781 @@
...
+            this.element.dispatchEvent(event);
+
+            if (event.defaultPrevented) {
+                return;
+            }
+        }
+
+        // Check for a redirect from the response, or use the redirect as specified in the options. This takes
+        // precedent over all other checks.
+        if (response.X_WINTER_REDIRECT || this.redirect) {
```
LukeTowers: Which would take precedence, a redirect set in the response or in the JS request? I feel like a redirect set in the request should take precedence since it's what the calling code explicitly requested.

modules/system/assets/js/snowboard/ajax/Request.js
```diff
@@ -0,0 +1,781 @@
...
+
+        const formData = new FormData(this.form || undefined);
+        if (Object.keys(data).length > 0) {
+            for (const [key, value] of Object.entries(data)) {
+                formData.append(key, value);
+            }
+        }
+
+        return formData;
+    }
```
LukeTowers, line 751: Ideally this should support the concept of "parent requests / elements", see https://github.com/octobercms/october/pull/5136

modules/system/assets/js/snowboard/ajax/Request.js
```diff
@@ -0,0 +1,781 @@
...
+        const error = new Error(message);
+        error.exception = exception || null;
+        error.file = file || null;
+        error.line = line || null;
+        error.trace = trace || [];
+        return error;
+    }
+}
+
+snowboard.addPlugin('request', Request);
```
LukeTowers:
```suggestion
Snowboard.addPlugin('request', Request);
```

modules/system/assets/js/snowboard/ajax/handlers/AttributeRequest.js
```diff
@@ -0,0 +1,322 @@
...
+
+    resetTrackInputTimer(element) {
+        if (element.dataset.trackInput) {
+            window.clearTimeout(element.dataset.trackInput);
+            element.dataset.trackInput = null;
+        }
+    }
+}
+
+winter.addModule('attributeRequest', AttributeRequest);
```
LukeTowers:
```suggestion
Snowboard.addPlugin('attributeRequest', AttributeRequest);
```

modules/system/assets/js/snowboard/extras/AttachLoading.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * Allows attaching a loading class on elements that an AJAX request is targeting.
+ *
+ * @copyright 2021 Winter.
+ * @author Ben Thomson <git@alfreido.com>
+ */
+ export default class AttachLoading extends Snowboard.Singleton {
```
LukeTowers: Does this plugin not get added to Snowboard as part of its definition?

modules/system/assets/js/snowboard/extras/Flash.js
```diff
@@ -0,0 +1,129 @@
...
+ *  - "type": The type of flash message. This is appended as a class to the flash message itself.
+ *  - "duration": How long the flash message will stay visible for, in seconds. Default: 7 seconds.
+ *
+ * Usage:
+ *      snowboard.flash('This is a flash message', 'info', 8);
+ *
+ * @copyright 2021 Winter.
+ * @author Ben Thomson <git@alfreido.com>
+ */
+export default class Flash extends Snowboard.PluginBase {
```
LukeTowers, line 18: Where does this get added to Snowboard?

modules/system/assets/js/snowboard/extras/Flash.js
```diff
@@ -0,0 +1,129 @@
...
+ * Flash messages will pop up at the top center of the page and will remain for 7 seconds by default. Hovering over
+ * the message will reset and pause the timer. Clicking on the flash message will dismiss it.
+ *
+ * Arguments:
+ *  - "message": The content of the flash message. HTML is accepted.
+ *  - "type": The type of flash message. This is appended as a class to the flash message itself.
+ *  - "duration": How long the flash message will stay visible for, in seconds. Default: 7 seconds.
+ *
+ * Usage:
+ *      snowboard.flash('This is a flash message', 'info', 8);
```
LukeTowers:
```suggestion
 *      Snowboard.flash('This is a flash message', 'info', 8);
```

modules/system/assets/js/snowboard/extras/FlashListener.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * Defines a default listener for flash events.
+ *
+ * Connects the Flash plugin to various events that use flash messages.
+ *
+ * @copyright 2021 Winter.
+ * @author Ben Thomson <git@alfreido.com>
+ */
+export default class FlashListener extends Snowboard.Singleton {
```
LukeTowers, line 9: Where does this plugin get added?

modules/system/assets/js/snowboard/extras/StripeLoader.js
```diff
@@ -0,0 +1,88 @@
+/**
+ * Displays a stripe at the top of the page that indicates loading.
+ *
+ * @copyright 2021 Winter.
+ * @author Ben Thomson <git@alfreido.com>
+ */
+ export default class StripeLoader extends Snowboard.Singleton {
```
LukeTowers: Where does this plugin get added?

modules/system/assets/js/snowboard/extras/Transition.js
```diff
@@ -0,0 +1,171 @@
...
+ *
+ *  - `in`: A class assigned to the element for the first frame of the transition, removed afterwards. This should be
+ *      used to define the initial state of the transition.
+ *  - `active`: A class assigned to the element for the duration of the transition. This should be used to define the
+ *      transition itself.
+ *  - `out`: A class assigned to the element after the first frame of the transition and kept to the end of the
+ *      transition. This should define the end state of the transition.
+ *
+ * Usage:
+ *      snowboard.transition(document.element, 'transition', () => {
```
LukeTowers:
```suggestion
 *      Snowboard.transition(document.element, 'transition', () => {
```

modules/system/assets/js/snowboard/extras/Transition.js
```diff
@@ -0,0 +1,171 @@
...
+ *          console.log('Remove element after 7 seconds');
+ *          this.remove();
+ *      }, '7s');
+ *
+ * @copyright 2021 Winter.
+ * @author Ben Thomson <git@alfreido.com>
+ */
+export default class Transition extends Snowboard.PluginBase {
```
LukeTowers, line 24: Where does this plugin get added?

modules/system/assets/js/snowboard/main/Snowcart.js
```diff
@@ -0,0 +1,347 @@
...
+ *
+ * This class represents the base of a modern take on the Winter JS framework, being fully extensible and taking advantage
+ * of modern JavaScript features by leveraging the Laravel Mix compilation framework. It also is coded up to remove the
+ * dependency of jQuery.
+ *
+ * @copyright 2021 Winter.
+ * @author Ben Thomson <git@alfreido.com>
+ * @link https://wintercms.com/docs/snowboard/introduction
+ */
+export default class Snowboard {
```
LukeTowers: Filename needs to be updated here

modules/system/assets/js/snowboard/main/Snowcart.js
```diff
@@ -0,0 +1,347 @@
...
+    /**
+     * Constructor.
+     *
+     * @param {boolean} autoSingletons Automatically load singletons when DOM is ready. Default: `true`.
+     * @param {boolean} debug Whether debugging logs should be shown. Default: `false`.
+     */
+    constructor(autoSingletons, debug) {
+        /* develblock:start */
+        this.debugEnabled = (typeof debug === 'boolean' && debug === true) ? true : false;
+        /* develblock:end */
```
LukeTowers: Why is this in a block remove statement?

modules/system/assets/js/snowboard/main/Snowcart.js
```diff
@@ -0,0 +1,347 @@
...
+     */
+    addPlugin(name, instance) {
+        const lowerName = name.toLowerCase();
+
+        if (this.hasPlugin(lowerName)) {
+            throw new Error(`A plugin called "${name}" is already registered.`);
+        }
+
+        if (typeof instance !== 'function' && instance instanceof PluginBase === false) {
+            throw new Error(`The provided plugin must be a PluginBase class instance or a callback function.`);
```
LukeTowers:
```suggestion
            throw new Error(`The provided plugin must be an instance of the PluginBase class or a callback function.`);
```

modules/system/assets/js/snowboard/main/Snowcart.js
```diff
@@ -0,0 +1,347 @@
...
+        }
+
+        if (this[lowername] !== undefined) {
+            throw new Error(`The given name is already in use for a property or method of the Snowboard class.`);
+        }
```
LukeTowers: Maybe another check to see if a same name plugin has already been added for a more specific error message?

modules/system/assets/js/snowboard/main/Snowcart.js
```diff
@@ -0,0 +1,347 @@
...
+     * Returns a PluginLoader object of a given plugin.
+     *
+     * @returns {PluginLoader}
+     */
+    getPlugin(name) {
+        if (!this.hasPlugin(name)) {
+            throw new Error(`No plugin called "${name}" has been registered.`);
+        }
+
+        return this.plugins[name];
```
LukeTowers: Needs handling for case-sensitivity

modules/system/assets/js/snowboard/main/Snowcart.js
```diff
@@ -0,0 +1,347 @@
...
+     * @returns {void}
+     */
+    debug() {
+        if (!this.debugEnabled) {
+            return;
+        }
+
+        console.log("%c[Snowboard]", "color: rgb(45, 167, 199);", ...arguments);
+    }
+    /* develblock:end */
```
LukeTowers: Debug function should be available in prod builds for plugins that don't have dev / prod builds but want to only log messages when app.debug is enabled.

modules/system/assets/js/snowboard/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "@wintercms/snowboard",
+    "version": "2.0.0",
```
LukeTowers: Why 2.0 for the initial version?

modules/system/assets/js/snowboard/snowboard.base.js
```diff
@@ -0,0 +1,10 @@
+import Snowboard from './main/Snowboard';
+
+((window) => {
+    const snowboard = new Snowboard();
+
+    // Cover all aliases
+    window.snowboard = snowboard;
+    window.Snowboard = snowboard;
```
LukeTowers, line 8: We should recommend `Snowboard` as the main usage in all our docs

modules/system/assets/js/snowboard/snowboard.extras.js
```diff
@@ -0,0 +1,13 @@
...
+import AttachLoading from './extras/AttachLoading';
+import StripeLoader from './extras/StripeLoader';
+
+((snowboard) => {
+    snowboard.addModule('transition', Transition)
+    snowboard.addModule('flash', Flash);
+    snowboard.addModule('flashListener', FlashListener);
+    snowboard.addModule('attachLoading', AttachLoading);
+    snowboard.addModule('stripeLoader', StripeLoader);
+})(window.snowboard);
```
LukeTowers:
```suggestion
((Snowboard) => {
    ```suggestion
    Snowboard.addPlugin('transition', Transition)
    Snowboard.addPlugin('flash', Flash);
    Snowboard.addPlugin('flashListener', FlashListener);
    Snowboard.addPlugin('attachLoading', AttachLoading);
    Snowboard.addPlugin('stripeLoader', StripeLoader);
})(window.Snowboard);
```

modules/system/assets/js/snowboard/utilities/JsonParser.js
```diff
@@ -0,0 +1,378 @@
+import Singleton from '../abstracts/Singleton';
+
+export default class JsonParser extends Singleton {
```
LukeTowers, line 3: Where is this plugin added?

modules/system/assets/js/snowboard/utilities/Sanitizer.js
```diff
@@ -0,0 +1,58 @@
+import Singleton from '../abstracts/Singleton';
+
+export default class Sanitizer extends Singleton {
```
LukeTowers, line 3: Where is this plugin added?

modules/system/assets/js/snowboard/utilities/Sanitizer.js
```diff
@@ -0,0 +1,58 @@
...
+    constructor(snowboard) {
+        super(snowboard);
+
+        // Add to global function for backwards compatibility
+        window.wnSanitize = (html) => {
+            return this.sanitize(html);
+        }
```
LukeTowers: Do we need an ocSanitize alias too?

modules/system/assets/js/snowboard/utilities/JsonParser.js
```diff
@@ -0,0 +1,378 @@
...
+    constructor(snowboard) {
+        super(snowboard);
+
+        // Add to global function for backwards compatibility
+        window.wnJSON = (json) => {
```
LukeTowers: Do we need an ocJson alias too?

modules/system/classes/MixAssets.php
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @return void
+     */
+    public function init()
+    {
+        // Call plugins
+        $packages = PluginManager::instance()->getRegistrationMethodValues('registerMixPackages');
```
LukeTowers, line 67: Need documentation added

modules/system/classes/MixAssets.php
```diff
@@ -0,0 +1,160 @@
...
+        if (count($packages)) {
+        }
```
LukeTowers, line 78: TODO?

modules/system/classes/MixAssets.php
```diff
@@ -0,0 +1,160 @@
...
+
+            if (count($mix)) {
+                foreach ($mix as $name => $file) {
+                    $path = PathResolver::resolve($theme->getPath() . '/' . $file);
+                    $pinfo = pathinfo($path);
+
+                    $this->registerPackage($name, $pinfo['dirname'], $pinfo['basename']);
+                }
+            }
+        }
```
LukeTowers, line 94: What does a theme definition look like? If it supports an array of mix files should it also support a single mix file as a string?

tests/js/README.md
```diff
@@ -0,0 +1,32 @@
+# Winter JS Framework Testing Suite
+
+The files in this directory form the testing suite for the new Winter JS Framework (`framework-next`). You must install
```
LukeTowers:
```suggestion
The files in this directory form the testing suite for the Snowboard JS Framework (`Snowboard`). You must install
```

tests/js/README.md
```diff
@@ -0,0 +1,32 @@
...
+npm install
+```
+
+You can then run the tests by using the following command:
+
+```bash
+npm run test
+```
+
+Please note that the tests are run against the "built" versions of Winter JS Framework and modules, in order to test
```
LukeTowers:
```suggestion
Please note that the tests are run against the "built" versions of Snowboard and plugins, in order to test
```

tests/js/README.md
```diff
@@ -0,0 +1,32 @@
...
+the exact same functionality that would be delivered to the end user. We do this by leveraging the `JSDOM` library to
+simulate an entire HTML document.
+
+Therefore, you must compile a new build if you make any changes and wish to run the tests:
+
+```bash
+php artisan mix:compile --package framework
```
LukeTowers: More of a comment for the `mix:compile` and `winter:test` commands, but we should have the same alias for both of them shouldn't we? Or does it make more sense for this to be `--package snowboard`?

tests/js/cases/framework/ajax/Request.test.js
```diff
@@ -0,0 +1,584 @@
...
+
+jest.setTimeout(2000);
+
+describe('Request AJAX library', function () {
+    it('can be setup via a global event', function (done) {
+        FakeDom
+            .new()
+            .addScript([
+                'modules/system/assets/js/framework-next/build/framework.js',
+                'modules/system/assets/js/framework-next/build/framework-js-request.js'
```
LukeTowers:
```suggestion
                'modules/system/assets/js/snowboard/build/framework.js',
                'modules/system/assets/js/snowboard/build/framework-js-request.js'
```
Side note, should the files be `framework.js` or `snowboard.js`?

tests/js/cases/framework/ajax/Request.test.js
```diff
@@ -0,0 +1,584 @@
...
+            ])
+            .render()
+            .then(
+                (dom) => {
+                    dom.window.winter.getModule('request').mock('doAjax', (instance) => {
```
LukeTowers:
```suggestion
                    dom.window.Snowboard.getPlugin('request').mock('doAjax', (instance) => {
```

tests/js/cases/framework/ajax/Request.test.js
```diff
@@ -0,0 +1,584 @@
...
+                        // Simulate success response
+                        const resolved = Promise.resolve({
+                            success: true
+                        });
+
+                        // Mock events
+                        instance.winter.globalEvent('ajaxStart', instance, resolved);
```
LukeTowers:
```suggestion
                        instance.Snowboard.globalEvent('ajaxStart', instance, resolved);
```

tests/js/cases/framework/ajax/Request.test.js
```diff
@@ -0,0 +1,584 @@
...
+                            const event = new dom.window.Event('ajaxPromise');
+                            event.promise = resolved;
+                            instance.element.dispatchEvent(event);
+                        }
+
+                        return resolved;
+                    });
+
+                    // Listen to global event
+                    dom.window.winter.addModule('testListener', class TestListener extends dom.window.winter.Singleton {
```
LukeTowers:
```suggestion
                    dom.window.Snowboard.addPlugin('testListener', class TestListener extends dom.window.winter.Singleton {
```

tests/js/cases/framework/ajax/Request.test.js
```diff
@@ -0,0 +1,584 @@
...
+                                ajaxSetup: 'ajaxSetup',
+                            };
+                        }
+
+                        ajaxSetup(instance) {
+                            instance.handler = 'onChanged';
+                        }
+                    });
+
+                    dom.window.winter.request(undefined, 'onTest', {
```
LukeTowers:
```suggestion
                    dom.window.Snowboard.request(undefined, 'onTest', {
```

tests/js/cases/framework/ajax/Request.test.js
```diff
@@ -0,0 +1,584 @@
...
+                }
+            );
+    });
+
+    it('can be cancelled on setup via a global event', function (done) {
+        FakeDom
+            .new()
+            .addScript([
+                'modules/system/assets/js/framework-next/build/framework.js',
+                'modules/system/assets/js/framework-next/build/framework-js-request.js'
```
LukeTowers:
```suggestion
                'modules/system/assets/js/snowboard/build/framework.js',
                'modules/system/assets/js/snowboard/build/framework-js-request.js'
```

tests/js/cases/framework/ajax/Request.test.js
```diff
@@ -0,0 +1,584 @@
...
+            ])
+            .render()
+            .then(
+                (dom) => {
+                    dom.window.winter.getModule('request').mock('doAjax', (instance) => {
```
LukeTowers:
```suggestion
                    dom.window.Snowboard.getPlugin('request').mock('doAjax', (instance) => {
```

tests/js/cases/framework/ajax/Request.test.js
```diff
@@ -0,0 +1,584 @@
...
+                        // Simulate success response
+                        const resolved = Promise.resolve({
+                            success: true
+                        });
+
+                        // Mock events
+                        instance.winter.globalEvent('ajaxStart', instance, resolved);
```
LukeTowers:
```suggestion
                        instance.Snowboard.globalEvent('ajaxStart', instance, resolved);
```

LukeTowers: You get the picture, all the tests need to be fixed 😉

themes/demo/layouts/default.htm
```diff
@@ -36,7 +37,10 @@
         <script src="{{ 'assets/vendor/jquery.js'|theme }}"></script>
         <script src="{{ 'assets/vendor/bootstrap.js'|theme }}"></script>
         <script src="{{ 'assets/javascript/app.js'|theme }}"></script>
-        {% framework extras %}
+        <script src="/modules/system/assets/js/framework-next/build/framework.js"></script>
+        <script src="/modules/system/assets/js/framework-next/build/framework-js-request.js"></script>
+        <script src="/modules/system/assets/js/framework-next/build/framework-attr-request.js"></script>
+        <script src="/modules/system/assets/js/framework-next/build/framework.extras.js"></script>
```
LukeTowers: These can be replaced with the `{% snowboard %}` tag now right?

themes/demo/layouts/default.htm
```diff
@@ -13,6 +13,7 @@
         <link rel="icon" type="image/png" href="{{ 'assets/images/winter.png'|theme }}">
         <link href="{{ 'assets/css/vendor.css'|theme }}" rel="stylesheet">
         <link href="{{ 'assets/css/theme.css'|theme }}" rel="stylesheet">
+        <link href="/modules/system/assets/css/framework.extras.css" rel="stylesheet">
```
LukeTowers: Should be unnecessary with the `{% snowboard %}` tag now right?

themes/demo/pages/ajax.htm
```diff
@@ -44,16 +49,15 @@ <h1>AJAX framework</h1>
...
+                <input type="text" class="form-control" value="15" name="value1" style="width:100px" data-track-input>
+                <select class="form-control" name="operation" style="width: 70px" data-track-input>
                     <option>+</option>
                     <option>-</option>
                     <option>*</option>
                     <option>/</option>
                 </select>
-                <input type="text" class="form-control" value="5" name="value2" style="width:100px">
-                <button type="submit" class="btn btn btn-primary" data-attach-loading>Calculate</button>
+                <input type="text" class="form-control" value="5" name="value2" style="width:100px" data-track-input>
```
LukeTowers, line 60: Should probably keep the calculate button anyways right?

LukeTowers (changes requested):

LukeTowers:
>A second extension - `AttributeRequest` - provides an augmentation to this functionality and allow the AJAX framework to also handle AJAX calls that use the `data-` attributes. By separating these features, the developer has the ability to turn off support for the `data-` attribute AJAX framework should they not need it.

Had another think on this, and I think that this should be included by default in the `{% snowboard %}` tag but allow for `{% snowboard no-attr %}` to be used to disable it. IMO the out of box experience of the `{% snowboard %}` tag should match what the out of box experience of the `{% framework %}` tag is to ensure that the maximum default support for the features is available for the plugin / theme ecosystem. We can still allow for the `no-` flags to cut out undesired pieces by the theme devs, but then that is an explicit action that accepts responsibility for that theme's ability to support the plugin ecosystem at large out of the box.

modules/system/assets/js/snowboard/ajax/Request.js
bennothommo: At the moment, the redirect in the response takes precedence, but I agree with you in that these should be swapped.

LukeTowers: Further on to that discussion, under what conditions would the developer provide a redirect value in the request in the first place and what would they be looking to achieve with that?

modules/system/assets/js/snowboard/extras/AttachLoading.js
bennothommo: It does. All scripts in the `extras` folder are compiled to the `snowboard.extras.js` build.

modules/system/assets/js/snowboard/extras/Flash.js
bennothommo: As above, this is compiled to `snowboard.extras.js`.

modules/system/assets/js/snowboard/main/Snowcart.js
bennothommo: The original intention was that when Snowboard was compiled for production, it would strip out the debugging method and calls to bring the filesize down, since there's no need for debugging in production. However, given that we're going to include debugging but only enable it when `app.debug` is true, I'm going to remove these comments.

modules/system/assets/js/snowboard/package.json
bennothommo: I guess this was left over from when the intention that this framework would replace the old one. ;)

modules/system/assets/js/snowboard/main/Snowcart.js
bennothommo: That check happens above - `if (this.hasPlugin(lowerName)) { ...`

modules/system/assets/js/snowboard/utilities/Sanitizer.js
bennothommo: Utilities are included through the main Snowboard class (method `loadUtilities`).

tests/js/README.md
bennothommo: Probably should be "snowboard" to make it clear what is being compiled - it'll also be less likely a name to be used anywhere else.

tests/js/cases/framework/ajax/Request.test.js
bennothommo: Yeah I hadn't fixed the tests yet haha ;)

themes/demo/layouts/default.htm
bennothommo: No, the Snowboard tag only adds the JS. The user still needs to manually add the CSS. This is the same as the original framework.

modules/system/classes/MixAssets.php
bennothommo: It has to be an array so that the "name" of the package can be defined.

I'm using this for the Workshop theme already - the config looks like this:

```yaml
mix:
    workshop-theme: winter-mix.js
```

themes/demo/pages/ajax.htm
bennothommo: I suppose, but it is redundant at this point.

bennothommo:
@LukeTowers 

> Had another think on this, and I think that this should be included by default in the `{% snowboard %}` tag but allow for `{% snowboard no-attr %}` to be used to disable it. IMO the out of box experience of the `{% snowboard %}` tag should match what the out of box experience of the `{% framework %}` tag is to ensure that the maximum default support for the features is available for the plugin / theme ecosystem. We can still allow for the `no-` flags to cut out undesired pieces by the theme devs, but then that is an explicit action that accepts responsibility for that theme's ability to support the plugin ecosystem at large out of the box.

I'm of the opposite opinion. I find it instantly more readable for the Snowboard tag to include which features it is using. If I'm looking at a couple of themes, I feel it would be easier to see `{% snowboard request data-attr %}` or `{% snowboard request extras %}`, etc. because then I can immediately tell they're using those features. By contrast, `{% snowboard no-data-attr %}` makes it clear they're *not* using data attributes, but doesn't say what features they are using.

I don't think it'll be a friction for people to have to explicitly add which features they want to use (or use `{% snowboard all %}` to use everything). The other benefit is that if we decide to add new features down the line, and add another key to the tag, it becomes opt-in - the theme developers specifically need to add it to the tag if they wish to include the new feature. The other thing is that the original framework made the "extras" opt-in, which I assume we would need to keep in Snowboard too. So we'll end up with `{% snowboard no-data-attr extras %}` which could be misinterpreted on first pass to say we're *not* using data attributes or extras.

modules/system/classes/MixAssets.php
LukeTowers: I meant in the docs PR 😉

modules/system/assets/js/snowboard/ajax/handlers/AttributeRequest.js
```diff
@@ -0,0 +1,322 @@
...
+        this.disableDefaultFormValidation();
+    }
+
+    /**
+     * Dependencies.
+     *
+     * @returns {string[]}
+     */
+    dependencies() {
+        return ['request', 'jsonParser'];
```
LukeTowers, line 40: How are dependencies handled if someone replaces the request class with their own implementation but still wants to be able to use the AttributeRequest class?

modules/system/assets/js/snowboard/winter-mix.js
```diff
@@ -0,0 +1,17 @@
...
+
+// Compile Snowboard framework
+mix
+    .js('./snowboard.base.js', './build/snowboard.base.js')
+    .js('./snowboard.base.debug.js', './build/snowboard.base.debug.js')
+    .js('./ajax/Request.js', './build/snowboard.request.js')
+    .js('./ajax/handlers/AttributeRequest.js', './build/snowboard.data-attr.js')
+    .js('./snowboard.extras.js', './build/snowboard.extras.js')
+    .polyfill({
+        enabled: mix.inProduction(),
```
LukeTowers, line 14: Does this mean that if Snowboard is compiled while app.debug = true (or some other indicator that the current .env is not production) then the polyfills won't be included? Wouldn't we want them to just always be included?

tests/js/cases/system/framework.test.js
```diff
@@ -1,673 +0,0 @@
-import { assert } from 'chai'
```
LukeTowers, line 1: Are we not running the tests for the original framework.js anymore or are they not really that useful?

themes/demo/layouts/default.htm
```diff
@@ -13,6 +13,7 @@
         <link rel="icon" type="image/png" href="{{ 'assets/images/winter.png'|theme }}">
         <link href="{{ 'assets/css/vendor.css'|theme }}" rel="stylesheet">
         <link href="{{ 'assets/css/theme.css'|theme }}" rel="stylesheet">
+        <link href="/modules/system/assets/css/snowboard.extras.css" rel="stylesheet">
```
LukeTowers, line 16: I could have sworn that including `extras` in the `{% framework %}` tag also included the stylesheet, at the very least the docs say that it does: https://wintercms.com/docs/ajax/extras

LukeTowers (changes requested):

modules/system/assets/js/snowboard/ajax/Request.js
bennothommo: I'm not sure of the rationale, but it was in the original framework so I've kept it as is.

modules/system/assets/js/snowboard/ajax/handlers/AttributeRequest.js
bennothommo: Dependencies work simply by checking if a plugin with the given name(s) are added. In the case above, `AttributeRequest` is expected a plugin named `request` and a plugin named `jsonParser` to exist, because there's no real way to make a "soft" reference to a class that may or may not exist in JS.

For your particular case, the developer would have two options. They could name their custom Request implementation as "request" which would satisfy the dependency above, or they could extend the `AttributeRequest` class in their own implementation and override the dependencies.

modules/system/assets/js/snowboard/winter-mix.js
bennothommo: This is more for framework development - `mix.inProduction()` is `true` if the `mix:compile` artisan command is run with the `production` flag. I don't include the polyfills in development because it slows down compilation, especially if I'm using `mix:watch` to watch for changes.

tests/js/cases/system/framework.test.js
bennothommo: If we're not going to be updated the original framework anymore (beyond the parent form functionality), it probably won't be necessary to test the old framework. It just doesn't change enough to justify running tests on it all the time.

themes/demo/layouts/default.htm
bennothommo: You're right, it does haphazardly include it with the JS files here: https://github.com/wintercms/winter/blob/5b58103107d1e8f7f5caa9042d6496b676af2724/modules/cms/twig/FrameworkNode.php#L50

I feel it would be better to have the developer explicitly include it or not, in the case that they completely customise the styling of the extras functionality. Thoughts?

modules/system/assets/js/snowboard/winter-mix.js
LukeTowers: gotcha, so I just need to remember to add the flag when committing new builds :)

themes/demo/layouts/default.htm
LukeTowers: Meh, the features added by the extras build are all UI based, so that combined with the fact that the current behaviour includes it makes me lean towards including it by default as well.

bennothommo: No probs, I'll include it then.

bennothommo:
Work continued in https://github.com/wintercms/winter/pull/401 to target the 1.1.8 release.

PR #334, opened 2021-10-13
Add Exception on wrong relation type in relation formwidget

RomainMazB:
Fix #333 

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/334"></a>

PR #358, opened 2021-11-28
Update Latvian language

arvislacis:
Pull request updates Latvian language translation - added/changed translation strings related to the new color picker widget (https://github.com/wintercms/winter/pull/324) and relation form widget error message (https://github.com/wintercms/winter/pull/334).

In addition I recompiled client sided JS files with `php artisan winter:util compile lang` so it includes also recompiled file of `lang.ru.js` - I guess @WebVPF didn't include these changes with https://github.com/wintercms/winter/pull/349

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/358"></a>


modules/system/assets/js/lang/lang.ru.js
```diff
@@ -6,7 +6,7 @@ if ($.oc === undefined) $.oc = $.wn
...
 );
 
 //! moment.js locale configuration v2.22.2
```
LukeTowers, line 12: Please revert the changes to the permissions on these two compiled files, they should be set to 644 by default in the repository, looks like when you compiled them it changed both compiled JS lang files to 755.

LukeTowers (changes requested):

modules/system/assets/js/lang/lang.ru.js
arvislacis: Can you please recheck this? Because at my local environment they have **644** permission, just for curiosity I changed permissions to 755 and git showed new changes, so the original commits had 644 permission (well, at least at my local git repository).

bennothommo: @arvislacis GitHub is definitely showing them as having been changed to 0755 permissions in your branch:

Screen Shot 2021-11-29 at 4 42 11 pm

arvislacis: @LukeTowers @bennothommo Should be fixed in the latest commit. Thanks for noticing it. :)

LukeTowers (approved):

PR #360, opened 2021-11-29
Update Ukrainian language

WebVPF:


WebVPF:
Commit `fix lang key` - I think it was a typo. I decided to fix it right away.

arvislacis:
@WebVPF Maybe you could also regenerate https://github.com/wintercms/winter/blob/develop/modules/system/assets/js/lang/lang.uk.js with `php artisan winter:util compile lang` and add it to this pull request.

WebVPF:
@arvislacis 
I did not write this commit on the server, but simply forked the repository in VSCode, made the edit and sent the commit. Therefore, I can not run console commands in php. And to be honest, I didn't know what else can be compiled for js.

arvislacis:
> @arvislacis как я понял, вы русский язык знаете? Можно я на русском вам отвечу. Я этот коммит писал не на сервере, а просто форкнул репозиторий в VSCode, сделал правку и отправил коммит. Поэтому не могу запустить консольные команды на php. И если честно, то я и не знал что ещё и для js можно компилировать.

Ok, no problem. :) I am sure JS files will be recompiled sooner or later (by maintainers or when there will another language update). Strings in `modules/system/lang/uk/client.php` are used in front-end by JavaScript components and filters so recompilation is needed because at the end `assets/js/lang/lang.uk.js` file is used for displaying translations to user, not `modules/system/lang/uk/client.php`.

PR #361, opened 2021-11-30
Constrain project's support PHP versions

ericp-mrel:
Setting this PHP version constraint should make it more obvious to other developers PHP 8.1 is not supported and will not work.

This might also prevent developers from creating a new project using `composer create-project` on PHP 8.1, although I haven't personally tested this. But, this would be the desired outcome and hopefully composer notifies them the PHP version they are using is not supported.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/361"></a>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "source": "https://github.com/wintercms/winter"
     },
     "require": {
-        "php": ">=7.2.9",
+        "php": "^7.2.9|^8.0",
         "winter/storm": "dev-develop as 1.1.999",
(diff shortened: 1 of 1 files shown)


LukeTowers:
Good catch @ericp-mrel. Laravel 6 LTS doesn't support PHP 8.1 which is the main reason why the 1.1 branch of Winter can't support it. We are actively working on 1.2 though which will use Laravel 9 LTS, so this is a good fix for the month or so before Laravel 9 is available.

PR #365, opened 2021-12-02
Update `cms` french lang file

damsfx:
Add missing translations for CMS  module

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/365"></a>


modules/cms/lang/fr/lang.php
```diff
@@ -99,7 +99,14 @@
...
         'is_enabled_comment' => 'Si activé, la page choisie ci-dessous sera affichée pour les visiteurs du site Web.',
-        'hint' => 'Le mode maintenance affichera la page de maintenance pour les visiteurs qui ne sont pas authentifiés dans l’interface d’administration.'
+        'hint' => 'Le mode maintenance affichera la page de maintenance pour les visiteurs qui ne sont pas authentifiés dans l’interface d’administration.',
+        'allowed_ips' => [
+            'name' => 'Adresses IP autorisées',
+            'description' => 'Adresses IP autorisées à consulter le site lorsque le mode de maintenance est actif.',
+            'prompt' => 'Ajouter une adresse IP',
+            'ip' => 'Adresse IP',
+            'label' => 'Description'
+        ]
```
LukeTowers:
```suggestion
            'label' => 'Description',
        ],
```
Always use trailing commas please.

modules/cms/lang/fr/lang.php
```diff
@@ -99,7 +99,14 @@
         'settings_menu_description' => 'Configurer la page de maintenance et ajuster ses options.',
         'is_enabled' => 'Activer la maintenance',
         'is_enabled_comment' => 'Si activé, la page choisie ci-dessous sera affichée pour les visiteurs du site Web.',
-        'hint' => 'Le mode maintenance affichera la page de maintenance pour les visiteurs qui ne sont pas authentifiés dans l’interface d’administration.'
+        'hint' => 'Le mode maintenance affichera la page de maintenance pour les visiteurs qui ne sont pas authentifiés dans l’interface d’administration.',
```
LukeTowers: Should `’` be used here or `'`, and then should be updated everywhere else?

damsfx: Difference between the typewritten or straight apostrophe `'` and the typographical or traditional apostrophe `’`!  

> At the time of the first typewriters, for reasons of economy of keys, the sign `'` was used to note approximately the apostrophe `’`. This is why the straight elision mark is called a typewriter apostrophe.  
> Later, again because of space limitations, computer keyboards and early programming languages picked up the hybrid sign `'`.  (They also gave it new uses, including prime `´` and acute accent `´`.)  
> In neat typography, the traditional apostrophe is preferred to the straight apostrophe, especially in print, but also in electronic documents.  

Source "Office québecois de la langue française" : [Apostrophes dactylographique et typographique](http://bdl.oqlf.gouv.qc.ca/bdl/gabarit_bdl.asp?id=5173) in french
See also : [What’s the right character for an apostrophe?](https://graphicdesign.stackexchange.com/questions/66809/what-s-the-right-character-for-an-apostrophe) in english

I confess I don't know what to answer in the end... A matter of taste?  
On a personal note, I use the traditional apostrophe in all my plugins.

LukeTowers: @bennothommo any thoughts?

bennothommo: I don't mind either way, to be honest. I use the typewriter apostrophe out of habit, but if a certain language has a strong preference for another character, then I think that's OK too if the lang file reflects that. I imagine we'll likely continue to use the typewriter apostrophe for the English language messages though.

LukeTowers: Okay, I'm going to say that we should go with the typewriter apostrophe for consistency then unless anyone can provide a strong reason to use a different character.

damsfx: Do you want me to update the PR to reflect this choice?

LukeTowers: Yes please @damsfx as well as the other ones.

PR #367, opened 2021-12-05
Fixing datepicker readOnly property

### LINKED ISSUE #363: readOnly: true on a type: datepicker is still editable (but grayed out)
### Winter CMS Build

1.1

### PHP Version

8.0

### Database engine

MySQL/MariaDB

### Plugins installed

Winter.Sitemap, Winter.Pages, Winter.TinyPNG, Winter.Translate

### Issue description

Models field.yaml
readOnly: true on a type: datepicker is still editable (but grayed out)

```
#  Field Definitions

fields:

    created_at:
        label: aic.account::lang.general.created_at
        type: datepicker
        mode: datetime
        format: d-m-Y
        span: storm
        cssClass: col-xs-6
        readOnly: true
```

### Steps to replicate

Create field with readOnly and type datepicker

### Workaround

_No response_

### Any further details?

_No response_

arvislacis:
Fixes https://github.com/wintercms/winter/issues/363

`&& !$field->disabled` check is used so `diabled` attribute doesn't get added the second time if also used in `<?= $field->getAttributes() ?>` (if `disabled: true` also is specified).

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/367"></a>

### DIFF
--- a/modules/backend/formwidgets/datepicker/partials/_picker_date.htm
+++ b/modules/backend/formwidgets/datepicker/partials/_picker_date.htm
@@ -7,6 +7,5 @@
         class="form-control align-right"
         autocomplete="off"
         <?= $field->getAttributes() ?>
-        <?= ($field->readOnly && !$field->disabled) ? 'disabled="disabled"' : '' ?>
         data-datepicker />
 </div>
--- a/modules/backend/formwidgets/datepicker/partials/_picker_date.htm
+++ b/modules/backend/formwidgets/datepicker/partials/_picker_date.htm
@@ -6,6 +6,7 @@
         id="<?= $this->getId('date') ?>"
         class="form-control align-right"
         autocomplete="off"
+        <?= $field->readOnly ? $field->disabled = true : '' ?>
         <?= $field->getAttributes() ?>
         data-datepicker />
 </div>
--- a/modules/backend/formwidgets/datepicker/partials/_picker_date.htm
+++ b/modules/backend/formwidgets/datepicker/partials/_picker_date.htm
@@ -7,5 +7,6 @@
         class="form-control align-right"
         autocomplete="off"
         <?= $field->getAttributes() ?>
+        <?= ($field->readOnly && !$field->disabled) ? 'disabled="disabled"' : '' ?>
         data-datepicker />
 </div>
--- a/modules/backend/formwidgets/datepicker/partials/_picker_time.htm
+++ b/modules/backend/formwidgets/datepicker/partials/_picker_time.htm
@@ -7,6 +7,5 @@
         class="form-control align-right"
         autocomplete="off"
         <?= $field->getAttributes() ?>
-        <?= ($field->readOnly && !$field->disabled) ? 'disabled="disabled"' : '' ?>
         data-timepicker />
 </div>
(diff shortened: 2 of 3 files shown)


bennothommo:
@arvislacis I think the better fix here would be to edit the JS file for the datepicker (`modules/system/assets/ui/js/datepicker.js`) to not instantiate a date/time picker on a field that is read only or disabled. `$field->getAttributes()` is already correctly populating the necessary attributes, so it's mainly the JS at fault here.

arvislacis:
@bennothommo Disabled attribute and state is already handled properly... I will try to look at JS implementation of fix.

Also one of possible alternative fixes is to just output value, like it is used here: https://github.com/wintercms/winter/blob/25cba5eabc84f302cbd5ba67e0c78fd4ae985d65/modules/backend/formwidgets/datepicker/partials/_datepicker.htm#L1-L6

^ What's your opinion about this approach?

Anyway, IMHO, at some time in the future all `previewMode`, `disabled`, `readOnly` and `$field->getAttributes()` states and their logic should be reviewed for field partial views because there are too many different approaches for different field types. But that's another topic.

arvislacis:
@bennothommo https://github.com/wintercms/winter/pull/367/commits/732a9bcdcd4d7db5ba2bceddd529a18dbf45c2fe <-- JS implementation of fix.

The main problem with date and time pickers is that they should be initialized first, otherwise they will have empty values in UI and also some dependent logics may not be triggered as expected.

If this solution also doesn't seem to be ok then the last option is to only output value as I mentioned in my previous comment.

While researching this I also found related issue to this in Pikaday library: https://github.com/Pikaday/Pikaday/issues/662

LukeTowers:
The first approach was probably fine, but shouldn't it have been `$field->readOnly || $field->disabled`?

arvislacis:
> The first approach was probably fine, but shouldn't it have been `$field->readOnly || $field->disabled`?

No because `$field->getAttributes()` already adds `disabled` and `readonly` attributes (if defined) to the input field so `!$field->disabled` was used to don't include disabled twice in the rendered ouput (for example, if readonly is used combination with disabled).

Also `disabled: true` and disabled attribute works as intended on datepicker fields, problem is only with readonly.

arvislacis:
@bennothommo Any further comments from your side?

bennothommo:
@arvislacis Looks good to me, we just need to test it.

LukeTowers:
@bennothommo why were you wanting those attributes defined in JS rather than PHP?

bennothommo:
@LukeTowers because the datepicker JS widget itself was ignoring the `readonly` and `disabled` attributes and still instantiating on the field, so it seemed to me that we should instead disable the JS widget entirely if either of those options are used.

LukeTowers:
Looking at both commits it appears that it does support looking at the disabled attribute, so I actually think that the proper fix is setting field->disabled to true when readOnly is set to true and doing that before getAttributes() is called. That's definitely the simplest solution.

arvislacis:
In https://github.com/wintercms/winter/pull/367/commits/835869b02487a7ba66639d19f98459b8c798d481 added 3rd implementation option suggested by @LukeTowers . I personally also like and prefer this one better that previous ones.

LukeTowers:
Fixed in https://github.com/wintercms/winter/commit/bd79b80d0141f70a8157df1a24a5469cb6b2749e, thanks for your help @arvislacis!

PR #370, opened 2021-12-06
Add md_line Twig filter to expose Markdown::parseLine()

matteotrubini:
If accepted, also docs should be updated accordingly.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/370"></a>


LukeTowers:
Could you provide an example of this being used and why it would be used?

matteotrubini:
`{{ 'Hello _Parsedown_!'|md }}` would print `<p>Hello <em>Parsedown</em>!</p>`

`{{ 'Hello _Parsedown_!'|md_line}}` would print `Hello <em>Parsedown</em>!`

Currently if I don't want to enclose the text inside a paragraph I cannot use `|md`, so I must use html and `|raw`.

As the method `parseLine()` already exists I think that can be convenient to make it available via filter.

Issue #376, opened 2021-12-13
Custom stylesheet editor throws wrong errors


AIC-BV:
### Winter CMS Build

1.1

### PHP Version

8.0

### Database engine

MySQL/MariaDB

### Plugins installed

Winter.Sitemap, Winter.Pages, Winter.TinyPNG, Winter.Translate

### Issue description

Noticed at `Customize backend -> styles` that the error messages displayed are wrong.
At first I thought it was a CSS file.
But apparently it supports SCSS and it throws errors (but it still works)
![image](https://user-images.githubusercontent.com/89913092/145796408-796ae1f5-7586-4edb-a8e3-895af5ae54ba.png)

### Steps to replicate

add scss in `Customize backend -> styles`

LukeTowers:
Yeah, at some point we're looking to replace the usage of Ace Editor with Monaco (the editor used by VS Code) which will probably solve those sorts of issues. Going to close this as I don't personally plan on doing anything to address this at the moment, but if anyone else is interested in making a PR (to either fix this or replace Ace with Monaco) then I'd be happy to reopen it.

LukeTowers:
Might be as simple as changing the language used for that field from CSS to SCSS btw, feel free to try that and report back.

AIC-BV:
> Might be as simple as changing the language used for that field from CSS to SCSS btw, feel free to try that and report back.

Yeah changing the language to scss solves the problem!

PR #377, opened 2021-12-14
Update custom_styles language from css to scss

AIC-BV:
Solves wrong error messages (see https://github.com/wintercms/winter/issues/376)

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/377"></a>

### DIFF
--- a/modules/backend/models/brandsetting/fields.yaml
+++ b/modules/backend/models/brandsetting/fields.yaml
@@ -106,4 +106,4 @@ tabs:
             type: codeeditor
             tab: backend::lang.branding.styles
             size: giant
-            language: scss
+            language: less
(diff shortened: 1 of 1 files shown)


modules/backend/models/brandsetting/fields.yaml
```diff
@@ -106,4 +106,4 @@ tabs:
             type: codeeditor
             tab: backend::lang.branding.styles
             size: giant
-            language: css
+            language: scss
```
LukeTowers:
```suggestion
            language: less
```
Looking at the code in BrandSetting.php the language in use is actually LESS. Technically it could be made to support any form of CSS compilation supported by the Asset Compiler but we'll just leave it set to LESS for now.

PR #387, opened 2021-12-25
Use HTTPS template for links in Markdown editor

WebVPF:


LukeTowers:
@WebVPF @bennothommo what does this change mean for HTTP only sites?

bennothommo:
Nothing - it just means that when they press the Link button, they'll have to remove the `s` from `https`, but I'd argue that it'll be more common for people to use `https` links this day and age.

PR #388, opened 2021-12-26
Popup Docs UI - wrap configuration in code tags

WebVPF:


modules/system/assets/ui/docs/popup.md
```diff
@@ -135,13 +135,13 @@ The partial for your rendered popup should follow this structure:
 - content: content HTML string or callback
```
bennothommo:
```suggestion
- `content` - content HTML string or callback
```

PR #401, opened 2022-01-12
[1.1.8] Implement an opt-in improved JS framework (Snowboard)

bennothommo:
Snowboard

Documentation: https://github.com/wintercms/docs/pull/45

This PR moves towards introducing a new JavaScript framework for Winter, that uses modern JavaScript and removes the hard-dependency to jQuery that has traditionally been the case with Winter and October before it.

If accepted, we intend to make this **opt-in**, and still provide the option of using the original framework for legacy apps and older browser support.

### Goals

- Rewrite the framework to be modular and give the user full control of what features are to be enabled in the framework, and not impose technical opinions on the developer.
- Use a vanilla JavaScript approach to remove the need for jQuery and decrease the asset bandwidth considerably.
- Leverage the power of ES6 classes to allow for inherited code, extensions and bring the code style more in-line with our PHP code.
- Take advantage of the Node.js ecosystem to provide minification and compatibility with a wide variety of browsers using Babel.
- Still provide exactly (or as close to as possible) the same functionality that the original framework provides, with a minimum need for rewriting established projects.

### Concepts

The four cornerstones of this framework are as follows:

#### The `Snowboard` class
Represents the global container for the application. This stores all activated plugins and provides instances of these modules as required. It is synonymous with the `Application` class of Laravel, or a `Vue` instance.

#### The `PluginLoader` class
The main conduit between a plugin and the Snowboard app. This provides the mechanism for delivering instances of each module (or a single instance of Singleton plugin) and resolves dependencies.

#### The `PluginBase` class
A plugin is a specific extension to the Snowboard application. It can contain code that provides new functionality or augments other functionality already registered in the application. A plugin can be reused - in essence, each call to the plugin will create a new instance and will be independent from other instances. This would be used for scenarios such as Flash messages, AJAX requests and the like.

#### The `Singleton` class
An extension of the `PluginBase` class. It signifies to the application that this plugin should exist once, and all uses of this plugin should use the same instance. This would be useful for things like event handlers, extending functionality of another plugin or global functionality.

A singleton is initialized automatically when the DOM is ready, if it is called directly or if it listens to an event that is fired. The singleton instance is then used no matter how many times it is called from the Snowboard class.

### What's included

By default, this framework only includes the framework architecture and a couple of utilities. This is an intentional choice to allow the greatest flexibility in how the framework is used. The utilities include:

- A JSON parser that is customized to handle the established data structure of Winter (and October before it).
- An HTML sanitizer that provides sanitization for HTML code.
- A debouncer, that acts similar to the `Vue.newTick` function in allowing code to run after the previous frame is rendered - useful for ensuring DOM changes are made and events are correctly fired.

In addition, some opt-in features include a rewritten `Request` class that represents a fresh take on the JavaScript AJAX framework available in Winter. This new `Request` class leverages the `fetch` method in all modern browsers to conduct AJAX calls. It supports all options and configuration that were previously available in the original framework.

A second extension - `AttributeRequest` - provides an augmentation to this functionality and allow the AJAX framework to also handle AJAX calls that use the `data-` attributes. By separating these features, the developer has the ability to turn off support for the `data-` attribute AJAX framework should they not need it.

### What's not included

- The HTML data attribute framework allowed for arbitrary JavaScript to be called from within `data-` attribute callbacks (eg. `data-request-success`, `data-request-error` etc.). This feature, at a minimum, prevented sites using content security policies because of its use of the `eval()` method in JavaScript - at worst, it presented a potential vector for naughtiness. We will not be including this in the framework.

### Further notes

- Due to the usage of Promises, ES6 classes and ES6 syntax, this framework *does not* support any version of Internet Explorer. It however is supported by all browsers available since 2017 (except for, I suppose, Opera Mini or the Samsung Browser).
- Node.js will only be required for development of the framework. There will be no need for Node.js usage in themes, plugins, components, etc. (of course, you'll be free to use it in your own development pipeline if you want).
- More details will be forthcoming as the framework solidifies.

### To-do

- [ ] Documentation ([wintercms/docs](https://github.com/wintercms/docs/pull/45))
- [x] Unit tests
- [x] Finish off all abilities of the HTML data attributes framework
- [x] Full testing ([test cases added to Workshop theme](https://github.com/wintercms/wn-workshop-theme))

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/401"></a>


modules/system/assets/js/snowboard/abstracts/PluginBase.js
```diff
@@ -0,0 +1,48 @@
...
+export default class PluginBase {
+    /**
+     * Constructor.
+     *
+     * The constructor is provided the Snowboard framework instance.
+     *
+     * @param {Snowboard} snowboard
+     */
+    constructor(snowboard) {
+        this.snowboard = snowboard;
```
LukeTowers, line 18: Should this reference be `this.Snowboard` to match the public API or am I overthinking it?

modules/system/assets/js/snowboard/ajax/Request.js
```diff
@@ -0,0 +1,789 @@
...
+            });
+        }
+
+        const ajaxPromise = new Promise((resolve, reject) => {
+            fetch(this.url, {
+                method: 'POST',
+                headers: this.headers,
+                body: this.data,
+                redirect: 'follow',
+                mode: 'same-origin',
```
LukeTowers: This might be something that we explore at a later date, but is it possible to make AJAX requests to another domain with the current code? I know I've done that with the old framework once before (required setting up the CORS plugin but otherwise worked) in order to embed the framework and a simple use of it on an external site that wasn't powered by Winter in order to still be able to use some of Winter's power.

modules/system/assets/js/snowboard/ajax/Request.js
```diff
@@ -0,0 +1,789 @@
...
+        // Check for a redirect from the response, or use the redirect as specified in the options. This takes
+        // precedent over all other checks.
+        if (this.redirect || response.X_WINTER_REDIRECT) {
+            this.processRedirect(this.redirect || response.X_WINTER_REDIRECT);
+            return;
+        }
+
+        if (this.flash && response.X_WINTER_FLASH_MESSAGES) {
+            this.processFlashMessages(response.X_WINTER_FLASH_MESSAGES);
+        }
```
LukeTowers: Should we process flash messages before processing redirects in case they display a flash message before the redirect? Also would help with redirects to file downloads. 

I ask because if the flash message has been sent then it's already been pulled from the session and theoretically won't be sent again. Most useful feature would be something like returning a redirect to a file download and displaying a flash message "Downloading file..."

modules/system/assets/js/snowboard/ajax/Request.js
```diff
@@ -0,0 +1,789 @@
...
+                const cookie = cookies[i].trim();
+
+                if (cookie.substring(0, 11) === 'XSRF-TOKEN=') {
+                    cookieValue = decodeURIComponent(cookie.substring(11));
+                    break;
+                }
+            }
+        }
+
+        return cookieValue;
```
LukeTowers: Is a nicer cookie handling API something that's worth including in the base Snowboard API?

modules/system/assets/js/snowboard/ajax/handlers/AttributeRequest.js
```diff
@@ -0,0 +1,327 @@
...
+    /**
+     * Sets up an AJAX request via HTML attributes.
+     *
+     * @param {Request} request
+     */
+    onAjaxSetup(request) {
+        const fieldName = request.element.getAttribute('name');
+
+        const data = {
+            ...this.getParentRequestData(request.element),
```
LukeTowers, line 224: Is this just the function that allows a data-request on a button element to include data-request-data defined on the form element above? Or is it related to the PR that I linked earlier? Also does it mirror the original behaviour, I could have sworn that the data only included up to the first form element?

modules/system/assets/js/snowboard/extras/StylesheetLoader.js
```diff
@@ -0,0 +1,36 @@
...
+        };
+    }
+
+    ready() {
+        let stylesLoaded = false;
+
+        // Determine if stylesheet is already loaded
+        document.querySelectorAll('link[rel="stylesheet"]').forEach((css) => {
+            if (css.href.endsWith('/modules/system/assets/css/snowboard.extras.css')) {
+                stylesLoaded = true;
```
LukeTowers, line 25: Is there a way to break out of the loop early once its found the target?

modules/system/assets/js/snowboard/main/Snowboard.js
```diff
@@ -0,0 +1,342 @@
...
+     * @param {string} name
+     * @returns {void}
+     */
+    removePlugin(name) {
+        const lowerName = name.toLowerCase();
+
+        if (!this.hasPlugin(lowerName)) {
+            /* develblock:start */
+            this.debug(`Plugin "${name}" already removed`);
+            /* develblock:end */
```
LukeTowers: This leftover?

LukeTowers (changes requested):

modules/system/assets/js/snowboard/abstracts/PluginBase.js
bennothommo: Not technically - `this.snowboard` is a reference to the global Snowboard instance, but is injected into all plugin instances like Laravel does with the Application class. The standard is for properties to be camelCase, so `snowboard` is correct here.

modules/system/assets/js/snowboard/ajax/Request.js
bennothommo: I did think about this the other day. I think I would be happy with allowing remote calls using Request, but not necessarily accepting partial changes from them or allowing them to directly impact the app - so you could use them to retrieve data from, say, a JSON API but then would need to handle the data and page manipulations yourself.

modules/system/assets/js/snowboard/ajax/Request.js
bennothommo: Yeah, makes sense.

modules/system/assets/js/snowboard/ajax/Request.js
LukeTowers: I'm not sure why we would care about restricting based on that, presumably if the developer is making calls to an external service they accept the risk. Also of note my example was loading the js from example.com on notexample.com, example.com being the winter instance that was both the host of the js and the target url, while notexample.com was just pulling in the framework and sending Ajax requests to example.com in order to have the power of the Ajax framework in a WordPress frontend

modules/system/assets/js/snowboard/ajax/Request.js
bennothommo: I've added a utility that wraps the [`js-cookie`](https://github.com/js-cookie/js-cookie) library, as it's super lightweight and has a much better API.

modules/system/assets/js/snowboard/ajax/Request.js
bennothommo: Alright, fair enough.

The URL was modifiable through the options, but the `fetch()` config was not. I've just pushed an update to allow the fetch config to be modified via the `options`, or through a global event.

modules/system/assets/js/snowboard/ajax/handlers/AttributeRequest.js
bennothommo: Yep, exactly. It includes the data from any parent element in the final "compiled" data.

The previous behaviour was here: https://github.com/wintercms/winter/blob/develop/modules/system/assets/js/framework.js#L94-L96

As far as I can tell, it got the data from any parent element and didn't stop at the nearest form.

modules/system/assets/js/snowboard/extras/StylesheetLoader.js
bennothommo: @LukeTowers no, there's no way to break out of a `forEach` loop, but given it's a simple string comparison, I didn't worry too much about that.

LukeTowers:
@bennothommo is the workshop-theme up to date with the latest changes?

modules/system/assets/js/snowboard/utilities/Cookie.js
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    set(name, value, options) {
+        return BaseCookie.set(name, value, options);
+    }
+
+    remove(name, options) {
+        return BaseCookie.remove(name, options);
+    }
+}
```
LukeTowers, line 136: Is there support for the `withAttributes()` method documented here: https://github.com/js-cookie/js-cookie/#setting-up-defaults?

LukeTowers: And Converters? https://github.com/js-cookie/js-cookie/#converters

modules/system/assets/js/snowboard/utilities/Debounce.js
```diff
@@ -0,0 +1,18 @@
...
+            cancelAnimationFrame(frame);
+        }
+
+        // Queue our function call for the next frame
+        frame = requestAnimationFrame(() => {
+            // Call our function and pass any params we received
+            fn(...params);
+        });
+    };
+};
```
LukeTowers: What does this plugin actually do / is it actually being used? I couldn't find any references to it beyond where it gets registered and added.

modules/system/assets/js/snowboard/utilities/Cookie.js
bennothommo: I'll probably replace Converters with global events, but I'll add the `withAttributes()` support.

modules/system/assets/js/snowboard/utilities/Debounce.js
bennothommo: I thought it was used by the Transition utlity, but apparently not (although it should).

This utility is synonymous with the Vue's `nextTick` feature - essentially allowing you to create a callback that will fire on the next "frame". It's useful for things like event handling and DOM manipulation where you need to ensure that everything in the previous frame is completed before running the callback.

For example, in the Transition library, a particular class is available for one frame to ensure that CSS transition is correctly initialised. We would use a debounce to remove the class in the next frame so the transition can take effect. This is similar to Vue transitions.

bennothommo:
> @bennothommo is the workshop-theme up to date with the latest changes?

I believe so :)

modules/system/assets/js/snowboard/utilities/Cookie.js
LukeTowers: Sounds good, we'll just want to update the docs for it then

modules/system/assets/js/snowboard/utilities/Debounce.js
LukeTowers: Gotcha, if you could update the transition to use it and then add some docs to the Utilities page that I added earlier today for it that would be great! Finish line is in sight, I'm pretty excited for this to be released!

modules/system/assets/js/snowboard/utilities/Cookie.js
```diff
@@ -0,0 +1,136 @@
...
+     * @param {Object} options
+     * @returns {String}
+     */
+    set(name, value, options) {
+        let saveValue = value;
+
+        // Allow plugins to override the value to save
+        this.snowboard.globalEvent('cookie.set', name, value, (newValue) => {
+            saveValue = newValue;
+        });
```
LukeTowers, line 113: Another thought that popped into my head, do we want to provide the ability to change the options data in the global event?

modules/system/assets/js/snowboard/utilities/JsonParser.js
```diff
@@ -0,0 +1,397 @@
+import Singleton from '../abstracts/Singleton';
+
+/**
+ * JSON Parser utility.
+ *
+ * This utility parses JSON-like data that does not strictly meet the JSON specifications in order to simplify development.
+ * It is a safe replacement for JSON.parse(JSON.stringify(eval("({" + value + "})"))) that does require the use of eval()
```
LukeTowers:
```suggestion
 * It is a safe replacement for JSON.parse(JSON.stringify(eval("({" + value + "})"))) that does not require the use of eval()
```

modules/system/assets/js/snowboard/utilities/Cookie.js
bennothommo: Hmmm, I think that's a bit too much over-reach for a plugin. Remember that these events are just to replace the value conversion feature in `js-cookie`.

What scenario do you envisage that it will be beneficial to modify the options?

LukeTowers: TBH I'm not sure, I don't do enough work with cookies in the frontend; I was just thinking for the sake of completeness. Is it a breaking change to implement additional parameters to the event in JS down the road if we decide to do so?

bennothommo: Nope, the listener method (the one that receives the event) does not need to accept all the parameters. It could be set up to, for example, only accept the name and value parameter in the event above.

LukeTowers: Sounds good to me, we can re-evaluate adding it later then.

LukeTowers:
**comment moved to #440**

LukeTowers (approved):

PR #404, opened 2022-01-14
[Enhancement] Refactor winter:env, winter:install to use new EnvFile & ArrayFile parsers

LukeTowers:
Related to wintercms/storm#40

TODO:
- [x] Refactor mail config mirroring to work like the database one now since the default mail config structure has changed.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/404"></a>


LukeTowers:
@jaxwilko are these tests only failing because it's not pulling in the correct version of Storm?

jaxwilko:
@LukeTowers should be yes, storm can get merged first as it doesn't overwrite anything in use, then this branch should run the tests fine :)

config/app.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 return [
-
+
```
LukeTowers: @jaxwilko This whitespace should be removed by our printer

config/app.php
```diff
@@ -244,14 +243,12 @@
     | this array to grant expanded functionality to your applications.
     |
     */
-
-    'providers' => array_merge(include(base_path('modules/system/providers.php')), [
-
+    
+    'providers' => array_merge(include base_path('modules/system/providers.php'), [
```
LukeTowers: @jaxwilko looks like the printer / parser barfs on nested functions, we'll need to support that

config/app.php
```diff
@@ -244,14 +243,12 @@
...
         // 'Illuminate\Html\HtmlServiceProvider', // Example
-
-        'System\ServiceProvider',
+        'System\\ServiceProvider',
```
LukeTowers: I'm not sure why this was changed, is it a fluke that the original value worked? Or is the printer misunderstanding that backslashes are only required to be escaped in single quote strings if they're right next to a value that would otherwise have to be escaped (i.e. a single quote)?

config/auth.php
```diff
@@ -1,39 +1,41 @@
...
-        | in to the Backend before being blocked for a specified number of minutes.
-        |
-         */
+                |--------------------------------------------------------------------------
+                | Enable throttling of Backend authentication attempts
+                |--------------------------------------------------------------------------
+                |
+                | If set to true, users will be given a limited number of attempts to sign
+                | in to the Backend before being blocked for a specified number of minutes.
+                |
```
LukeTowers: @jaxwilko indentation is borked here, not sure why.

config/logging.php
```diff
@@ -42,74 +42,66 @@
...
+            'level' => env('LOG_LEVEL', 'critical'),
+            'url' => env('LOG_SLACK_WEBHOOK_URL'),
             'username' => 'Winter Log',
         ],
-
         'papertrail' => [
-            'driver'       => 'monolog',
-            'handler'      => env('LOG_PAPERTRAIL_HANDLER', \Monolog\Handler\SyslogUdpHandler::class),
+            'driver' => 'monolog',
+            'handler' => env('LOG_PAPERTRAIL_HANDLER', \Monolog\Handler\SyslogUdpHandler::class),
```
LukeTowers, line 74: @jaxwilko Laravel actually imports these classes at the top of their files (see https://github.com/laravel/laravel/blob/9.x/config/app.php#L3), which might be a setup that we could encounter (thus failing the file must start with return requirement), is it easy to support that or should we just say that we don't support it?

config/mail.php
```diff
@@ -32,53 +32,46 @@
...
-            'port'       => env('MAIL_PORT', 587),
-            'timeout'    => null,
-            'transport'  => 'smtp',
-            'username'   => env('MAIL_USERNAME'),
+            'host' => env('MAIL_HOST', 'smtp.mailgun.org'),
+            'password' => env('MAIL_PASSWORD'),
+            'port' => env('MAIL_PORT', 587),
+            'timeout' => null,
+            'transport' => 'smtp',
+            'username' => env('MAIL_USERNAME'),
```
LukeTowers, line 44: @jaxwilko just to confirm, is trailing commas enforced in the ArrayPrinter?

config/session.php
```diff
@@ -123,12 +125,9 @@
...
     |
     */
-
-    'cookie' => env(
-        'SESSION_COOKIE',
-        str_slug(env('APP_NAME', 'winter'), '_').'_session'
-    ),
-
+    
+    'cookie' => env('SESSION_COOKIE', str_slug(env('APP_NAME', 'winter'), '_') . '_session'),
```
LukeTowers, line 129: Looks like this one came across properly, although it would be nice if it didn't eat the formatting. Easy fix or no?

config/app.php
jaxwilko: When you pass a variable it'll pass the value of that variable, i.e. if you were to pass `CURLOPT_CRLF` it would set the value 27.

To set a namespace you'd either need to use `\\\` as a separator or you could use the `constant()` method and set the full class reference, i.e. `Path\Name\MyClass::class`

config/logging.php
jaxwilko: Should be fairly easy to check for and ignore imports prior to a return stmt

config/mail.php
jaxwilko: It was at some point yes, if you check the early posts on the PR there was an example of it, it'd be worth looking at the tests to confirm though

config/session.php
jaxwilko: You'd have to embed custom formatting into the printer, which, ehh, we can have a look but no promises 😂

LukeTowers:
Blocked by comments, especially the issue with the `include()` broken parsing / rendering.

config/logging.php
jaxwilko: Support added in https://github.com/wintercms/storm/pull/40/commits/ec040435c486b0a5cb369137cf3d5dfa2d932c2e

config/mail.php
jaxwilko: Yes, it is

config/app.php
jaxwilko: Added with backtracing to ensure includes were () correctly dependent upon position: https://github.com/wintercms/storm/pull/40/commits/c098853f8e082e4a29e73c8dc21d873195337777
Removed above and simply always appended (): https://github.com/wintercms/storm/pull/40/commits/892ed2a1dd87e4bb75adce332fe96288fcf16260

LukeTowers:
Blocked by remaining TODO item and https://github.com/wintercms/storm/pull/40#issuecomment-1068693492

PR #405, opened 2022-01-16
Updated richeditor.js initFroala function to not overwrite the follow…

Nimdoc:
### Changes
Updated richeditor.js initFroala function to not overwrite the following editor option defaults if they exist:

- imageStyles
- linkStyles
- paragraphStyles
- paragraphFormat
- tableStyles
- tableCellStyles
- htmlAllowedTags
- htmlAllowedAttrs
- htmlDoNotWrapTags
- htmlRemoveTags
- lineBreakerTags

Updated build-plugins-min.js with lastest minified js.

### Issue
I wanted to customize the backend editor to add a couple custom image styles to. I searched around and have seen that others have tried to customize the backend editor for their purposes too. However some of the editor options (partial list above) are overridden when the editor is initialized. An example of a customization I wanted to have in the editor was an img-max-100 style for images to limit their size to 100%, and I wanted to put that in the froala defaults with the following code:

```
+function ($) { "use strict";
    $(document).render(function() {
        if ($.FroalaEditor) {
            $.FroalaEditor.DEFAULTS.imageStyles['max-img-100'] = 'Max 100%';
        }        
    })
}(window.jQuery);```

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/405"><img src="https://gitpod.io/button/open-in-gitpod.svg"/></a>


LukeTowers:
@Nimdoc this makes more sense now; did you try to use the Backend Editor Styles settings page to add your customizations?

Screen Shot 2022-01-16 at 12 42 27 PM

Nimdoc:
Huh, I didn't see that, that's even better than what I wanted. Well at least if someone searches for "WinterCMS add editor styles" or what not they'll see this. Thanks for the help!

Nimdoc:
@LukeTowers I got more time to look into customizing the editor with your suggestion, but I realized now that the settings are stored in the database. I think perhaps these changes may actually be useful for:

1. Keeping editor settings between environments without worrying about setting up the database.
2. Letting plugin developers add custom values to the froala editor for their purposes.

Or is there a better way to add styles with php like somewhere with [EditorSetting.php](https://github.com/wintercms/winter/blob/develop/modules/backend/models/EditorSetting.php) or [RichEditor.php](https://github.com/wintercms/winter/blob/develop/modules/backend/formwidgets/RichEditor.php)?

LukeTowers:
Hmm, this is worth considering further. We could look at providing control over those values through a config file like we do with the brand settings currently; or we could perhaps add support to themes (since themes are ultimately responsible for rendering a lot of richeditor content; although not solely responsible so we'd have to take that into consideration as well.

@bennothommo @jaxwilko any thoughts on this?

bennothommo:
I think the theme providing a baseline which can then be extended from in the Backend would probably be best. You generally use richeditors for theme content.

I'd probably do it in the following:
- The theme provides a list of classes for each of these configs - these could be displayed in the Backend but are immutable. If you want to change them, you change the theme settings instead.
- The Backend defines additional classes which can be added, modified and removed like they can now.
- A new config option for richeditors could be added to define whether the theme-provided classes are made available (`useThemeClasses`?) or not, to cater for the richeditors that are not modifying front-end content.

bennothommo:
We've got an [open issue](https://github.com/wintercms/winter/issues/536) to allow the theme to provide richeditor settings, which I feel should cover the scenario that this PR intends to rectify, so I'm going to close this for now. @Nimdoc if you feel your PR covers another scenario, please let me know and I'll re-open.

PR #415, opened 2022-01-20
Improved Russian translation

WebVPF:
New value and fix commas.


bennothommo:
@WebVPF loving the clean-up and translation work you're doing! Keep it up!

PR #417, opened 2022-01-22
VSCode recommended extensions

WebVPF:
Documentation and description https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

For VSCode we recommend installing the extension **Winter CMS Template Language** https://marketplace.visualstudio.com/items?itemName=JosephCrowell.wintercms-template-language

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/417"></a>

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+	"recommendations": [
+		"josephcrowell.wintercms-template-language"
+	]
+}
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,5 +1,5 @@
 {
-	"recommendations": [
-		"josephcrowell.wintercms-template-language"
-	]
+    "recommendations": [
+        "josephcrowell.wintercms-template-language"
+    ]
 }


.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+	"recommendations": [
+		"josephcrowell.wintercms-template-language"
+	]
+}
```
LukeTowers:
```suggestion
{
    "recommendations": [
        "josephcrowell.wintercms-template-language"
    ]
}
```

LukeTowers:
I'd feel more comfortable about doing this if the extension was a codebase we controlled under the Winter CMS organization, @bennothommo any thoughts?

WebVPF:
This entry does not require the mandatory installation of the extension. It just pops up once, which is recommended. Once when opening a project for the first time. Unobtrusively.
The extension will also be visible in the recommended extensions block.

If this extension is already installed in VSCode, then there will be no messages.

In the language association for some files, we have indicated `wintercms`. I think it's a good idea to recommend an extension that works with the `wintercms` language (ini, php, twig in one file).

bennothommo:
@LukeTowers @WebVPF personally, I would prefer we keep this sort of config out of the main repo, as it's clutter for people who *don't* use VSCode... however, this will be handy for Gitpod.

PR #422, opened 2022-01-26
Improved Russian translation

WebVPF:


WebVPF:
Changing translation of "layout" word in Russian: https://github.com/octobercms/october/pull/5291

bennothommo:
@WebVPF just out of curiousity, what is the specific reason for changing "Content" and "Assets" back to English?

WebVPF:
To avoid misleading users.
It is better to leave the original folder names from the English language for these menu items. Translation into Russian only makes things worse.
In Oсtober 2 the problem with naming was solved by removing the menu by making the folder-file structure.

This is not a spontaneous decision to change to English. This has been discussed many times before in the Russian-speaking community of October (even before the fork).

Until the release is released, I decided to fix it.

bennothommo:
Alright - I'll merge this then, but will leave a note in the commit linking to your comment, given that there is the possibility that someone else may see this as an untranslated string in the future and submit a PR to translate them back.

PR #444, opened 2022-02-10
Add automatic package registration and improved installation to Laravel Mix support

LukeTowers:
Refs: #440, #401.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/444"></a>


LukeTowers:
@bennothommo with these changes all that's really needed from the package registration is the relative path to the `winter.mix.js` file from the base path. Do you see any value in continuing to have an array of values (path, package.json, winter.mix.js) tied to the package name instead of just the one relative path?

Personally I'm thinking not, even if we wanted to expand our dependency checking logic in the future to look at related package.json files it would be better to autogenerate the path to those files instead of have separate config items for it since supporting them being set values in an array opens us up to the possibility that they could not match the mixJs path at some point in the future.

modules/system/ServiceProvider.php
```diff
@@ -570,7 +570,7 @@ protected function registerAssetBundles()
         });
 
         MixAssets::registerCallback(function ($mix) {
-            $mix->registerPackage('snowboard', '~/modules/system/assets/js/snowboard');
+            $mix->registerPackage('snowboard', '~/modules/system/assets/js/snowboard/winter.mix.js');
```
bennothommo, line 573: If we're going to be assuming a package has a `winter.mix.js` file, should we need to define that filename when registering the package? I think, even in the case of multiple packages under the one extension, we should be fine with assuming the end target is `winter.mix.js`, therefore they only need to specify the path to the directory where that file resides.

modules/system/classes/MixAssets.php
```diff
@@ -72,26 +83,69 @@ public function init()
...
+        }
 
-                    $this->registerPackage($name, $pinfo['dirname'], $pinfo['basename']);
+        // Search themes for Mix packages to autoregister
+        if (in_array('Cms', $enabledModules)) {
+            $themes = Theme::all();
+            foreach ($themes as $theme) {
+                $path = $theme->getPath() . '/' . $this->mixJs;
+                if (File::exists($path)) {
+                    $packagePaths["theme-" . $theme->getId()] = $path;
```
bennothommo, line 134: Do we want to allow unused / disabled themes to be built as well? I think for the most part, we'd only want to build the active theme.

modules/system/console/MixCompile.php
```diff
@@ -33,21 +38,41 @@ class MixCompile extends Command
      */
     public function handle(): int
     {
+        // Exit early if node_modules isn't available yet
+        if (!File::exists(base_path('node_modules'))) {
+            $this->error('node_modules is not available, try running mix:install first.');
```
bennothommo:
```suggestion
            $this->error('The Node dependencies are not available, try running mix:install first.');
```

modules/system/console/MixCompile.php
```diff
@@ -95,40 +168,52 @@ protected function mixPackage($package)
...
     {
+        $basePath = base_path();
         $command = $this->argument('webpackArgs') ?? [];
         array_unshift(
             $command,
-            $package['path'] . implode(DIRECTORY_SEPARATOR, ['', 'node_modules', 'webpack', 'bin', 'webpack.js']),
+            $basePath . '/node_modules/webpack/bin/webpack.js',
             '--progress',
-            '--config=' . $package['path'] . DIRECTORY_SEPARATOR . 'mix.webpack.js'
+            '--config=' . $this->getWebpackJsPath($mixJsPath)
```
bennothommo, line 187: Does this work on Windows? While PHP might normalise paths, I'm not sure that the command line (or Process) does?

LukeTowers: Php doesn't normalize paths, windows does.

modules/system/classes/MixAssets.php
LukeTowers: I don't see why not. Personally for our multisite usage of winter we'll have a theme for every site and yet winter considers only one of them "active" at a time. There's always the p filter option

modules/system/console/MixCompile.php
bennothommo: I'd still want this to be tested - I distinctly remember Windows being a pain with pathing when I was writing the PathResolver class, so there's some contexts where Windows doesn't play nice with Unix paths.

modules/system/ServiceProvider.php
LukeTowers: Yeah I went back and forth on this but ultimately what's the point in allowing them to drop the file? It's a mere 13 characters difference one time to write it out for much greater clarity, especially when searching the project for uses of the file. Additionally this manual registration is only required when the developer is going outside the norm anyways because of auto registration.

modules/system/console/MixCompile.php
LukeTowers: One of our devs runs windows so I'm fine with releasing and making him test it and then fixing it if it doesn't work 😂

LukeTowers: The path issues mostly come up when we are processing paths that have come from windows since they don't match Unix but sending paths to windows usually works fine

modules/system/ServiceProvider.php
bennothommo: Fair enough. If they forget to add a file to the path, are we going to assume they were targeting the Mix file?

LukeTowers: No, I played around with that but it was causing the code to be messier and more complex than I liked. Currently it will complain if a valid .js file isn't provided which I think is fine, if you're going with convention (auto registration) then it's easy for you, if you're breaking convention then you need to be explicit. Comes back to the "searching the project for uses of the file" piece, the convention is clearly documented and easy to understand, but supporting half-baked manual registrations is of no benefit to anyone IMO. It was a good solution for the initial approach, but now that we have autoregistration it doesn't really serve any purpose to support that IMO.

Screen Shot 2022-02-13 at 10 49 19 AM

modules/system/classes/MixAssets.php
```diff
@@ -59,8 +69,9 @@ public function init()
         *   public function registerMixPackages()
         *   {
         *       return [
-        *           'package-name-1' => 'winter-mix.js',
+        *           'package-name-1' => 'winter.mix.js',
         *           'package-name-2' => 'assets/js/build.js',
+        *           'package-name-3' => 'assets/css', // winter.mix.js is assumed to be the config file in this path
```
bennothommo: @LukeTowers per [your comment](https://github.com/wintercms/winter/pull/444/files#r805385923) here regarding not assuming a file-less path to be pointing to the `winter.mix.js` file, we'll need to remove this example.

```suggestion
```

bennothommo:
@LukeTowers I started writing a PHP class to handle `package.json` manipulation for my `wip/snowboard-workspaces` branch here: https://github.com/wintercms/winter/commit/e33fa94d3f4afa4fba74f716fbd6ecb9bf082ef2

Would that be at all useful for this? Might be able to move some of that procedural code into this class.

LukeTowers:
@bennothommo I was looking at that but I didn't really see the point to be honest. It felt a bit too much "enterprisey" without a clear reason for its existence or usefulness. Could you provide an example of how we could improve this code by using that class and / or what else that class could be used for?

PR #453, opened 2022-02-16
Update settings.json

WebVPF:
`components/partials/file.htm` ===> https://github.com/wintercms/winter/blob/21e6783165a487df9933c3e7da2dfc8a64f2bf6a/.vscode/settings.json#L14

This rule applies to `components/partials/folder/file.htm`

---
## Question

Tell me, is there an ini block in the component partials files?

```
description = "qwerty"
==
<p>Partial content</p>
```

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/453"></a>


.vscode/settings.json
```diff
@@ -12,6 +12,7 @@
 
         "**/plugins/*/*/behaviors/*/partials/*.htm": "php",
         "**/plugins/*/*/components/*/*.htm": "wintercms-twig",
+        "**/plugins/*/*/components/partials/*/*.htm": "wintercms-twig",
```
LukeTowers: Can the above line be changed to `"**/plugins/*/*/components/**/*.htm": "wintercms-twig",` and work the same?

WebVPF: Yes. So it's more correct.

LukeTowers (approved):

PR #469, opened 2022-02-22
Fix taglist default value in relation mode

### LINKED ISSUE #467: Taglist in relation mode don't support default values
### Winter CMS Build

1.1

### PHP Version

7.4

### Database engine

SQLite

### Plugins installed

_No response_

### Issue description

When you have a taglist in relation mode, you can't set default values.

### Steps to replicate

- Create 2 models, with a manyToMany relationship between them (eg : Post and Author)
- In post model, add that field in your fields.yaml
        authors:
        label: Authors
        type: taglist
        mode: relation
        nameFrom: lastname
        placeholder: Please choose
        default: [1]
- Create an Author entry, make sure it's id is = 1
- Create a Post entry, you will see that the Author taglist is empty instead of seleting the Author you just created

I also tried to extend Post form fields and model like this : 

    public function formExtendFields($form) {
        if($form->getContext() == 'create') {
            $form->getField('title')->value = "Test"; // work
            $form->getField('authors')->value = ['1']; // don't work
            // $form->getField('authors')->value = ['Doe']; // don't work
        }
    }

Also tried this : 

    public function formExtendModel($model) {
        if ($this->formGetContext() === 'create') {
            $model->title = 'Super'; // Work
            $author = Author::findOrFail(1);
            $model->authors->push($author); // Dont work
        }
    }

In both case, I tried to reproduce the $form or $model object compared to the ones I dumped when editing existing Post. Each time, the Authors taglist is empty.

Thank you

Alex

### Workaround

_No response_

mjauvin:
Fixes #467 

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/469"></a>


mjauvin:
@LukeTowers if you get time to merge this, someone was eager for the fix... ;)

Since it's applied against wip/1.2, they'll need to test the new code to use the fix...

LukeTowers:
@jaxwilko @bennothommo @mjauvin it might be better if we supported taglist as a mode for the relation widget rather than supporting the relation mode on the taglist widget, any thoughts?

mjauvin:
@LukeTowers I don't know, `taglist` formwidgets already have a relation mode, this PR is a simple fix for its shortcoming in respect to setting a default value.

I have no problem if we add taglist mode to `relation` formwidget, but this should be addressed seperately in my opinion.

PR #471, opened 2022-02-23
Move scaffold commands

### LINKED ISSUE #270: Move scaffolding commands into System module
### Package targeted

Both

### Description

Currently, the scaffolding commands in Winter CMS (`create:plugin`, `create:controller`, etc.) reside in the Storm library for historical purposes.

It may, however, make more sense for these to reside in the System module in Winter CMS, as these functions relate directly to Winter CMS functionality, and require a Winter CMS install to be used effectively. They would not be able to be used in isolation with the Storm library.

If accepted, I intend to put forward a PR that moves all the `create:*` commands into the System module, but leaving the base Scaffolding classes and Service Provider within the Storm library.

The main benefit of this change will be that we can write test cases for the scaffold commands, as currently these are not tested within the Storm library, ostensibly because it's impossible to load the plugin architecture within the Storm library tests.

### Will this change be backwards-compatible?

Mostly backwards-compatible, as most people will just use the commands directly, but it would break any custom implementations of these commands which may extend the original command.

bennothommo:
Fixes #270.

Moves the Scaffolding commands out of Storm, and into the modules that the scaffold generates code for.

As discussed in the issue above, this makes more sense to keep Storm agnostic, and allows for the commands to be hidden if a particular module is not used in a Winter instance.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/471"></a>

### DIFF
--- a/modules/backend/ServiceProvider.php
+++ b/modules/backend/ServiceProvider.php
@@ -20,6 +20,7 @@ class ServiceProvider extends ModuleServiceProvider
      */
     public function register()
     {
+        $this->registerConsole();
         $this->registerMailer();
         $this->registerAssetBundles();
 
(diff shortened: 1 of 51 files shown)


LukeTowers:
Thanks @bennothommo!

PR #474, opened 2022-02-25
[FIX] Switched file to storage for plugin disabled

jaxwilko:
Simple fix to allow for `disabled.json` to use the storage facade rather than file, this creates support for read only filesystems

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/474"></a>


mjauvin:
Elegant!

PR #485, opened 2022-03-08
Undefined variable $packagePath

marcomessa:
When the package is not present in package.json, the correct error is not thrown because $packagePath not exists. I'm not sure what kind of feedback you would like to send, with this edit it'll be something like:

"Unable to watch "packagename", packagename was not found in the package.json's workspaces.packages property. Try running mix:install first."

Cheers!

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/485"></a>

### DIFF
--- a/modules/system/console/MixWatch.php
+++ b/modules/system/console/MixWatch.php
@@ -45,7 +45,7 @@ class MixWatch extends MixCompile
         $relativeMixJsPath = $package['mix'];
         if (!$this->canCompilePackage($relativeMixJsPath)) {
             $this->error(
-                sprintf('Unable to watch "%s", %s was not found in the package.json\'s workspaces.packages property. Try running mix:install first.', $name, $packagePath)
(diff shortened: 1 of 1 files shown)


modules/system/console/MixWatch.php
```diff
@@ -45,7 +45,7 @@ public function handle(): int
         $relativeMixJsPath = $package['mix'];
         if (!$this->canCompilePackage($relativeMixJsPath)) {
             $this->error(
-                sprintf('Unable to watch "%s", %s was not found in the package.json\'s workspaces.packages property. Try running mix:install first.', $name, $packagePath)
+                sprintf('Unable to watch "%s", %s was not found in the package.json\'s workspaces.packages property. Try running mix:install first.', $name, $name)
```
LukeTowers:
```suggestion
                sprintf('Unable to watch "%s", %s was not found in the package.json\'s workspaces.packages property. Try running mix:install first.', $name, $relativeMixJsPath)
```

bennothommo:
Thanks @marcomessa!

PR #494, opened 2022-03-14
Convert afterFetch() method to an event handler for model.afterFetch

mjauvin:
This change allows a plugin to handle the `model.afterFetch` event AFTER the Model handler.

The current code uses the `afterFetch()` method which always runs AFTER any event handler.

ref. https://github.com/wintercms/storm/blob/develop/src/Database/Model.php#L162-L168

The alternative would be to modify the code referenced above in the base Database/Model, but that could lead to breakage in backward compatibility.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/494"></a>

### DIFF
--- a/modules/cms/models/ThemeData.php
+++ b/modules/cms/models/ThemeData.php
@@ -53,6 +53,16 @@ class ThemeData extends Model
     protected static $instances = [];
 
     /**
+     * Constructor
+     */
+    public function __construct(array $attributes = [])
+    {
+        parent::__construct();
+
+        $this->bindEvent('model.afterFetch', [$this, 'afterFetchHandler']);
+    }
+
+    /**
      * Before saving the model, strip dynamic attributes applied from config.
      * @return void
      */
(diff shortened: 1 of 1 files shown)


modules/cms/models/ThemeData.php
```diff
@@ -52,6 +52,16 @@ class ThemeData extends Model
      */
     protected static $instances = [];
 
+    /**
+     * Constructor
+     */
+    public function __construct(array $attributes = [])
+    {
+        parent::__construct();
```
LukeTowers: Why not do this in afterBoot instead?

LukeTowers: also if you're overriding the constructor you need to pass along the provided arguments

mjauvin: Good points!

damsfx:
My test did not work.

Steps to reproduce : 
 - New installation of WinterCMS
 - Edit the Demo theme to include a field
 ```yaml
 form:
    fields:
        header_images:
            label: Header images
            span: left
            type: fileupload
            mode: image
            imageWidth: 200
            imageHeight: 200
 ```
 When going to theme settings, an error occurs :   
  
  **Model 'Cms\Models\ThemeData' does not contain a definition for 'header_images'.**

mjauvin:
Closed in favor of #495

Issue #498, opened 2022-03-16
Redevelop the Inspector widget


bennothommo:
The [Inspector](https://wintercms.com/docs/ui/inspector) widget in Winter CMS is one of the many unique and powerful features of the framework, providing contextual "popup" forms for various elements, most prominently in the configuration of CMS components, but also used very liberally throughout some first party plugins like the Builder plugin.

![image](https://user-images.githubusercontent.com/15900351/158505358-5c7d5ee2-4ab9-499a-8f6e-01b43a5fdc5d.png)

The widget has remained virtually unchanged throughout the history of Winter CMS and October CMS before it due to its complexity (beyond a rewrite to support features in the Builder plugin), therefore we have determined that it is due an upgrade and refactor to simplify its usage. This issue will detail the goals that we intend to achieve with redeveloping this widget.

## Primary Goals

- **Rewrite the widget using Vue and Snowboard:** A lot of the functionality of this widget lends itself to using Vue. The current Inspector has a lot of boilerplate to handle things like templating, validation, data manipulation that could be done much easier using Vue's powerful templating and data management. Using our new Snowboard plugin and architecture, we can rewrite the Inspector to take advantage of Vue.
- **Document the entire front-end and back-end API:** The docs for using or implementing the Inspector in your own plugins have been traditionally lacking. The documentation linked above only deal with the "data attributes" API of using the Inspector in a HTML template, with no real guidance towards handling the backend functionality beyond some sporatic code examples. We will tidy up and fully document this widget to allow people to harness the full potential of the Inspector.
- **Make form configuration consistent with the Form widget:** The schema for the Inspector uses a custom schema which shares some common attributes to the Form schema, but also introduces some potential gotchas - for example, a `text` field in a Form is a single-line input, but in the Inspector, it is a multi-line input. To keep the development experience consistent, we will adopt the Form schema for all fields available in the Inspector, whilst keeping backwards-compatibility with the original schema.

## Secondary Goals

- **Redesign the widget:** While the widget does look great as it is currently, we feel it can be so much more. Using the power of single-file components in Vue, we will be able to construct several layouts for the widget to allow full customisation of the look and feel of the Inspector.
- **Ensure full keyboard accessibility:** The way that the widget is currently formatted implies that the widget should be easily fillable using only keyboard navigation, but unfortunately, this is not entirely the case at the moment. We will put efforts into making sure that the entire Inspector form can be tabbed and entered using only the keyboard.
[...]

PR #499, opened 2022-03-16
Redevelop the Inspector widget

bennothommo:
Please see #498 for more information on this PR, and keep this thread only for discussing the technical implementation.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/499"></a>


modules/system/console/fixtures/mix.webpack.js.fixture
```diff
@@ -11,26 +11,26 @@ module.exports = async () => {
 
     mix.listen('init', function (mix) {
         // Setup Winter path aliases
-        mix._api.alias({
+        mix.api.alias({
```
LukeTowers, line 14: What's the reasoning for this change?

bennothommo: I was getting errors using `mix:watch` as soon as I introduced Vue compiling. Turns out that `api` is the correct variable to use for this, and `_api` is the internal reference. The `api` variable actually creates the `_api` variable if it doesn't exist, so I believe that perhaps when Vue compiling is used, the Mix initialisation happens earlier.

LukeTowers: Huh. Is this something that we should fix in a main branch right now?

bennothommo: I'd say so, but I've just done it here for now so I can be sure it's working.

jaxwilko:
@what-the-diff 🤔

PR #500, opened 2022-03-17
Sync MailSettings model init with L9 mail config changes

mjauvin:
L9 mail config is no longer a flat structure.

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/500"></a>

### DIFF
--- a/modules/system/models/MailSetting.php
+++ b/modules/system/models/MailSetting.php
@@ -55,16 +55,17 @@ class MailSetting extends Model
     public function initSettingsData()
     {
         $config = App::make('config');
-        $this->send_mode = $config->get('mail.driver', static::MODE_MAIL);
+        $mailers = $config->get('mail.mailers', []);
+        $this->send_mode = $config->get('mail.default', static::MODE_MAIL);
         $this->sender_name = $config->get('mail.from.name', 'Your Site');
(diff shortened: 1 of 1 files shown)


modules/system/models/MailSetting.php
```diff
@@ -55,16 +55,17 @@ class MailSetting extends Model
     public function initSettingsData()
     {
         $config = App::make('config');
-        $this->send_mode = $config->get('mail.driver', static::MODE_MAIL);
+        $mailers = $config->get('mail.mailers', []);
```
LukeTowers:
```suggestion
        $mailers = $config->get('mail.mailers', [
            'sendmail' => ['path' => $config->get('mail.sendmail', '/usr/sbin/sendmail')],
            'smtp' => [
                'host' => $config->get('mail.host'),
                'port' => $config->get('mail.port', 587),
                'username' => $config->get('mail.username'),
                'password' => $config->get('mail.password'),
                'encryption' => $config->get('mail.encryption'),
            ],
        ]);
```

mjauvin: Oh, ok, you're allowing the old config to be used, that's interesting.

LukeTowers: Yes, updating doesn't touch the config files so any time we change them we should attempt to still support the old style.

Issue #493, opened 2022-03-13
Revamp Plugin Management


LukeTowers:
There are a number of minor annoyances with the existing plugin management UX, DX, and API. The following is a dump of my notes on the process so far, this issue will be gradually improved and fleshed out to define my desired changes to the Plugin Management system in Winter and then the work can be done piecemeal, ideally with any potentially breaking changes done first to make it into 1.2.

This also ties into the ClassLoader improvements

### Issues / improvements:
- [x] Get rid of bindContainerObjects (only there because it was matching the updateManager)
- [ ] Simplify public API space, should have a single function that handles loading and initializing all the plugins and if needed a way to reset all of them for unit tests
- [ ] Emphasize that class_exists(Plugin::class) is not an acceptable way to check for plugins since the plugin.php class has to be loaded if it’s present even if it’s disabled
- [x] Simplify the identifier normalization logic and make it more consistent
- [ ] Would be nice to be able to identify the specific circular reference tree that causes issues instead of an arbitrary limit
- [ ] Would be nice to cache the resolved order of plugins to load (so that we don’t have to recalculate plugin dependencies
- [ ] Would be nice to move some more logic (like disablePlugin) to the plugin base class
- [ ] Would be nice to provide some events to plugins
    - plugin.install
    - Plugin.uninstall
    - plugin.disable
    - plugin.enable
- [ ] Updates controller has method called getPluginVersionFile(), returns an array of version update messages keyed by version number, but there’s also a getPluginVersion() method in the plugin base that loads the newest version present in the version.yaml file

### Reasons / ways a plugin can be disabled:
- Plugin is in database but not present in the filesystem
- Plugin has a replacement present that is taking over
- Plugin is a replacement but can't replace the original due to version constraints
- Plugin is explicitly disabled by the user in the database
- Plugin is explicitly disabled by the developer in the config file
- Plugin is implicitly disabled by the system because of unmet dependencies
- Plugin is programatically disabled for the current request only

### Existing API of PluginManager:
- getAllPlugins() -> All plugins that are detected on the filesystem
- getPlugins() -> All plugins that are detected minus plugins that have been flagged disabled
- loadPlugins() -> kicks off the detection and plugin loading process
- loadPlugin($namespace, $path) -> l
    - loads and instantiates the plugin’s Plugin.php
    - Checks if the plugin has been disabled by anything, sets the disabled flag on the plugin if so and ends here
    - sets up class autoloading for plugin classing
    - Sets up plugin replacement if specified
[...]

PR #501, opened 2022-03-17
[WIP] Revamp Plugin Manager

LukeTowers:
See https://github.com/wintercms/winter/issues/493 for context

- [x] use Yaml::parseFile() for pluginDetails instead of parse(file_get_contents())
- [x] Implement custom Cache serviceprovider that is not deferred (https://github.com/wintercms/winter/commit/9314a4442e2695b3665e2fa2e0cf1788809e02d8)
- [x] Implement custom Redis serviceprovider that is not deferred

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/501"></a>


modules/system/controllers/Updates.php
```diff
@@ -845,52 +845,45 @@ public function onBulkAction()
...
                     // Enables plugin's updates.
                     case 'freeze':
-                        $plugin->is_frozen = 1;
+                        $manager->freezePlugin($code);
                         break;
 
                     // Disables plugin's updates.
                     case 'unfreeze':
-                        $plugin->is_frozen = 0;
+                        $manager->unfreezePlugin($code);
```
LukeTowers, line 868: @bennothommo @jaxwilko unfreeze or thaw?

modules/system/classes/PluginManager.php
```diff
@@ -7,9 +7,9 @@
 use File;
 use Lang;
 use View;
+use Cache;
```
LukeTowers, line 10: Requires Cache to not be a deferred service provider because of the Laravel issue with using deferred providers in the registration process of other providers.

modules/system/controllers/Updates.php
bennothommo: "unfreeze" is my vote. Delete actions are "undeleted", for example. It makes it clear it's a reversal of a previous action.

modules/system/classes/PluginManager.php
LukeTowers: @jaxwilko this will require a change to the Cache service provider to not implement the deferred provider interface

modules/system/classes/PluginManager.php
```diff
@@ -398,12 +396,7 @@ public function bootPlugin(PluginBase $plugin): void
      */
     public function getPluginPath(string|PluginBase $plugin): ?string
     {
-        $path = null;
-        $plugin = $this->findByIdentifier($plugin, true);
-        if ($plugin) {
-            $path = $plugin->getPluginPath();
-        }
-        return $path;
+        return $this->findByIdentifier($plugin, true)?->getPluginPath();
```
LukeTowers, line 405: Hmm, not sure I'm a fan of this syntax, could be easily missed. @bennothommo any thoughts?

bennothommo: I have used it before (not in core, in my own projects). It can be easy to miss, but it does take out the need for a lot of conditionals and is great for fluent/chained calls.

modules/system/classes/PluginManager.php
```diff
@@ -154,41 +149,83 @@ public function loadPlugin($namespace, $path)
...
+        $plugins = $loadedPlugins + $configDisabledPlugins;
+
+        return 'system.pluginmanager.state.' . md5(implode('.', $plugins));
+    }
+
+    /**
+     * Loads the plugin flags (disabled & replacement states) from the cache
+     * regenerating them if required.
+     */
+    public function loadPluginFlags()
```
jaxwilko:
```suggestion
    public function loadPluginFlags(): void
```

modules/system/classes/PluginManager.php
```diff
@@ -154,41 +149,83 @@ public function loadPlugin($namespace, $path)
...
+            ];
+        });
+
+        list($this->pluginFlag, $this->replacementMap, $this->activeReplacementMap) = $data;
+    }
+
+    /**
+     * Reset the plugin flag cache
+     */
+    protected function clearFlagCache()
```
jaxwilko:
```suggestion
    protected function clearFlagCache(): void
```

PR #506, opened 2022-03-22
Allow disabling/enabling of sorting with Repeater widget

Flynsarmy:
`Repeater` widget currently supports accepting a `sortable` option but it doesn't actually do anything. This PR removes formwidget sortability when `sortable: false` is passed to it.

I've made the widget sortable by default for backwards compatibility.

### DIFF
--- a/modules/backend/formwidgets/repeater/partials/_repeater_item.htm
+++ b/modules/backend/formwidgets/repeater/partials/_repeater_item.htm
@@ -7,9 +7,11 @@
     class="field-repeater-item">
 
     <?php if (!$this->previewMode): ?>
-        <div class="repeater-item-handle <?= $this->getId('items') ?>-handle">
-            <i class="icon-bars"></i>
-        </div>
+        <?php if ($sortable): ?>
+            <div class="repeater-item-handle <?= $this->getId('items') ?>-handle">
+                <i class="icon-bars"></i>
+            </div>
+        <?php endif; ?>
 
         <div class="repeater-item-remove">
             <button
--- a/modules/backend/formwidgets/repeater/partials/_repeater_item.htm
+++ b/modules/backend/formwidgets/repeater/partials/_repeater_item.htm
@@ -11,7 +11,7 @@
             <div class="repeater-item-handle <?= $this->getId('items') ?>-handle">
                 <i class="icon-bars"></i>
             </div>
-        <?php endif ?>
+        <?php endif; ?>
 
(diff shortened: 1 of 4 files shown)


bennothommo:
@Flynsarmy unfortunately this doesn't work. The sorting functionality is done here in the JS side: https://github.com/wintercms/winter/blob/develop/modules/backend/formwidgets/repeater/assets/js/repeater.js#L79-L85

Removing the two attributes in the HTML just removes the containment of the items, which means child repeater items can be sorted into their parent repeaters, which we obviously don't want :P

Flynsarmy:
Whoops! This version should be better. It removes the sort handles for non-sortable widgets. No need to change the JS. I tested with a two level deep repeater.

bennothommo:
I'd still prefer if we didn't bind the Sortable functionality via the JS side as well if sorting is disabled, so that performance is kept high and there's no chance for any latent sorting functionality to trigger.

Flynsarmy:
Give this one a shot

Flynsarmy:
OK give this version a shot. Took me a bit to get the hang of how your JS works

bennothommo:
I claim no ownership of that JS :P - thanks, will give it a look through shortly.

modules/backend/formwidgets/repeater/partials/_repeater_item.htm
```diff
@@ -11,7 +11,7 @@
             <div class="repeater-item-handle <?= $this->getId('items') ?>-handle">
                 <i class="icon-bars"></i>
             </div>
-        <?php endif; ?>
+        <?php endif ?>
```
LukeTowers: oy what's this then???

mjauvin: yeah, the semicolon is required there! (or is it?)

LukeTowers: Even if it's not it's required for my sanity 😜 Same for the endif in the other file.

bennothommo: But... but... it's used everywhere in the Backend?

image

LukeTowers:
Can we have a quick docs PR @Flynsarmy?

Flynsarmy:
> Can we have a quick docs PR @Flynsarmy?

Done https://github.com/wintercms/docs/pull/65

PR #511, opened 2022-03-29
Fix "inline-options" class for checkbox list field

ArchimedeSolutions:
In checkboxlist fields, `.inline-options` is not working. When applied it search for:
```
.field-checkboxlist:not(.is-scrollable)
```
that is the first container, not acting on checkbox items.

Others `.field-checkboxlist` classes are specifically named to:

```
.field-checkboxlist-scrollable
```

```
.field-checkboxlist-inner
```

So, i changed .less definitions to reflect this naming.

In addition, i added a little margin to single item for a better UI/UX.

 - replaced `:not(.is-scrollable)` selector with direct classes `.field-checkboxlist-scrollable` and `.field-checkboxlist-inner`
 - added margins to single checkbox for improved mobile UI/UX

### DIFF
--- a/modules/system/assets/ui/less/checkbox.less
+++ b/modules/system/assets/ui/less/checkbox.less
@@ -173,12 +173,16 @@
 }
 
 .inline-options {
-    .field-checkboxlist:not(.is-scrollable) {
+    .field-checkboxlist-scrollable {
+        padding-left: 0;
+    }
+
+    .field-checkboxlist-inner {
         padding: 10px 20px 20px 20px !important;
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
@ArchimedeSolutions can you provide a before / after screenshot?

ArchimedeSolutions:
Sure @LukeTowers

Before
Firefox_Screenshot_2022-03-29T17-23-12 357Z

After
Firefox_Screenshot_2022-03-29T17-18-00 497Z

PR #513, opened 2022-03-30
Update _datepicker.htm

AIC-BV:
1. Add timestamp in preview mode.

2. Display the date better in secondaryTabs:
    - Bigger date width
    - Smaller time width

Perhaps not the correct way. Doesn't include AM/PM if this exists.


bennothommo:
@AIC-BV could you include the screenshots you sent through on Discord before, and perhaps an additional one that shows the entire form with this field just so I can see if it looks alright with all other fields around it?

modules/backend/formwidgets/datepicker/partials/_datepicker.htm
```diff
@@ -1,6 +1,6 @@
 <?php if ($this->previewMode): ?>
     <div class="form-control"><?= Backend::dateTime($value, [
-        'format' => $format,
+        'format' => $format . ' H:i',
```
bennothommo: @AIC-BV wouldn't people just change the format if they want the time displayed, as opposed to forcing it onto the end of the string?

LukeTowers: Yes, I don't think we can accept this change without a further justification as to why it is required.

AIC-BV: Datetime in preview mode only shows date, not the time:
image
To me, this feels like a bug because datetime is dateTIME not date, which is why I suggested adding H:i in preview mode.

Possible workaround:
```
created_at@preview:
    label: Created at
    type: datepicker
    mode: datetime
    format: d/m/Y H:i
    default: now

created_at:
    context: ['create', 'update']
    label: Created at
    type: datepicker
    mode: datetime
    format: d/m/Y
    default: now
```

If format created at is d/m/Y H:i, the H:i is also rendered in the date field (seems to be 00:00 in my quick test), and the H:i is also rendered in the time field in expected
image

AIC-BV:
> 
image
image
image

@bennothommo

modules/backend/formwidgets/datepicker/partials/_datepicker.htm
LukeTowers: The correct behaviour would perhaps be to in the FormWidget class itself automatically set the format property if it hasn't been changed from the default when previewMode is true.

bennothommo:
I think this new layout does make sense - the date is generally larger than the time in most instances. Personally, I would like it if it was one field that contained both the date and time, but that's a change for another day.

modules/backend/formwidgets/datepicker/partials/_datepicker.htm
AIC-BV: So how does this continue?
What are the next steps? Do I have to look in to the FormWidget and come up with another solution?

LukeTowers: Instead of making the change here, add some code to DatePicker.php that checks if the format property hasn't been set, it's in previewMode, and the mode is datetime before it sets the format property itself to include the time.

AIC-BV: > if the format property hasn't been set

So in my case, where format is set to `d/m/Y` it won't work and the previewMode is still without time indicator 🤔

LukeTowers: Yes, and you can work around that for your case by setting the context property on your current field to create and update and then add a new field with the format including the time using the @ context suffix method for the preview context

AIC-BV: > created_at@preview:
>     label: Created at
>     type: datepicker
>     mode: datetime
>     format: d/m/Y H:i
>     default: now
> 
> created_at:
>     context: ['create', 'update']
>     label: Created at
>     type: datepicker
>     mode: datetime
>     format: d/m/Y
>     default: now

If I understand you correctly, my only solution remains this?

LukeTowers: That looks good to me yep

AIC-BV:
> I think this new layout does make sense - the date is generally larger than the time in most instances. Personally, I would like it if it was one field that contained both the date and time, but that's a change for another day.

I'm struggeling with github, I never use it except for storing my code 😅 
Can you remove this PR and change (if OK) the column width yourself?

LukeTowers:
@AIC-BV could you resubmit this PR with your desired changes? Looks like all it does at the moment is entirely delete the partial that renders the field.

PR #516, opened 2022-03-30
Add previewMode check

AIC-BV:
RelationControllers display the toolbar buttons in every context.
The toolbar buttons should be hidden when you are previewing the record.
Right now, the toolbar buttons are displayed and you can edit the, for example, hasOne relation while in preview mode.
This PR fixes that problem.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -349,7 +349,7 @@ class RelationController extends ControllerBehavior
             $this->applyExtraConfig($extraConfig);
         }
 
-        if (\Request::is('*/preview/*')) { 
+        if (\Request::is('*/preview/*')) {
             $this->previewMode = true;
         }
 
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -216,6 +216,11 @@ class RelationController extends ControllerBehavior
     public $deferredBinding = false;
 
     /**
+     * @var bool Returns if controller is in preview mode or not
+     */
+    public $previewMode = false;
+    
+    /**
      * Behavior constructor
      * @param Backend\Classes\Controller $controller
      */
@@ -342,6 +347,10 @@ class RelationController extends ControllerBehavior
 
         if ($extraConfig = post(self::PARAM_EXTRA_CONFIG)) {
             $this->applyExtraConfig($extraConfig);
+        }
+
+        if (\Request::is('*/preview/*')) { 
+            $this->previewMode = true;
         }
 
         $this->alias = camel_case('relation ' . $field);
--- a/modules/backend/behaviors/relationcontroller/partials/_toolbar.htm
+++ b/modules/backend/behaviors/relationcontroller/partials/_toolbar.htm
@@ -1,3 +1,4 @@
+<?php if (!$this->previewMode): ?>
 <div data-control="toolbar">
 
     <?php foreach ($relationToolbarButtons as $type => $text): ?>
@@ -16,3 +17,4 @@
     <?php endforeach ?>
 
 </div>
+<?php endif ?>


bennothommo:
@AIC-BV could you send through a screenshot of how it looks in preview, before and after your change?

Also, I'm not sure about using the URL to determine if it's in preview or not. I'm fairly certain you can use the form context for that.

LukeTowers:
The correct way to handle this is by passing the readOnly property to the `relationRender` method. 

Something like this in your partial that renders the relation controller:

```php
<?= $this->relationRender('myrelation', ['readOnly' => $field->disabled]); ?>
```

should do the trick.

AIC-BV:
As shown in the screenshot, review is updatable in preview context
image
Your code @LukeTowers disables a field but does not affect the toolbar in any way?

LukeTowers:
Are you using a custom toolbar partial or are you using toolbarButtons? https://github.com/wintercms/winter/blob/develop/modules/backend/behaviors/RelationController.php#L598-L600 should be preventing the buttons from showing up. If it's not, then it's possible that the relation controller config itself needs to be modified earlier in the request as soon as it's detected that it's in preview mode, but I think that would probably be up to the plugin implementing the relation controller, not the relation controller itself.

AIC-BV:
> Are you using a custom toolbar partial or are you using toolbarButtons? https://github.com/wintercms/winter/blob/develop/modules/backend/behaviors/RelationController.php#L598-L600 should be preventing the buttons from showing up. If it's not, then it's possible that the relation controller config itself needs to be modified earlier in the request as soon as it's detected that it's in preview mode, but I think that would probably be up to the plugin implementing the relation controller, not the relation controller itself.

No it's in config_relation.yaml `toolbarButtons: update|delete`
`$this->readOnly` dumps null, not true in preview context (also null in update context)

LukeTowers:
Try
```php
<?= $this->relationRender('items', ['readOnly' => $field->disabled || $field->readOnly]); ?>
```

AIC-BV:
> Try
> 
> ```
> <?= $this->relationRender('items', ['readOnly' => $field->disabled || $field->readOnly]); ?>
> ```

'Solved!'
I found two solutions thanks to you, **but these solutions also disable the toolbarbuttons in update context**.

1.
fields.yaml
```
review:
            tab: Review
            type: partial
            path: ~/path/to/_review.htm
            readOnly: true
```

_review.html
```
<?= $this->relationRender('review', ['readOnly' => $field->disabled || $field->readOnly]); ?>
```

2.
Alternatively, you can do this from the docs: (remove readOnly property on field)
https://wintercms.com/docs/backend/relations#relation-display

BUT

both solutions also disable the toolbar in update context. I am looking for a way to only disable these buttons in preview context.
I can easily create two fields, @update @preview, if you guys don't call this a bug, it's not a problem for me. It just seems wrong..? (double field declaration for something that shouldn't be updatable in preview mode?)

AIC-BV:
I implemented it like this because it seems like the only solution at the moment.

```
review@preview:
    tab: Review
    type: partial
    path: ~/path/to/_review_preview.htm

review:
    context: ['create', 'update']
    tab: Review
    type: partial
    path: ~/path/to/_review.htm
```
and in _review_preview.htm: https://wintercms.com/docs/backend/relations#relation-display

The PR offers an alternative which handles this automatically, but not in the best way.
If you want you can remove this PR! 🙂

LukeTowers:
@AIC-BV I think the solution we actually need is to make the previewMode variable accessible to the partial. Currently, `type: partial` fields have no access to the underlying Form widget, and the `disabled` property is only automatically set on `FormWidget` fields, not `partial` fields.

If you change the https://github.com/wintercms/winter/blob/develop/modules/backend/widgets/form/partials/_field_partial.htm file to pass `'formWidget' => $this` then you would be able to do the following:

```php
<?= $this->relationRender('myrelation', ['readOnly' => $formWidget->previewMode || $field->disabled || $field->readOnly]); ?>
```

LukeTowers:
@AIC-BV did you have any thoughts on my last comment?

AIC-BV:
> @AIC-BV I think the solution we actually need is to make the previewMode variable accessible to the partial. Currently, `type: partial` fields have no access to the underlying Form widget, and the `disabled` property is only automatically set on `FormWidget` fields, not `partial` fields.
> 
> If you change the https://github.com/wintercms/winter/blob/develop/modules/backend/widgets/form/partials/_field_partial.htm file to pass `'formWidget' => $this` then you would be able to do the following:
> 
> ```
> <?= $this->relationRender('myrelation', ['readOnly' => $formWidget->previewMode || $field->disabled || $field->readOnly]); ?>
> ```

I am currently still using [my solution](https://github.com/wintercms/winter/pull/516#issuecomment-1091405630), but yours sounds even better and more flexible for future users.

LukeTowers:
This should also be solved by using the relationmanager formwidget once #904 is merged.

PR #517, opened 2022-03-30
Adding some usefull rules to .htaccess

damsfx:
This is a proposal to add some rules often used to improve SEO: 

- forced addition or removal of www in the domain
- adding a rule to remove the trailing slash to avoid duplicate content
- adding a rule for the robots.txt file if it is not managed by a third party module

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -19,6 +19,29 @@
     # RewriteRule (.*) https://%{SERVER_NAME}/$1 [L,R=301]
 
     ##
+    ## Uncomment following lines to force remove www from url
+    ##
+    # RewriteCond %{HTTP_HOST} ^www\.(.*)$ [NC]
+    # RewriteRule ^ %{REQUEST_SCHEME}://%1%{REQUEST_URI} [R=301,L]
+
+    ##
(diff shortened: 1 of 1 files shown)


bennothommo:
@damsfx thanks for this.

We've had similar proposals to this in the past when we were still involved with October. The general gist of what we felt was that it was best to keep the `.htaccess` file lean because there's no telling if people are already doing this on a server level or through a plugin (for example, my [URL Normaliser](https://github.com/bennothommo/wn-url-normaliser-plugin) plugin handles all this as well). While you have it commented out already so there wouldn't be any conflict, it's still something that may or may not be needed.

I'd say this would be an excellent addition to the docs though - perhaps as a new section for "Optimizations", perhaps? What do you think?

damsfx:
@bennothommo  No trouble at all for me.

Having it in an "optimizations" section in the winter doc sounds like a good idea, I didn't want to impose anything, these are just rules that I use frequently.  
Besides, as you point out, I never use the redirection for HTTPS, it is managed at the server level.

LukeTowers:
Feel free to submit a PR to the docs @damsfx, would be much appreciated :)

PR #518, opened 2022-03-30
fix null coalescent operator expression

mjauvin:
- add parenthesis around expression, was breaking when compiling the js asset without this
- recompile all js assets


bennothommo (approved):
Looks fine to me, if it's all tested and works, then feel free to merge.

@mjauvin you're right, null coalesce is definitely a valid operator, so the problem is with the asset compiler. I don't think the same problem would occur with the new Mix compiler.

LukeTowers:
This might be worth reporting to [wikimedia](https://phabricator.wikimedia.org/tag/wikimedia-minify/) (I believe the JS minifier is now using https://github.com/wikimedia/minify).

LukeTowers:
@mjauvin are you able to report the issue to wikimedia?

mjauvin:
@LukeTowers I tried to open an issue, but couldn't find a way without creating a developper account.

Do you have such an account yourself?

LukeTowers:
No, you'd have to create one.

mjauvin:
Man, forget it. Ever heard of "La maison des fous" ? You get bounced all over the place and never end-up in the right place... this is MADENING.

Why can't they have github issue enabled and that's it?

mjauvin:
Issue opened at wikimedia:
https://phabricator.wikimedia.org/T305277

PR #522, opened 2022-04-02
Update MailSettings for new Laravel mail config structure.

mjauvin:
Requires https://github.com/wintercms/storm/pull/79

Removed the following drivers configs not native in Laravel:
- Mailgun
- Mailchimp Mandrill
- Postmark
- Sparkpost
- Amazon SES


mjauvin:
I've tested the following drivers:
- php mail
- sendmail
- smtp
- log file

modules/system/models/MailSetting.php
```diff
@@ -85,62 +83,52 @@ public function getSendModeOptions()
...
 
             case self::MODE_SENDMAIL:
-                $config->set('mail.sendmail', $settings->sendmail_path);
+                $config->set('mail.mailers.sendmail.path', $settings->sendmail_path);
                 break;
 
             case self::MODE_MAILGUN:
                 $config->set('services.mailgun.domain', $settings->mailgun_domain);
                 $config->set('services.mailgun.secret', $settings->mailgun_secret);
                 break;
```
LukeTowers: If I remember correctly, the mailgun driver also isn't supported by default in Laravel 9 and will require a driver plugin for as well.

mjauvin: I see this in Laravel MailManager:

```
    /**
     * Create an instance of the Symfony Mailgun Transport driver.
     *
     * @param  array  $config
     * @return \Symfony\Component\Mailer\Bridge\Mailgun\Transport\MailgunApiTransport
     */
    protected function createMailgunTransport(array $config)
    {   
        $factory = new MailgunTransportFactory();

        if (! isset($config['secret'])) {
            $config = $this->app['config']->get('services.mailgun', []);
        }

        return $factory->create(new Dsn(
            'mailgun+'.($config['scheme'] ?? 'https'),
            $config['endpoint'] ?? 'default',
            $config['secret'],
            $config['domain']
        ));
    }
```

mjauvin:
All done @LukeTowers 

Winter's drivers plugins are also fully functional (but not tested at transport level)

mjauvin:
@LukeTowers I'm pretty confident about those changes and the related https://github.com/wintercms/storm/pull/79

I've done a lot of tests back and forth between the different base drivers (as well as the new Mail driver plugins) and it all works as expected.

Only thing left is to confirm the Sparkpost & Mailgun drivers mapping, but it doesn't prevent this from being merged.
(the fact the other 3 tested driver works means all parts are "speaking to each other" as expected).

PR #524, opened 2022-04-05
Can't pass null to non-nullable arguments

Denoder:
This is in regards to: https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg

I'm getting this deprecation warning throughout all of winter with php 8.1 and it's happening in both winter and storm

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -926,7 +926,7 @@ class Lists extends WidgetBase
             $config['searchable'] = false;
         }
 
-        $columnType = $config['type'] ?? null;
+        $columnType = $config['type'] ?? '';
 
         $column = new ListColumn($name, $label);
         $column->displayAs($columnType, $config);
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -687,8 +687,8 @@ class Filter extends WidgetBase
      */
     protected function makeFilterScope($name, $config)
     {
-        $label = $config['label'] ?? null;
-        $scopeType = $config['type'] ?? null;
+        $label = $config['label'] ?? '';
+        $scopeType = $config['type'] ?? '';
 
         $scope = new FilterScope($name, $label);
         $scope->displayAs($scopeType, $config);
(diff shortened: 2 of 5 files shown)


modules/backend/widgets/Lists.php
```diff
@@ -926,7 +926,7 @@ protected function makeListColumn($name, $config)
             $config['searchable'] = false;
         }
 
-        $columnType = $config['type'] ?? null;
+        $columnType = $config['type'] ?? '';
```
LukeTowers, line 929: Should this default to text instead?

Denoder: the issue is i'm going around the entirety of winter and there are a lot of things that are giving the warnings with no context  as to what is "using" the internal args, just that the functions are being used and null is being passed to them.

LukeTowers: So is this PR still a work in progress then?

Denoder: yes, there's still a lot of areas that I have to go through, I also have to do a pr for winter storm as well.

Denoder: are you not getting these warnings on php 8.1?

LukeTowers: I'm not running any projects on 8.1 at the moment, wanted to focus more on the Laravel 9 specific issues rather than have to deal with both at the same time.

mjauvin:
@Teranode are you still getting those warnings? If so, where and how? I'm running 8 projects on latest wip/1.2 and have not seen those.

Denoder:
I went back to php8.0 cause the errors were getting pretty annoying, so Idk if it's still happening, I'll check tomorrow to see if it's still occurring

LukeTowers:
@Teranode if you could check to see if this is still a problem without your changes and then resubmit the PR if it is that would be great; thanks!

Denoder:
@LukeTowers yes it is still happening with php 8.1

image

Denoder:
There's also this one, I've not seen any indication of `str_starts_with` being use, but a lot of `starts_with` being used so it's complicated for me as I can't root it down.
image

PR #531, opened 2022-04-10
Fix error using undefined var

jaxwilko:
On error this var is not defined, which leads to:
image


jaxwilko:
Also, we should probably look at implementing some error state tests

LukeTowers:
@jaxwilko could you target develop with this since we will be releasing a 1.1.9 that already has some asset compilation tweaks?

jaxwilko:
I'll re open with the right history

PR #538, opened 2022-04-24
Add contextual information to event log

zaxbux:
In the logging docs, it says that contextual data can be included with the log entry:

> ## Logs
> #### Contextual information
> An array of contextual data may also be passed to the log methods. This contextual data will be formatted and displayed with the log message:
> ```php
> Log::info('User failed to login.', ['id' => $user->id]);
> ```

This PR adds `$event->context` to the `EventLog::add()` call so that the context data is stored in the `details` column in the `system_event_logs` table.

### DIFF
--- a/modules/system/tests/console/WinterUtilTest.php
+++ b/modules/system/tests/console/WinterUtilTest.php
@@ -6,14 +6,34 @@ use System\Tests\Bootstrap\TestCase;
 
 class WinterUtilTest extends TestCase
 {
+    protected string $defaultClient;
+    protected string $langClient;
+    protected string $langCountryClient;
+    protected array $createdDirs = [];
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->defaultClient = base_path('modules/system/lang/en/client.php');
+        $this->langClient = base_path('lang/en/system/client.php');
+        $this->langCountryClient = base_path('lang/en-gb/system/client.php');
+
+        // Backup original files
+        if (file_exists($this->defaultClient)) {
+            rename($this->defaultClient, $this->defaultClient . '.backup');
+        }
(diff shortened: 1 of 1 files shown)


modules/system/ServiceProvider.php
```diff
@@ -304,7 +304,7 @@ protected function registerLogging()
     {
         Event::listen(\Illuminate\Log\Events\MessageLogged::class, function ($event) {
             if (EventLog::useLogging()) {
-                EventLog::add($event->message, $event->level);
+                EventLog::add($event->message, $event->level, $event->context);
```
LukeTowers:
```suggestion
                EventLog::add($event->message, $event->level, ['context' => $event->context]);
```
Good idea @zaxbux! Let's add it under a `context` key though for easier scoping of the data. Keeping in mind as well that we plan on implementing support for the Ignition error page in the core and it provides a lot of contextual information.

LukeTowers:
@zaxbux any ideas for how we can display this contextual data in the backend event logs view?

zaxbux:
Maybe pretty-printing JSON or by using `Symfony\Component\VarDumper\Dumper\HtmlDumper` for example:

backend-log-context

That screenshot is of a somewhat crude implementation that required changes to JS and CSS, although there's probably a more elegant solution.

LukeTowers:
@zaxbux a simple `dump($context)` should be fine, it can always be improved upon later.

LukeTowers:
@zaxbux are you still interested in this?

zaxbux:
Admittedly, I forgot to commit my changes back in May.

LukeTowers:
@zaxbux so what does this PR look like now? Do you have a screenshot you can share?

PR #543, opened 2022-04-28
add spanish localization for 'auth.invalid_login'

mjauvin:


mjauvin:
Credit to Federico Schafer for the translation.

PR #548, opened 2022-05-04
Establish base Snowboard framework in Backend

### LINKED ISSUE #541: Snowboard.js isn't picking up on the CSRF token automatically in the backend
When making a request via Snowboard in the backend it will always fail with invalid token errors because it is not automatically including the CSRF token from the meta tag in the head.

bennothommo:
Fixes https://github.com/wintercms/winter/issues/541

This PR establishes a base Snowboard framework in the Backend. While we won't likely have any specific Snowboard widgets or functionality in the 1.1 branch, it will allow people to use Snowboard in the Backend should they wish.

It includes the following:

- New Asset Manager handling for backend assets and pre-loading JavaScript, CSS and image files.
- Proper handling of CSRF token in the backend when using Snowboard requests.
- All Snowboard extras are available for use in the Backend.

Some changes were needed to be made to Snowboard itself to facilitate these changes

- Implemented proper handling of assets returned in Request responses, including new events for asset management.
- Added some new methods to Snowboard itself - `on`, `off` and `ready`. These allow simple callbacks to Snowboard events for ad-hoc functionality. The `ready` method is a shorthand for `on('ready', ...)`, and is a nice shortcut in lieu of `$(document).ready`.


modules/backend/assets/ui/js/ajax/AssetLoader.js
```diff
@@ -0,0 +1,107 @@
...
+                resolve();
+            });
+            img.addEventListener('error', () => {
+                this.snowboard.globalEvent('backend.assetError', 'image', image, img);
+                reject(new Error(`Unable to load image file: "${image}"`));
+            });
+            img.src = image;
+        });
+    }
+}
```
LukeTowers: @bennothommo can we make this a more general Snowboard utility so that it can be used in the frontend as well? Also @jaxwilko please review and compare to your changes that you made for our internal project.

modules/backend/assets/ui/js/ajax/Handler.js
```diff
@@ -0,0 +1,59 @@
+export default class Handler extends Snowboard.Singleton {
```
LukeTowers, line 15: Can we add a docblock here?

modules/backend/assets/ui/js/ajax/AssetLoader.js
```diff
@@ -0,0 +1,107 @@
+export default class AssetLoader extends window.Snowboard.Singleton {
```
LukeTowers: Can we add documentation here?

modules/backend/assets/ui/js/ajax/AssetLoader.js
bennothommo: Sure thing :)

modules/backend/controllers/index/index.htm
```diff
@@ -1,21 +1,21 @@
...
-            $('#dashReportContainer').removeClass('loading')
-        })
-    })
+Snowboard.ready(() => {
+    Snowboard.request(null, 'onInitReportContainer', {
+        success: () => {
+            $('#dashReportContainer').removeClass('loading');
+        },
+    });
+});
```
LukeTowers, line 21: Is there a way that we can move this into the scripts block instead so that it doesn't force the script tag to be injected adhoc in the middle of the body element? This causes problems with wrapping the entire contents of the backend in a Vue instance; although that can be worked around it'd be nice to encourage scripts to be put in the same place together at the end of the body tag.

modules/backend/main.js
```diff
@@ -0,0 +1,6 @@
+/******/ (() => { // webpackBootstrap
+/******/ 	"use strict";
+/******/ 	
+/******/ 	
+/******/ })()
+;
```
LukeTowers: ?

modules/backend/package.json
```diff
@@ -0,0 +1,42 @@
...
+    ],
+    "license": "MIT",
+    "bugs": {
+        "url": "https://github.com/wintercms/winter/issues"
+    },
+    "homepage": "https://wintercms.com/",
+    "dependencies": {
+        "@popperjs/core": "^2.11.4",
+        "js-cookie": "^3.0.1",
+        "vue": "^3.2.31"
```
LukeTowers: Is Vue being loaded for the backend right now?

modules/backend/package.json
```diff
@@ -0,0 +1,42 @@
...
+        }
+    ],
+    "license": "MIT",
+    "bugs": {
+        "url": "https://github.com/wintercms/winter/issues"
+    },
+    "homepage": "https://wintercms.com/",
+    "dependencies": {
+        "@popperjs/core": "^2.11.4",
+        "js-cookie": "^3.0.1",
```
LukeTowers: Isn't this just a system requirement for snowboard's cookie utility?

modules/backend/package.json
```diff
@@ -0,0 +1,42 @@
...
+            "url": "https://wintercms.com"
+        }
+    ],
+    "license": "MIT",
+    "bugs": {
+        "url": "https://github.com/wintercms/winter/issues"
+    },
+    "homepage": "https://wintercms.com/",
+    "dependencies": {
+        "@popperjs/core": "^2.11.4",
```
LukeTowers: What is this used for?

package.json
```diff
@@ -0,0 +1,11 @@
...
+    "devDependencies": {
+        "laravel-mix": "^6.0.41"
+    },
+    "workspaces": {
+        "packages": [
+            "modules/backend",
+            "modules/system"
+        ]
+    }
+}
```
LukeTowers:
```suggestion
}

```

LukeTowers (changes requested):

modules/backend/main.js
bennothommo: I've had that happen a couple of times if I've forgotten a `-p` argument for `mix:compile` when trying to compile a single module... I think it's some weird NPM flex. I'll make it an issue at some point and will clear it out of here.

package.json
bennothommo: This file shouldn't be in core. I'll remove it.

modules/backend/package.json
bennothommo: It's not, it's not even included in the build since we're not using it. I'll remove it - I think I borrowed this config from the Inspector rewrite which is using Vue.

modules/backend/package.json
bennothommo: It was part of the Inspector rewrite. It's not used anywhere at the moment, so I'll remove it.

bennothommo:
I've included a Data Config plugin in this PR. It's not actually being used yet, but it's quite a common feature in our widgets and may find use on the CMS side as well.

modules/backend/controllers/index/index.htm
bennothommo: I can do it for this instance, but I believe there's quite a lot of inline scripts littered throughout the Backend. I found it easier to focus Vue on a certain section as opposed to making the entire Backend a Vue instance. I think if we wanted to go down that road, we'll have to significantly rewrite the Backend.

modules/backend/assets/ui/js/ajax/Handler.js
```diff
@@ -0,0 +1,107 @@
+/**
+ * Backend AJAX handler.
+ *
+ * This is a utility script that resolves some backwards-compatibility issues with the functionality
+ * that relies on the old framework, and ensures that Snowboard may work within the Backend
```
LukeTowers:
```suggestion
 * that relies on the old framework, and ensures that Snowboard works well within the Backend
```

modules/system/assets/js/snowboard/extras/DataConfig.js
```diff
@@ -0,0 +1,208 @@
+/**
+ * Data configuration provider.
+ *
+ * Provides a mechanism for passing configuration data through an element's data attributes. This
+ * is generally used for widgets or UI interactions to configure them.
+ *
+ * @copyright 2022 Winter.
+ * @author Ben Thomson <git@alfreido.com>
+ */
+export default class DataConfig extends Snowboard.PluginBase {
```
LukeTowers, line 10: If nothing is currently using it I'd prefer that we don't include it for now.

modules/system/assets/js/snowboard/extras/DataConfig.js
```diff
@@ -0,0 +1,208 @@
...
+
+        // Null value
+        if (stringValue === 'null') {
+            return null;
+        }
+
+        // Undefined value
+        if (stringValue === 'undefined') {
+            return undefined;
+        }
```
LukeTowers, line 183: Can we also support base64 encoded data?

LukeTowers (changes requested):
Can you also submit a PR to the docs with the new utilities / any other relevant changes?

modules/system/assets/js/snowboard/extras/DataConfig.js
bennothommo: Have a read of the docs when they're done, and let me know if you still feel that way ;)

modules/system/assets/js/snowboard/extras/DataConfig.js
bennothommo: What would be the intended purpose of that?

LukeTowers: Easier to include longer form data in the data attributes without having to worry about if a given value is going to escape the attribute quotes. May be needed by the inspector for more complex field types like richeditor.

bennothommo: Unless we add some form of prefix or indicator that it is a base64 string, there's no way to tell from the outset if a given string is a base64 string or if it is indeed a legitimate value (like, say, a hash).

bennothommo:
Docs are ready: https://github.com/wintercms/docs/pull/77

modules/system/assets/js/snowboard/extras/DataConfig.js
LukeTowers: How about we go with supporting the data:uri syntax for base64 https://en.wikipedia.org/wiki/Data_URI_scheme (down the road we can support other forms of Data URIs I suppose)?

bennothommo: @LukeTowers I've added support for base64 - it will look for a `base64:` prefix in the config value.

PR #447, opened 2022-02-12
Icon font update

WebVPF:
This pull request updates FontAwesome icons from version 4.7.0 to 6.0.0

 version                                                                                                     | Icons | Brands Icons
------------------------------------------------------------------------------|-------|------------------
[FontAwesome 4](https://fontawesome.com/v4/icons/)                          | 675   | 177
[FontAwesome 6 (Free)](https://fontawesome.com/search?m=free)       | 1748 | 459

## Icons

Original:

```css
.fa-snowflake::before {content:"\f2dc"}
```

For Winter CMS:

```css
.wn-icon-snowflake::before,
.oc-icon-snowflake::before,
.icon-snowflake::before {content:"\f2dc"}
```

## Brands icons

Original:

```css
.fa-github:before {content:"\f09b"}
```

For Winter CMS:

```css
.wn-icon-github:before,
.oc-icon-github:before,
.icon-github:before {content:"\f09b";font-family:'FontAwesome Brands'}
```

## Old Icons FontAwesome 4.7.0

All old icons that are not in the new library are supported.

```css
.wn-icon-snowflake-o:before,
.oc-icon-snowflake-o:before,
.icon-snowflake-o:before {content:"\f2dc";font-family:'FontAwesome 4'}
```

---

There is a change that changes styles. It was:
https://github.com/wintercms/winter/blob/cf5ae34f5adfa17f18d7f8f381a2fc73330b744a/modules/system/assets/ui/storm.css#L4117

It became:
```css
.icon-lg {font-size:1.25em;line-height:0.05em;vertical-align:-0.075em}
```

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/447"></a>

PR #445, opened 2022-02-11
[WIP] Icon font update

WebVPF:
This pull request updates FontAwesome icons from version 4.7.0 to 6.0.0

## Icons

Original:

```css
.fa-snowflake::before {content:"\f2dc"}
```

For Winter CMS:

```css
.wn-icon-snowflake::before,
.oc-icon-snowflake::before,
.icon-snowflake::before {content:"\f2dc"}
```

## Brands icons:

Original:

```css
.fa-github:before {content:"\f09b"}
```

For Winter CMS:

```css
.wn-icon-github:before,
.oc-icon-github:before,
.icon-github:before {content:"\f09b";font-family:'FontAwesome Brands'}
```

## Compatibility

There is no full support for old icons. Many old icons are supported, but some are not.

Example FA-4.7.0 `fa-snowflake-o` => FA-6.x `fa-snowflake`

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/445"></a>

PR #551, opened 2022-05-06
Upgrade Font Awesome 4 to 6

bennothommo:
Replaces https://github.com/wintercms/winter/pull/445 and https://github.com/wintercms/winter/pull/447.

This PR upgrades Font Awesome 4 to Font Awesome 6, bringing the number of available icons from 675 to 1748, as mentioned in the above PRs. It has been implemented using Font Awesome's LESS files, but keeps everything backwards compatible by shimming the old icon variables and ensuring the old classes point to the correct font weighting and icon.

This does increase the Backend assets from 4.84MB to 5.06MB (an increase of ~220kb) uncompressed - I haven't tested this using GZIP but would estimate it would equate to around 50-60kb, so shouldn't be too much of a hit.


bennothommo:
### Screenshots

#### Before

Screenshot 2022-05-06 at 13-54-25 Settings Winter CMS

#### After

Screenshot 2022-05-06 at 13-54-00 Settings Winter CMS

modules/system/assets/ui/font/LICENSE.txt
```diff
@@ -0,0 +1,165 @@
...
+learn about Font Awesome.
+
+--------------------------------------------------------------------------------
+
+# Brand Icons
+
+All brand icons are trademarks of their respective owners. The use of these
+trademarks does not indicate endorsement of the trademark holder by Font
+Awesome, nor vice versa. **Please do not use brand logos for any purpose except
+to represent the company, product, or service to which they refer.**
```
LukeTowers: Why do we have this file included twice?

modules/system/assets/css/styles.css
```diff
@@ -368,3176 +368,7744 @@ blockquote.pull-right .small:after{content:'\00A0 \2014'}
 blockquote:before,
 blockquote:after{content:""}
 address{margin-bottom:20px;font-style:normal;line-height:1.42857143}
+.close{float:right;font-size:21px;font-weight:bold;line-height:1;color:#000;text-shadow:0 1px 0 #fff;font-family:sans-serif;opacity:0.2;filter:alpha(opacity=20)}
+.close:hover,
+.close:focus{color:#000;text-decoration:none;cursor:pointer;opacity:0.5;filter:alpha(opacity=50)}
+button.close{padding:0;cursor:pointer;background:transparent;border:0;-webkit-appearance:none}
+.icon-0::before,
+.wn-icon-0::before,
+.oc-icon-0::before{content:"\30"}
```
LukeTowers: Can we break these icon content definitions into their own compiled file and just reference them as needed so they aren't adding 11k lines to several different asset files? Also, can you verify that the oc- BC prefix is only applied on the old icon classes?

LukeTowers (changes requested):

modules/system/assets/css/styles.css
bennothommo: I'm not sure I follow. Do you mean separating the different icon types? We haven't got a mechanism yet that detects when icons are in use, so we can't conditionally load the icon styles. I just followed the same compilation process that the original icons used.

bennothommo: As for the original oc- polyfills, it's only shimming the original LESS variables, but it is applying an oc prefix to all icons, including the new ones. I can limit this to just the original icons if needed.

LukeTowers: @bennothommo I mean that the styles.css and storm.css both had ~10k lines added / changed, would probably be better to load those styles in an icons.css that gets added via the AssetMaker trait methods whenever the its required. If it's not too much bother I'd prefer if the oc- prefixes were only added to the original icon names, that's a lot of duplicated code that is unnecessary and shouldn't be encouraged to be used.

bennothommo: I think it's reasonable to assume that the icons are used *everywhere* in the Backend. Plus, there would be people who are using icons in an ad-hoc way (myself included) who would be forced to add an `addCss` call to their plugins.

My thoughts are this: for 50kb-60kb extra, you're getting nearly 1100 extra icons. It wouldn't be a tall order for a lot of people.

modules/system/assets/ui/font/LICENSE.txt
bennothommo: Must've been leftover when I reorganised the new assets. Removed.

modules/system/assets/css/styles.css
LukeTowers: How about at least we provide it via a CSS import at the top of the file then? Then it's being cached separately and can be shared between the two compiled files.

LukeTowers:
@bennothommo are you also able to update the UI Icon docs in this PR? Should we like to the FontAwesome 6 site instead of hosting it all ourselves?

bennothommo:
@LukeTowers I was thinking that, but we do use a different class structure for the icons. They provide a JSON schema of all icons so it may not be too difficult to create a component that can generate the list in our docs.

LukeTowers:
16k icons is a **lot**. I'd be fine with just including instructions for how to change their class names to ours and then linking to their docs. They have a custom algolia integration in order to search all of the icons, I don't really want to go down that road on our end too.

modules/system/assets/css/styles.css
bennothommo: I've opted to compile the icons into a separate icon CSS file, which is loaded on all Backend pages. I've re-adjusted how the icons are compiled as well so that the base icon styles can be loaded, and allow us to be selective on icons to load, like the exception pages and maintenance or missing database views. This was previously loading all icons (through `modules/system/assets/css/style.css`).

LukeTowers (approved):

WebVPF:
@bennothommo there is no `content` property for brand icons in css.
Example: https://github.com/wintercms/winter/blob/wip/1.2/modules/system/assets/ui/icons.css#L6299-L6301
```
.icon-css3:before,
.wn-icon-css3:before,
.oc-icon-css3:before{font-family:'Font Awesome 6 Brands';font-weight:400}
```

bennothommo:
@WebVPF sorry, it appeared I hadn't included the brand icons in the compiled icon CSS.

They're included now. :)

PR #552, opened 2022-05-12
[WIP] 1.2 - AssetMaker respect app.asset_url

jaxwilko:
Simple fix that updates the AssetMaker trait to respect the `app.asset_url` config value.

### DIFF
--- a/modules/system/traits/AssetMaker.php
+++ b/modules/system/traits/AssetMaker.php
@@ -299,6 +299,10 @@ trait AssetMaker
             return $fileName;
         }
 
+        if ($assetUrl = config('app.asset_url')) {
+            $assetPath = $assetUrl . '/' . ltrim($assetPath, '/');
+        }
+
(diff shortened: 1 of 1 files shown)


modules/system/traits/AssetMaker.php
```diff
@@ -299,7 +299,7 @@ public function getAssetPath($fileName, $assetPath = null)
             return $fileName;
         }
 
-        return $assetPath . '/' . $fileName;
+        return Url::asset($assetPath . '/' . $fileName);
```
LukeTowers, line 302: @jaxwilko I think it also needs to be applied on line 299 right above this, since any asset starting with `/` will return early there instead of being passed through the Url::asset() method.

jaxwilko: Added in d635c581dd75bffa2f8ee2b6d3a1e3c14170a96b

PR #430, opened 2022-02-02
[WIP] Added support for --disk in winter:mirror command

jaxwilko:
Added support for mirroring to disks with `--disk` flag. This should add support for https://github.com/wintercms/winter/issues/427

<a href="https://gitpod.io/#https://github.com/wintercms/winter/pull/430"></a>

PR #559, opened 2022-05-17
[WIP] 1.2 winter:mirror Enhancements

jaxwilko:
This PR adds support for `--copy` & `--ignore "/pattern/"` to the `winter:mirror` command. These flags allow the mirror command to be used by Laravel Vapor to extract assets during the build stage.

The changes here supersede those made in:
https://github.com/wintercms/storm/pull/68
https://github.com/wintercms/winter/pull/430

Originally Vapor only supported 400 assets per project, meaning that we had to use the mirror command to mirror to a disk (remote s3 bucket), however with [this update](https://blog.laravel.com/vapor-unlimited-assets-uploads-10x-faster) this is no longer required and we therefore don't need the extra functionality, instead we can use copy the files so Vapor can pick them up and ignore development files or files not required for remote deployment.

### DIFF
--- a/modules/system/console/WinterMirror.php
+++ b/modules/system/console/WinterMirror.php
@@ -2,7 +2,6 @@
 
 use File;
 use Event;
-use JetBrains\PhpStorm\ArrayShape;
 use StdClass;
 use Winter\Storm\Console\Command;
 
--- a/modules/system/console/WinterMirror.php
+++ b/modules/system/console/WinterMirror.php
@@ -140,6 +140,11 @@ class WinterMirror extends Command
             $this->mirrorWildcard($wildcard);
         }
 
+        $index = $this->getDestinationPath() . '/index.php';
+        if ($this->option('copy') && File::isFile($index)) {
+            $this->updateIndexPaths($index);
+        }
+
         $this->output->writeln('<info>Mirror complete!</info>');
     }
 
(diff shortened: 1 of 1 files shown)


modules/system/console/WinterMirror.php
```diff
@@ -201,7 +197,49 @@ protected function mirror($src, $dest)
...
+            $this->info('Linked: ' . $dest);
+            return true;
+        }
+
+        if (is_dir($src) && !is_dir($dest)) {
+            mkdir($dest, 0755, true);
+        }
+
+        if (is_file($src)) {
+            !is_dir(dirname($dest)) && mkdir(dirname($dest), 0755, true);
```
LukeTowers: Should these permissions come from the `Winter\Storm\Filesystem\Filesystem->getFolderPermissions() / getFilePermissions()`? Since those come from `cms.defaultMask.file` and `cms.defaultMask.folder`?

LukeTowers:
See also https://github.com/octobercms/october/pull/4227 for some other changes that might be required for this PR to fully work for the purposes of having a working public folder that can be served from

modules/system/console/WinterMirror.php
jaxwilko: see [2faeed5](https://github.com/wintercms/winter/pull/559/commits/2faeed51b99aea5b1d47813a9b2dab20133a8d31)

jaxwilko:
> See also [octobercms/october#4227](https://github.com/octobercms/october/pull/4227) for some other changes that might be required for this PR to fully work for the purposes of having a working public folder that can be served from

This support was added in [0a2778c](https://github.com/wintercms/winter/pull/559/commits/0a2778ca85dce0e556e3e763ef6010d27cccb30b)

modules/system/console/WinterMirror.php
```diff
@@ -2,10 +2,9 @@
 
 use File;
 use Event;
+use JetBrains\PhpStorm\ArrayShape;
```
jaxwilko:
```suggestion
```

LukeTowers:
@jaxwilko was there anything left to do on this?

Issue #554, opened 2022-05-12
Add winter:util command to purge the resized images folder


LukeTowers:
See https://github.com/wintercms/winter/pull/178#issuecomment-841948643

LukeTowers:
Archiving, ping me to reopen if desired.

PR #574, opened 2022-06-13
Add winter:util purge resized cache command

leighboz:
Fix for issue #554 

The utilPurgeResizedCache() function is basically a copy of the utilPurgeUploads() function but without the check to skip valid files.

### DIFF
--- a/modules/system/console/WinterUtil.php
+++ b/modules/system/console/WinterUtil.php
@@ -292,6 +292,63 @@ class WinterUtil extends Command
         }
     }
 
+    protected function utilPurgeResizedCache()
+    {
+        if (!$this->confirmToProceed('This will PERMANENTLY DELETE all images in the resized directory.')) {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


modules/system/console/WinterUtil.php
```diff
@@ -292,6 +292,63 @@ protected function utilPurgeThumbs()
...
 
+    protected function utilPurgeResizedCache()
+    {
+        if (!$this->confirmToProceed('This will PERMANENTLY DELETE all images in the resized directory.')) {
+            return;
+        }
+
+        $totalCount = 0;
+        $uploadsPath = Config::get('filesystems.disks.local.root', storage_path('app'));
+        $uploadsPath .= '/resized';
```
LukeTowers: Are you able to get this to support remote disks by looking at the `cms.storage.resized` configuration: https://github.com/wintercms/winter/blob/develop/config/cms.php#L329-L333?

Issue #575, opened 2022-06-15
PluginManager doesn't find replaced plugins


cstorus:
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

Skripteria.SiteSearch, Winter.Blog, Winter.Pages

### Issue description

The Winter PluginManager->hasPlugin() method doesn't correctly identify replaced plugins. This appears to be caused by searching the replacementMap array using the normalized identifier, however the array is keyed using non-normalized identifiers.

```
    public function hasPlugin(PluginBase|string $plugin): bool
    {
        $normalized = $this->getNormalizedIdentifier($plugin);

        return isset($this->plugins[$normalized]) || isset($this->replacementMap[$normalized]);
    }
```
![image](https://user-images.githubusercontent.com/84441253/173840100-d08053f4-8336-4843-af37-9f318968c1d5.png)

### Steps to replicate

Create a replacement plugin with a replaced identifier
Search for the original plugin identifier using the PluginManager

### Workaround

Amend the hasPlugin function in PluginManager:476 to use the non-normalized identifier when looking up in the replacementMap.

mjauvin:
@jaxwilko looks like this is for you!

jaxwilko:
@cstorus @mjauvin hey guys, I've had a look at it and I'm not sure I see the issue. I added this test (I'll push it later with anything else needed for this issue), and everything passed.

![image](https://user-images.githubusercontent.com/31214002/173873551-db01195c-1306-4515-9a72-aa09d340b31d.png)

Basically the rational for this is:

- `hasPlugin(): bool` - Tells us if the plugin is on the system (when querying a replaced plugin, if the replacement is installed and the original is not then `hasPlugin` will return true).
- `isDisabled(): bool` - Tells us if the plugin is disabled (false = enabled).
- `findByIdentifier(): Plugin|null` - Fetches the plugin class, this will always return a replacement if queried for an original plugin if the second param is not true.

@cstorus could you explain how `hasPlugin` is affecting what you're trying to do?

mjauvin:
@jaxwilko I think your test above succeeds because `Winter.Original` is still installed.

@cstorus is right, the test to see if the plugin is in the replacement map does so with the NORMALIZED identifier of the plugin... but the replacementMap does NOT use normalized keys.

jaxwilko:
@mjauvin ahh I see, thanks for explaining. So basically the issue is the following:

[...]

PR #576, opened 2022-06-15
[FIX] Has plugin fix

jaxwilko:
This PR resolves the issue raised in https://github.com/wintercms/winter/issues/575.

Simply it ensures that `normalizeIdentifier()` returns the value passed if the key is not found in the `normalizedMap` array.

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -455,8 +455,14 @@ class PluginManager
             return $identifier;
         }
 
-        if (!$ignoreReplacements && is_string($identifier) && isset($this->replacementMap[$identifier])) {
-            $identifier = $this->replacementMap[$identifier];
+        // TODO: we should refactor the plugin manager to handle identifiers internally as lowers
+        $replacementMap = array_combine(
+            array_map('strtolower', array_keys($this->replacementMap)),
+            array_values($this->replacementMap)
+        );
+
+        if (!$ignoreReplacements && is_string($identifier) && isset($replacementMap[strtolower($identifier)])) {
+            $identifier = $replacementMap[strtolower($identifier)];
         }
 
         if (!isset($this->plugins[$identifier])) {
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -473,7 +473,10 @@ class PluginManager
     {
         $normalized = $this->getNormalizedIdentifier($plugin);
 
-        return isset($this->plugins[$normalized]) || isset($this->replacementMap[$normalized]);
+        return isset($this->plugins[$normalized])
+            || isset($this->replacementMap[$normalized])
+            // TODO: we should refactor the plugin manager to handle identifiers internally as lowers
+            || in_array(strtolower($normalized), array_map('strtolower', array_keys($this->replacementMap)));
     }
 
     /**
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -571,8 +571,7 @@ class PluginManager
      */
     public function normalizeIdentifier(string $code): string
     {
-        $code = strtolower($code);
-        return $this->normalizedMap[$code] ?? $code;
+        return $this->normalizedMap[strtolower($code)] ?? $code;
     }
 
     /**
--- a/tests/fixtures/plugins/winter/replacenotinstalled/Plugin.php
+++ b/tests/fixtures/plugins/winter/replacenotinstalled/Plugin.php
@@ -1,0 +1,18 @@
+<?php namespace Winter\ReplaceNotInstalled;
+
+use System\Classes\PluginBase;
+
+class Plugin extends PluginBase
+{
+    public function pluginDetails()
+    {
+        return [
+            'name' => 'Winter Sample Plugin',
+            'description' => 'Sample plugin used by unit tests.',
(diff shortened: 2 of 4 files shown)


jaxwilko:
Probably helps if I commit the new plugin that the tests are targeting

jaxwilko:
As a note, I've added some consts for plugin count in the `PluginManagerTest` class, this just makes it easier to maintain and reduces the amount of magic numbers in the test

mjauvin:
@jaxwilko are you sure this fixes the issue, though? I don't see the link with `replacementMap[]`

jaxwilko:
@mjauvin the replacement map correctly stores normalized identifiers `'Winter.Original' => 'Winter.Replacement'`

image

The normalized map also correctly stores lower keys to normalized identifiers `'lower.case' => 'Camel.Case'`

image

The issue is when you normalize an identifier via `normalizeIdentifier()` it should `strtolower()` the input, check the `normalizedMap` and return any value found, returning the original input if no key was found. However the returned value was the string that had been ran through `strtolower()`

mjauvin:
But this line will never return true for a replaced plugin:
https://github.com/wintercms/winter/blob/develop/modules/system/classes/PluginManager.php#L184-L189

jaxwilko:
Expanding on the above, when you call `hasPlugin()` it normalizes the identifier, then checks the replacement map, for the example of the test `Winter.NotInstalled` is not installed and therefore does not exist in the `normalizedMap` so returns the lower case version `winter.notinstalled`, which is not in either `$this->plugins` or `$this->replacementMap`. By adding the fix to return the passed value, `Winter.NotInstalled` is in `$this->replacementMap` and therefore `hasPlugin()` returns true

mjauvin:
Oh, I get it now... but you still need to use the plugin name with the exact capitalization, otherwise it won't work... this is not true for the replacement plugin, though

jaxwilko:
> But this line will never return true for a replaced plugin: https://github.com/wintercms/winter/blob/develop/modules/system/classes/PluginManager.php#L184-L189

The reason for this is we don't know the true normalized identifier for the plugin being replaced, so we just take whatever the replacement plugin suggests, if a developer accidentally provides:
image
Then they we will use that as the normalized identifier of the plugin being replaced as we have no way of checking for that plugins real code (as it may not be installed on the system).

It is down to the plugin developer to correctly implement the replacement, they should notice when testing that the plugin wont be flagged as replaced if they mistype.

mjauvin:
```
>>> $pm->hasPlugin('Rainlab.Blog');
=> false

>>> $pm->hasPlugin('RainLab.Blog');
=> true

>>> $pm->hasPlugin('rainlab.blog');
=> false
```
vs
```
>>> $pm->hasPlugin('winter.blog');
=> true

>>> $pm->hasPlugin('Winter.blog');
=> true

>>> $pm->hasPlugin('Winter.Blog');
=> true
```

mjauvin:
> > But this line will never return true for a replaced plugin: https://github.com/wintercms/winter/blob/develop/modules/system/classes/PluginManager.php#L184-L189
> 
> The reason for this is we don't know the true normalized identifier for the plugin being replaced, so we just take whatever the replacement plugin suggests, if a developer accidentally provides: image Then they we will use that as the normalized identifier of the plugin being replaced as we have no way of checking for that plugins real code (as it may not be installed on the system).
> 
> It is down to the plugin developer to correctly implement the replacement, they should notice when testing that the plugin wont be flagged as replaced if they mistype.

But why don't we put everything as lowercase internally, and always normalize the provided identifiers for any of the methods?

jaxwilko:
@mjauvin honestly that's what I'd prefer to do, we have to transform everything everywhere and it doesn't make much sense doing that. Would be really nice to just `strtolower()` everything and ignore casing.

Re: mixed casing `hasPlugin` we could look at something like this:

image

But I really don't like it

mjauvin:
What's wrong with:
```diff
diff --git a/modules/system/classes/PluginManager.php b/modules/system/classes/PluginManager.php
index f2ba9cdbb..73a88c651 100644
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -180,7 +180,7 @@ class PluginManager
         $replaces = $pluginObj->getReplaces();
         if ($replaces) {
             foreach ($replaces as $replace) {
-                $this->replacementMap[$replace] = $classId;
+                $this->replacementMap[$this->getNormalizedIdentifier($replace)] = $classId;
             }
         }  
```

jaxwilko:
I think that would work with the pervious version of `normalizeIdentifier()` which would return `strtolower($code)`. However if we fix `normalizeIdentifier()` it wont solve the `hasPlugin` when calling with mixed cases.

We could revert back to the old version of `normalizeIdentifier()`, but I'm not a massive fan of that side effect.

@LukeTowers any thoughts?

jaxwilko:
I've marked this as a draft because while it resolves issues reported in https://github.com/wintercms/winter/issues/575 it adds in a lot of complexity with how we handle things internally.

This PR is more of a hack than a real fix at this time.

A proper solution would be to convert the internals of `PluginManager` to use lower cased identifiers, simplifying the logic for resolving plugins internally.

LukeTowers:
Closing in favour of https://github.com/wintercms/winter/commit/fba6446d3f8a0f4f9920e605d283de02803df68b

Issue #571, opened 2022-06-09
ViewMaker::makePartial() regression


mjauvin:
### Winter CMS Build

Other (please specify below)

### PHP Version

8.0

### Database engine

Other (please specify below)

### Issue description

The commit that changed the default extension for backend partials from `.htm` to `.php` broke BC.

ref. 643a581ce93cc5f8efb3fc8a2420c8dc014c8787

### Steps to replicate

When calling the ViewMaker trait makePartial() method, the following worked in previous Winter versions:
```
$obj->makePartial('modules/backend/widgets/form/partials/field_dropdown', ...);
```

LukeTowers:
Fixed in https://github.com/wintercms/winter/commit/7a25524f737b8c4a612f917900e0a276d43bd447

PR #587, opened 2022-06-24
Fix site relative partial paths failing in 1.2

bennothommo:
Fixes https://github.com/wintercms/winter/issues/571

This PR changes the behaviour in the `ViewMaker` trait to more closely match the `ViewMaker` in 1.1, while still retaining the multiple extension support introduced in https://github.com/wintercms/winter/commit/643a581ce93cc5f8efb3fc8a2420c8dc014c8787#diff-47d4d0a41b9809a09c2b0c209309cd6b776645682610fa2c1906ed66d6ef1b4b.

If a path that is relative the base path is encountered in the `getViewPath` method, and with one of the allowed extensions matches a valid file, this will be accepted as a partial path. If not, then the ViewMaker will attempt to same within the view paths added at that point. Previously, the base relative paths would only be checked if the path was symbolized in some fashion.

This PR also introduces a fix to the `winter:test` command by ensuring that the current working directory is always the base path. This matches the behaviour of a web request which would presume the same, and thus ensures that the changes above work in the tests as well.

### DIFF
--- a/modules/system/console/WinterTest.php
+++ b/modules/system/console/WinterTest.php
@@ -143,7 +143,7 @@ class WinterTest extends Command
 
         $process = new Process(
             array_merge([$this->phpUnitExec, '--configuration=' . $config], $args),
-            dirname($config),
+            base_path(),
             null,
             null
         );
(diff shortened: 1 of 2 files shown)


LukeTowers:
@bennothommo you have added a submodule for the winter/test plugin, needs to be removed.

Issue #569, opened 2022-06-06
Console create:plugin - PascalCase namespace


WebVPF:
### Winter CMS Build

dev-develop

### PHP Version

8.0

### Database engine

MySQL/MariaDB

### Issue description

Winter CMS 1.2
Console command `create:plugin` incorrectly changes namespace. Camelcase names are changed to capitalized names.
Example:
```bash
php artisan create:plugin WebVPF.MyName
```

Plugin.php file generation result:
```php
<?php namespace Webvpf\Myname;
...
```

Must be:
```php
<?php namespace WebVPF\MyName;
```

### Steps to replicate

Create a plugin

arvislacis:
Some comments for clarification:

1) Issue name and description should contain **PascalCase** not **camelCase**;
2) Issue is probably more relevant in `wintercms/storm` repository but that's just side node;
3) If we look at https://wintercms.com/docs/help/developer/guide#vendor-naming then it states - "The vendor or author code in a namespace must begin with an uppercase character and should not contain underscores or dashes." - nothing much about Pascal case here.
4) Pascal Case is mentioned under https://wintercms.com/docs/help/developer/guide#class-naming
5) If we look into WinterCMS own plugins then it looks like mostly PascalCase is used in the namespace names, for example:
- https://github.com/wintercms/wn-pwnedpasswords-plugin/blob/main/Plugin.php
- https://github.com/wintercms/wn-mailchimp-plugin/blob/main/Plugin.php
- https://github.com/wintercms/wn-drivermandrill-plugin/blob/main/Plugin.php
- etc.

So, yes, could be an issue.

LukeTowers:
Just need to change a single line here: https://github.com/wintercms/winter/blob/wip%2F1.2/modules/system/console/BaseScaffoldCommand.php#L47. PRs welcome.

RomainMazB:
@LukeTowers Actually, to be synced with the scaffold methods' code style, we'll need another modification here to add the `ucfirst` case: https://github.com/wintercms/storm/blob/develop/src/Scaffold/GeneratorCommand.php#L144-L146

LukeTowers:
@RomainMazB I'm not sure I understand, could you expand on that?

RomainMazB:
Sure @LukeTowers.

To convert "WebVPF" to "WebVPF" and ensuring the first letter is uppercase to keep the PascalCase style, we need to use the php native `ucfirst` function, or the `Str::ucfirst` laravel helper.

[...]

PR #588, opened 2022-06-29
[Fix] 1.2 - PluginManager Normalization

jaxwilko:
Fix for https://github.com/wintercms/winter/issues/569

This PR updates the `PluginManager` to ensure that when calling `getNormalizedIdentifier` or `normalizeIdentifier` the input is returned if the not found in the normalize map. 

Also as these methods were performing similar functions, and internally calling each other, I've updated `normalizeIdentifier` to call `getNormalizedIdentifier` and act as an alias.

I've added tests to ensure we do not run into regressions in the future.

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -571,11 +571,12 @@ class PluginManager
      * Normalizes the provided plugin identifier (author.plugin) and resolves
      * it case-insensitively to the normalized identifier (Author.Plugin)
      * Returns the provided identifier if a match isn't found
+     *
+     * This is an alias for `getNormalizedIdentifier()`
      */
     public function normalizeIdentifier(string $code): string
     {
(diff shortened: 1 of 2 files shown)


RomainMazB:
LGTM 👍🏻

modules/system/classes/PluginManager.php
```diff
@@ -571,11 +571,12 @@ public function getNamespace(PluginBase|string $plugin): string
      * Normalizes the provided plugin identifier (author.plugin) and resolves
      * it case-insensitively to the normalized identifier (Author.Plugin)
      * Returns the provided identifier if a match isn't found
+     *
+     * This is an alias for `getNormalizedIdentifier()`
```
bennothommo: This isn't really an alias as the methods have differing behaviours :)

```suggestion
```

Issue #503, opened 2022-03-21
Flysystem disks visibility


mjauvin:
### Winter CMS Build

Other (please specify below)

### PHP Version

8.0

### Database engine

MySQL/MariaDB

### Issue description

`Laravel\Illuminate\Filesystem\FilesystemManager` now use `Visibility::PRIVATE` by default now when instanciating the LocalAdapter class.

ref. https://github.com/laravel/framework/blob/9.x/src/Illuminate/Filesystem/FilesystemManager.php#L169-L174

So we need to pass `visibility => 'public'` to the Disk config if we want the files/folder created under that disk to be publicly accessible.

The Image Resizer seems to be using the "local" disk in some parts of the code(folders/files creation) and the "system" disk in other parts (for mapping assets resources).

### Steps to replicate

- add this markup to a test page: {{ 'assets/images/my-test-image.jpg' | theme | resize(600,400) }}
- Remove folder `storage/app/resized`
- Open the page URL
- look at the permissions of the `storage/app/resized` folder (and subfolders)

### Workaround

Add `'visibility' => 'public'` to the local disk definition.

LukeTowers:
Sounds like instead we should set the permissions property of the system disk config to `cms.defaultMask.folder`

mjauvin:
You mean like this, but for the system disk:
```
    protected function registerNativeFilesystem()
    {   
        $this->app->singleton('files', function () {
            $config = $this->app['config'];
            $files = new Filesystem;
            $files->filePermissions = $config->get('cms.defaultMask.file', null);
            $files->folderPermissions = $config->get('cms.defaultMask.folder', null);
            $files->pathSymbols = [
                '~' => base_path(),
                '$' => base_path() . $config->get('cms.pluginsDir', '/plugins'),
                '#' => base_path() . $config->get('cms.themesDir', '/themes'),
            ];
            return $files;
        });
    }
```

mjauvin:
Adding this to `Winter\Storm\Filesystem\FilesystemManager` works:
```
    protected function resolve($name, $config = null)
    {   
        if (is_null($config)) {
            $config = $this->getConfig($name);
        }
        if ($name === 'local') {
            $config['visibility'] = 'public';
        }

        return parent::resolve($name, $config);
    }
```

The problem with the `permissions` property of the system disk is that it's more complex:
ref. https://laravel.com/docs/9.x/filesystem#local-files-and-visibility

But we could also just pass the `visibility` parameter to `Storage::put()`:
[...]

PR #592, opened 2022-07-03
Fix image resizer folder/file permissions

mjauvin:
Fixes https://github.com/wintercms/winter/issues/503

Least favorite solution, only fixes the image resizer.

### DIFF
--- a/modules/system/classes/ImageResizer.php
+++ b/modules/system/classes/ImageResizer.php
@@ -352,7 +352,7 @@ class ImageResizer
             Event::fire('system.resizer.afterResize', [$this, $tempPath]);
 
             // Store the resized image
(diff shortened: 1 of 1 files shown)


LukeTowers:
@jaxwilko are there any concerns about adding these attributes when the resized disk is set to S3 or any other remote storage?

jaxwilko:
Will this generate public thumbnails for protected items?

Apart from that i don't think so but i will do a proper test with s3 tonight :)

mjauvin:
@LukeTowers what about applying this change only `if $disk->getDriver() === 'local'` ?

LukeTowers:
Closing in favour of https://github.com/wintercms/storm/pull/92

PR #602, opened 2022-07-13
Update EditorConfig link

WebVPF:
This is a useful extension that should be recommended to users who do not have it installed.
https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,4 @@
-# EditorConfig is awesome: http://EditorConfig.org
+# EditorConfig is awesome: https://EditorConfig.org
 
 # top-most EditorConfig file
 root = true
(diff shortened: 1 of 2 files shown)


bennothommo:
I think my only hesitation on including this is that I don't want to enforce a coding style on people writing their own projects. We use `.editorconfig` mainly to enforce standards on development of the core itself. The EditorConfig extension, I've found, can be a pain if you need to make a change that doesn't follow the rules - for example, some test cases needed a file that had no final newline, but EditorConfig kept adding it in, so in the end, I had to disable the extension.

@wintercms/maintainers any thoughts on this?

LukeTowers:
What's the UX for this? VSCode just asks if you want to install recommended extensions?

bennothommo:
@LukeTowers that's correct.

WebVPF:
Ben is right. The recommendation has been removed.

PR #609, opened 2022-07-15
Config - MySQL - default username

WebVPF:
For MySQL, the default value for `username` is `root`.
This makes it easy to install on localhost.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -55,7 +55,7 @@ return [
             'strict' => true,
             'unix_socket' => env('DB_SOCKET', ''),
             'url' => env('DATABASE_URL'),
-            'username' => env('DB_USERNAME', 'winter'),
+            'username' => env('DB_USERNAME', 'root'),
         ],
         'pgsql' => [
             'charset' => 'utf8',


arvislacis:
I think it's not good practice to use `root` here..., and even if it would be then probably same thing could be applied also to other database configs.

WebVPF:
@arvislacis I not use others. I fixed it as it should be by default in MySQL.
I just did a fresh install via composer. Default should be `root`. Why complicate?

arvislacis:
@WebVPF IMHO for security reasons... Ok, I understand that you are using local development environment and so on but I am sure that if we leave that value to default root then there is possibility that some users will deploy that value also to production servers with MySQL server main root password..., so if someone compromises that Winter CMS installation then all DB server would be compromised too.

LukeTowers:
Laravel uses "forge" as their default, and I personally would prefer "homestead", but until we have our own managed hosting service I don't see a good reason to change this right now.

What works for you as your default development environment doesn't necessarily mean that everyone else is using the same. I'll close this for now, but I'm still open to hearing other thoughts & opinions on the issue.

PR #616, opened 2022-07-18
Fix typo in MediaManager.php

cstorus:
Fix a typo in reference to MediaLibrary, fixing an error when getting vector image thumbnails in the media manager.


LukeTowers:
Thanks @cstorus!

Issue #296, opened 2021-09-01
Revamp authentication events


LukeTowers:
This is a proposal to revamp the authentication events system within Winter CMS. Feedback and suggestions very much welcome. Related: https://github.com/wintercms/wn-user-plugin/pull/11 & https://github.com/wintercms/storm/pull/48

The current events are as follows:
- `backend.user.login` -> Global event for backend users logging in
- `model.auth.afterImpersonate` -> Local event triggered after impersonation is stopped
- `model.auth.beforeImpersonate` -> Local event triggered before impersonation is started
- `canBeImpersonated()` -> Model method used if impersonation events don't return anything
- `winter.user.activate` -> Global event triggered after Winter.User user's activate for the first time
- `winter.user.reactivate` -> Global event triggered after Winter.User user's activate after being deactivated 
- `winter.user.login` -> Global event triggered after Winter.User user's login
- `winter.user.deactivate` -> Global event triggered after Winter.User user's deactivate
- `winter.user.activate` -> Global event triggered after Winter.User user's activate
- `winter.user.getNotificationVars` -> Not technically part of the auth system, used for extending the variables made available to the Winter.Notify notifications
- `winter.user.logout` -> Global event triggered after Winter.User user's logout
- `winter.user.beforeAuthenticate` -> Global event triggered immediately before Winter.User user's credentials are checked
- `winter.user.beforeRegister` -> Global event triggered before Winter.User user's registration is processed
- `winter.user.register` -> Global event triggered after Winter.User user's registration is processed

The revised system would remove any direct event logic from the Auth system and instead move it to model event methods in the form of the following:

Model events:
- `beforeRegister()`
- `afterRegister()`
- `beforeLogin()` (already exists)
- `afterLogin()` (already exists)
- `beforeLogout()`
- `afterLogout()`
- `beforeActivate($activationCode)`
- `afterActivate($activationCode)`
- `beforeResetPassword($resetCode, $newPassword)`
- `afterResetPassword()`
- `beforeImpersonate($impersonator)`
- `afterImpersonate($impersonator)`
- `beforeEndImpersonation($impersonator)`
- `afterEndImpersonation($impersonator)`
- `beforeSuspend()`
- `afterSuspend()`
- `beforeUnsuspend()`
- `afterUnsuspend()`
- `beforeBan()`
- `afterBan()`
- `beforeUnban()`
- `afterUnban()`

The Backend module & Winter.User plugin would then use those methods to fire the externally accessible events:
[...]

PR #617, opened 2022-07-18
Allow redirecting to a custom URL after sign-out

lex0r:
Backend users can be redirected to an pre-defined URL instead if the hardcoded one. Can be useful with SSO mechanisms when logging out of a session in a Winter-based app also requires an SSO sign-out action which is typically as simple as accessing a URL

### DIFF
--- a/modules/backend/formwidgets/NestedForm.php
+++ b/modules/backend/formwidgets/NestedForm.php
@@ -18,7 +18,7 @@ class NestedForm extends FormWidgetBase
     protected $defaultAlias = 'nestedform';
 
     /**
-     * @var [] Form configuration
+     * @var array Form configuration
      */
     public $form;
 


config/cms.php
```diff
@@ -93,6 +93,17 @@
...
+    /*
+    |--------------------------------------------------------------------------
+    | Back-end post sign-out redirect URL.
+    |--------------------------------------------------------------------------
+    |
+    | This acts as the default URL (relative or absolute) where a back-end user
+    | will be redirected to after she or he signs out.
+    */
+
+    'postSignoutRedirectUrl' => 'backend',
```
LukeTowers: This would be better located in the auth.php under the `auth.backend.redirectAfterLogout` config key; otherwise looks good!

LukeTowers: And actually, it might be better to implement the auth event system changes proposed in https://github.com/wintercms/winter/issues/296 instead and then rather than a config value we would just provide an event that you could listen to in order to do any sort of custom processing you desire, including redirecting.

config/cms.php
```diff
@@ -93,6 +93,17 @@
 
     'backendSkin' => \Backend\Skins\Standard::class,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Back-end post sign-out redirect URL.
+    |--------------------------------------------------------------------------
+    |
+    | This acts as the default URL (relative or absolute) where a back-end user
+    | will be redirected to after she or he signs out.
```
bennothommo:
```suggestion
    | will be redirected to after they sign out.
```

config/cms.php
lex0r: Event based approach is no doubt a better option! The only advantage of the PR is that it solves the issue here and now.

LukeTowers: In the interest of meeting both our goals (yours a solution now, mine a solution that I won't have to revert and replace later), how about we just implement the `backend.user.logout` event for now here and make it so that if a response is returned then it returns the response otherwise it does the default redirect action?

PR #621, opened 2022-07-20
Fix plugin dependencies not being sorted

### LINKED ISSUE #620: [v1.2] Plugin dependencies are not sorted in the correct order
### Winter CMS Build

1.2

### PHP Version

8.0

### Database engine

No database

### Plugins installed

Winter.Location, Winter.Pages, Winter.TailwindUI

### Issue description

I have a plugin which depends on other plugins in order to run, however it looks like when I run the database migrations for this plugin it is being run before it's dependencies and so the migrations are unable to successfully run since it depends on the other plugin's tables.

Here is my Plugin's $require definition:
```php
public $require = [
        'Bdx.Toolbox',
        'Winter.Location',
    ];
```
But when I run the following command to see what order plugins are being loaded in, this is what I get:
Command to get list of plugins:
```php
dd(array_keys(PluginManager::instance()->getPlugins()));
```
Result:
```bash
array:5 [
  0 => "Bdx.Locations"
  1 => "Bdx.Toolbox"
  2 => "Winter.Location"
  3 => "Winter.Pages"
  4 => "Winter.TailwindUI"
]
```

This is the correct result:
```bash
array:5 [
  0 => "Bdx.Links"
  1 => "Bdx.Toolbox"
  2 => "Winter.Demo"
  3 => "Winter.Location"
  4 => "Bdx.Locations"
]
```

### Steps to replicate

Add dependencies to Plugin.php file.

```php
public $require = [
    'Winter.Location',
];
```

And run `php artisan winter:up`

### Workaround

_No response_

ericp-mrel:
Fixes #620

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -901,8 +901,6 @@ class PluginManager
         }
 
         return array_map(function ($require) {
-            $require = $this->getNormalizedIdentifier($require, true);
-
             return $this->replacementMap[$require] ?? $require;
         }, is_array($plugin->require) ? $plugin->require : [$plugin->require]);
     }
@@ -984,6 +982,8 @@ class PluginManager
                 $depends = $this->getDependencies($plugin);
 
                 $depends = array_map(function ($depend) {
+                    $depend = $this->getNormalizedIdentifier($depend, true);
+
                     if (isset($this->replacementMap[$depend])) {
                         return $this->replacementMap[$depend];
                     }
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -901,6 +901,8 @@ class PluginManager
         }
 
         return array_map(function ($require) {
+            $require = $this->getNormalizedIdentifier($require, true);
+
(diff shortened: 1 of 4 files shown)


ericp-mrel:
I can work on adding some tests to make sure dependencies are sorted in the correct order.

LukeTowers:
That would be great @ericp-mrel, thanks!

ericp-mrel:
@LukeTowers I've added a test to check the sort order of dependencies and make sure they come first before the dependent plugin. 

Let me know if you think of a way to further improve the test.

LukeTowers (approved):

LukeTowers:
Thanks @ericp-mrel!

mjauvin:
@ericp-mrel could you tell me what problem the plugin loading issue was causing for your plugin ?

ericp-mrel:
@mjauvin, the issue I was having was I had two plugins and one depended on the other. For example I had Plugins A and B, Plugin A depended on Plugin B, but because the `PluginLoader` class was sorting them only in alphanumerical order (or something like that) I couldn't not successfully run the migrations for Plugin A, because the migrations for Plugin B had not run yet.

PR #623, opened 2022-07-22
Update german translation

simonmannsfeld:
I updated all german translations and tripplechecked every single translation for mistakes, in detail:

- Added missing keys
- Added missing trailing commas
- Changed string qualifiers from `"` to `'`
- Translated missing keys
- Translated keys that were still in english
- Typo correction (missing characters, two spaces in row, grammar)
- Changed some words that could be misleading
- Corrected upper- / lower-case

It will help german users to navigate the backend with ease.


modules/backend/lang/de/lang.php
```diff
@@ -516,44 +547,46 @@
...
+            'iso_8859_1' => 'ISO-8859-1 (Latin-1, Westeuropäisch)',
+            'iso_8859_2' => 'ISO-8859-2 (Latin-2, Mitteleuropäisch)',
+            'iso_8859_3' => 'ISO-8859-3 (Latin-3, Südeuropäisch)',
+            'iso_8859_4' => 'ISO-8859-4 (Latin-4, Nordeuropäisch)',
+            'iso_8859_5' => 'ISO-8859-5 (Latin, Kyrillisch)',
+            'iso_8859_6' => 'ISO-8859-6 (Latin, Arabisch)',
+            'iso_8859_7' => 'ISO-8859-7 (Latin, Griechisch)',
+            'iso_8859_8' => 'ISO-8859-8 (Latin, Hebräisch)',
+            'iso_8859_9' => 'ISO-8859-9 (Latin-5, Türkisch)',
+            'iso_8859_10' => 'ISO-8859-10 (Latin-6, Nordisch)',
```
LukeTowers, line 572: Are these typically translated in other systems? Our general approach is to not translate them here.

simonmannsfeld: @LukeTowers it's hard to find a CMS or e-commerce software with encoding settings.

I took a look to wikipedia and common windows tools. Indeed in german language it seems to be common to translate the encoding partial standard. You can see in the collected screenshots below.

translated

Should I change it back anyway to English? I guess to leave it in English should not be a problem for users. I guess that a user who knows what encoding is, should even know this few words in English. Tell me how you like it.

LukeTowers: In that case I'm fine with leaving these translated to German, the precedent seems fine to me.

PR #631, opened 2022-08-02
Fix -c option in create:model command

RomainMazB:
Not sure how and when this issue was introduced

image


arvislacis:
@RomainMazB From here: https://github.com/wintercms/winter/commit/574c7d4ef377da61f7bdc2980937770a0bcd332e

LukeTowers:
Thanks @RomainMazB!

Issue #629, opened 2022-07-28
Error when remaning folder in media manager


AllARTSoftworks:
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

Winter.Demo, Winter.Translate, Marcomessa.Vite, Utopigs.Seo, Winter.Pages, Excodus.TranslateExtended

### Issue description

When trying to rename a folder in media manager a popup with this text shows up:

_"Backend\Widgets\MediaManager::onLoadRenamePopup(): Return value must be of type array, string returned" on line 367 of [path]/modules/backend/Widgets/MediaManager.php_

![error](https://user-images.githubusercontent.com/14626691/181457582-ddc1e739-69c9-48f3-8f45-87bf0bc3b91d.jpg)

### Steps to replicate

Go to Media Manager, hover on folder, click on the right icon to open popup

arvislacis:
Because https://github.com/wintercms/winter/blob/3be17f520b84bc7b22f3ea1435b030483fcaf3a0/modules/backend/widgets/MediaManager.php#L355 expects `array` as return type of function but the https://github.com/wintercms/winter/blob/3be17f520b84bc7b22f3ea1435b030483fcaf3a0/modules/backend/widgets/MediaManager.php#L367 looks like return only string of partial content.

So either return type of function should be changed (most likely) to string or mixed, or returned data structure.

LukeTowers:
@arvislacis you are correct, the return type should be changed to string. Would you like to submit a PR to fix that for us?

arvislacis:
@LukeTowers Done in https://github.com/wintercms/winter/pull/632

PR #632, opened 2022-08-02
Fix return types for MediaManager widget functions

arvislacis:


arvislacis:
Fixes https://github.com/wintercms/winter/issues/629

PR #633, opened 2022-08-02
Fix status widget output on clean installation

arvislacis:
This fixes following error/incorrect output of fresh Winter CMS 1.2 installation:

Later this could be rewritten to use system boot date from `Parameter` model; if no parameter value is set then the parameter value should be stored from date of the first installed plugin (similar like current logics) or if it's fresh installation then set from current date.

### DIFF
--- a/modules/system/reportwidgets/Status.php
+++ b/modules/system/reportwidgets/Status.php
@@ -70,12 +70,7 @@ class Status extends ReportWidgetBase
         $this->vars['requestLogMsg'] = LogSetting::get('log_requests', false) ? false : true;
 
         // TODO: Store system boot date in `Parameter`
-        $this->vars['appBirthday'] = null;
-        $appBirthday = PluginVersion::orderBy('created_at')->first();
-
-        if ($appBirthday) {
-            $this->vars['appBirthday'] = $appBirthday->created_at;
-        }
+        $this->vars['appBirthday'] = PluginVersion::orderBy('created_at')->first()?->created_at;
     }
 
     public function onLoadWarningsForm()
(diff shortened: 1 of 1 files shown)


modules/system/reportwidgets/Status.php
```diff
@@ -70,7 +70,12 @@ protected function loadData()
         $this->vars['requestLogMsg'] = LogSetting::get('log_requests', false) ? false : true;
 
         // TODO: Store system boot date in `Parameter`
-        $this->vars['appBirthday'] = PluginVersion::orderBy('created_at')->first()->created_at;
+        $this->vars['appBirthday'] = null;
+        $appBirthday = PluginVersion::orderBy('created_at')->first();
+
+        if ($appBirthday) {
+            $this->vars['appBirthday'] = $appBirthday->created_at;
+        }
```
LukeTowers:
```suggestion
        $this->vars['appBirthday'] = PluginVersion::orderBy('created_at')->first()?->created_at;
```

LukeTowers:
@arvislacis give my commit a try and let me know if that works for you

modules/system/reportwidgets/Status.php
bennothommo: @LukeTowers I thought you weren't a fan of the nullable arrows? 😛

arvislacis:
@LukeTowers Your commit works fine too so we can use it instead of mine.

I personally also don't like the null safe operator/arrow (`()?->`) and, as far as I can see, then it's also not used anywhere else in Winter CMS code base but I guess if we are now on PHP 8+ then probably that's way to go.

LukeTowers:
I'm not the biggest fan of it either, but it is supported by PHP 8 which is our minimum supported version now and it is also the cleanest solution to the issue so 🤷

arvislacis:
@LukeTowers Yeah, although some further improvements could be done to properly store and handle app birthday date because currently it shows date of the first installed plugin (if any) not the true installation date. I may look into making PR for this later on although that's just a tiny and not so important problem.

PR #634, opened 2022-08-04
Fix generation of aliased-only css assets

RomainMazB:
When dealing with aliased-only css assets like so:
```twig
<link rel="stylesheet" href="{{ ['@framework.extras.css'] | theme }}">
```
Because aliased assets are inserted in both `$combineJs` and `$combineJs`, the `if (count($combineCss) > count($combineJs))` condition results to false.
This results in a failure because then the css aliased assets is handled as a js one and forcefully is not found:
image

To handle this use case where we can't infer the asset type programmatically,  
I suggest to give to the programmer the responsibility to not merge css files with js ones.  
To do this I just merged all the aliases together and look for matching aliases in this big merged array.


RomainMazB:
I just noticed that there is a _conflict_ between two similar aliases existing in both css and js alias:
https://github.com/wintercms/winter/blob/929e8f2dd335dd6524f2cdbed85e04785e7a8c5f/modules/system/classes/CombineAssets.php#L160
https://github.com/wintercms/winter/blob/929e8f2dd335dd6524f2cdbed85e04785e7a8c5f/modules/system/classes/CombineAssets.php#L162

Which results in the opposite bug with this commit: when importing only js aliased files like this:
```twig
<script src="{{ [
            '@jquery',
            '@framework',
            '@framework.extras'
        ] | theme }}"></script>
```
The framework.extras shortcut (without the `.js` or `.css` extension) results in the `.css` file due to the fact that it's registered after.
It could easily be solved from a programmer view by just using the alias with the desired extension.

LukeTowers:
I don't really understand what this PR is seeking to solve

RomainMazB:
> I don't really understand what this PR is seeking to solve

When the assets combiner attempts to combine a css aliased-assets without any other external css file: with the actual logic, the combiner just do the wrong guess that the assets is a js file, and try to load it from the js aliases.

You can try yourself with a single line of code inside a layout.
This fails 🔴
```twig
<link rel="stylesheet" href="{{ ['@framework.extras.css'] | theme }}">
```
This works 🟢 
```twig
<link rel="stylesheet" href="{{ ['@framework.extras.css', 'another-user-css-file.css'] | theme }}">
```
This is due to this logic, when an alias is used, it's added to both css and js assets stack:
https://github.com/wintercms/winter/blob/09aad658b5f03effd734dbfede84a98ecb6d7d73/modules/system/classes/CombineAssets.php#L316-L323
Because of that, this `if` statement is false and results to a "js-guessed" wrong extension:
https://github.com/wintercms/winter/blob/09aad658b5f03effd734dbfede84a98ecb6d7d73/modules/system/classes/CombineAssets.php#L338-L348

This PR reinforces the type-guessing logic to make sure css aliases are correctly handled.

LukeTowers:
Are you able to add a test case for this change?

RomainMazB:
I can give it a try, sure.

RomainMazB:
Done :)

modules/system/classes/CombineAssets.php
```diff
@@ -338,19 +338,32 @@ protected function prepareAssets(array $assets)
...
             $assets = $combineCss;
         }
-        else {
+        elseif ($cssCount < $jsCount) {
             $extension = 'js';
             $assets = $combineJs;
         }
+        else {
+            $extension = null;
+            $assets = array_unique([...$combineCss, ...$combineJs]);
```
LukeTowers: What case does this cover?

modules/system/tests/classes/CombineAssetsTest.php
```diff
@@ -104,4 +104,28 @@ public function testResetCache()
...
+        $this->assertNotNull($url);
+        $this->assertRegExp('/\w+[-]\d+/i', $url); // Must contain hash-number
+
+        $url = $combiner->combine(
+            ['@css1', '@css2'],
+            base_path() . '/modules/system/tests/fixtures/themes/test'
+        );
+        $this->assertNotNull($url);
+        $this->assertRegExp('/\w+[-]\d+/i', $url); // Must contain hash-number
+    }
```
LukeTowers, line 221: What happens if both CSS & JS assets are present? What about if an alias that is registered for both is used? Can you add the following test cases:

- 0 aliases - nothing
- 1 css - css
- 1 js - js
- 2 css 1 js - css, js not added
- 1 css 2 js - js, css not added
- 2 css 2 js - ?
- 1 combined - js
- 1 combined 1 css - css
- 1 combined 1 js - js
- 1 combined 1 js 1 css - js

modules/system/classes/CombineAssets.php
```diff
@@ -338,19 +338,32 @@ protected function prepareAssets(array $assets)
...
         /*
          * Apply registered aliases
          */
         if ($aliasMap = $this->getAliases($extension)) {
+            if ($extension === null) {
+                $aliasMap = array_map(function ($map) {
+                    return is_array($map) ? $map : [$map];
+                }, $aliasMap);
+                $aliasMap = array_reduce($aliasMap, 'array_merge', []);
+            }
```
LukeTowers: What does this section do?

modules/system/classes/CombineAssets.php
RomainMazB: The actual case I try to solve actually:
Using at least one css not-aliased file, the if branch will match.
Using at least one js not-aliased file, the else if branch will match.
Using only aliased file(s) or an equality of css and js + aliases, both css and js stacks counts are equal, so I merge all of them with a unique rule (because both framework.extra exists as a css alias and js alias).

https://github.com/wintercms/winter/blob/6e8c8d90f11cf38560f3e5f1f1fe3ffac904086c/modules/system/classes/CombineAssets.php#L160-L163

modules/system/tests/classes/CombineAssetsTest.php
RomainMazB: > What happens if both CSS & JS assets are present?

Both will be combined, into a file with the extension of the dominant stack count (if 2 js and 1 css, it will return a `.js` file, not sure for 1 vs 1), there is actually absolutely no filter in the combiner to exclude programmer's errors.

> What about if an alias that is registered for both is used?

With the actual logic, the alias with the extension alphabetically ordered after is dominant (js over css see https://github.com/wintercms/winter/pull/634#discussion_r942267173 to understand why).  
As of today, that's what happens `framework.extras` js alias which overrides the css alias. If you explicitly ask the combiner for a css file (passing a css not-aliased file into the combiner), the css alias will be correctly used.  

But in all of the other cases (equal css/js stacks counts and js dominant), the js alias will be used.

I will add these tests later in the day. Just to be sure what do you call "1 combined"?

modules/system/classes/CombineAssets.php
RomainMazB: When using [the `getAliases` method](https://github.com/wintercms/winter/blob/6e8c8d90f11cf38560f3e5f1f1fe3ffac904086c/modules/system/classes/CombineAssets.php#L765-L776) (to build the `$aliasMap` on the line 361) with a null value, it returns the whole aliases array with extensions as keys:
```php
'css' => [
  'framework.extras' => '~/modules/system/assets/css/framework.extras.css',
  'framework.extras.css' => '~/modules/system/assets/css/framework.extras.css',
  'snowboard.extras.css' => '~/modules/system/assets/css/snowboard.extras.css'
],
'js' => [
  'framework.extras' => '~/modules/system/assets/js/framework.extras.js',
  'framework.extras.js' => '~/modules/system/assets/js/framework.extras.js',
  'snowboard.extras.js' => '~/modules/system/assets/js/snowboard.extras.js'
]
// other extensions aliases ...
```
I don't know at this level which extensions are registered, so I used the `array_reduce` combined with the `array_merge` to not get any responsibility for this.
This results to the one-level array in `$aliasMap` that we need:
```php
[
  'framework.extras.css' => '~/modules/system/assets/css/framework.extras.css',
  'snowboard.extras.css' => '~/modules/system/assets/css/snowboard.extras.css',
  'framework.extras' => '~/modules/system/assets/js/framework.extras.js',
  'framework.extras.js' => '~/modules/system/assets/js/framework.extras.js',
  'snowboard.extras.js' => '~/modules/system/assets/js/snowboard.extras.js'
]
```
_This is the reason of the `framework.extras` js alias dominance over the css one.`_

The reason why I needed to use the "array casting" on line 361-363 into the `array_map` is that I faced something that wasn't an array, but re-reading at the code of `registerAlias`, I don't get how this could happen. I will take a deeper look to understand what happened.

RomainMazB:
@LukeTowers
The more this conversation grows, the more I feel like the combiner needs more refactoring to be more strict about the types.

Something like a third `$type` argument for the `combine` method could be added when combining only aliased assets would avoid all any type guessing for this edge-case.

Plus, we should not allows to combine a css file with a js one. We may throw an exception when this happens.

#### How I see these changes
Here, no problem:
```twig
<link rel="stylesheet" href="{{ ['@framework.extras.css', 'my-file.css'] | theme }}">
```
This will throws an exception as we are trying to combine a js file with a css one
```twig
<link rel="stylesheet" href="{{ ['@framework.extras.css', 'my-file.css', 'my-file.js'] | theme }}">
```
This will throws a **warning** as the system is type-guessing the extension of the alias, which in this case is the wrong one (will return the js alias).
```twig
<link rel="stylesheet" href="{{ ['@framework.extras'] | theme }}">
```
As of today, the only system conflicts is `@framework.extras`, the warning could includes something like >
Warning: both `framework.extras` exists as a js and css alias, you should use `@framework.extras.css` or `@framework.extras.js`.

modules/system/tests/classes/CombineAssetsTest.php
LukeTowers: 1 combined means an alias that refers to both JS & CSS, i.e. `framework.extras`

RomainMazB: I started writing the tests but I'm not sure that we are on the same line on what some tests should return:
- mixed assets type (ie 1 or more css + 1 or more js). As of today it combines "properly" and returns a mixed css/js file, should we keep this behavior without any warning?
- 1 combined and 1 combined 1 js 1 css. As of today the combined asset will be the js one due to js aliases dominance (detailed here: https://github.com/wintercms/winter/pull/634#discussion_r942267173), and I have no way into the combine method to select the css if I would like to.
Plus, if you mean 1 js (or 1 css) as "1 aliased js asset", because aliases are actually added to both the css and js stack, all the 4 last cases will results in the combined to be the js framework for the same reason. The only way to select the css framework is to pass a not-aliased css file to the combiner.

I'm still going to finish writing tests as draft so we can easily modify them together before going further in refactorization.

modules/system/tests/classes/CombineAssetsTest.php
```diff
@@ -107,25 +107,125 @@ public function testResetCache()
...
+        // 1 js
+        $url = $combiner->combine(
+            ['@js1'],
+            base_path() . '/modules/system/tests/fixtures/themes/test'
+        );
+        $this->assertNotNull($url);
+        $response = $this->get($url);
+        $assertContainsScript1($response);
+
+        // 2 css 1 js
```
RomainMazB, line 157: Should we throw a warning when combiner attempts to combine mixed file types?

modules/system/tests/classes/CombineAssetsTest.php
```diff
@@ -107,25 +107,125 @@ public function testResetCache()
...
+            base_path() . '/modules/system/tests/fixtures/themes/test'
+        );
+        $this->assertNotNull($url);
+        $response = $this->get($url);
+        $assertContainsStyle1($response);
+        $assertContainsStyle2($response);
+        $assertContainsScript1($response);
+        $assertContainsScript2($response);
+
+        // 1 combined
```
RomainMazB, line 186: This will always results to js alias, we have no way to select the css file with the current combiner's API

modules/system/tests/classes/CombineAssetsTest.php
```diff
@@ -107,25 +107,125 @@ public function testResetCache()
...
+        $url = $combiner->combine(
+            ['@framework'],
+            base_path() . '/modules/system/tests/fixtures/themes/test'
+        );
+        $this->assertNotNull($url);
+        $response = $this->get($url);
+        $assertContainsFrameworkCss($response);
+
+        // 1 combined 1 css
```
RomainMazB, line 195: Same here and below, the combined results to the js alias, every time due to js dominance when only aliased files are combined.

RomainMazB (commented):
@LukeTowers I'm adding a review to my own draft tests.
When we will agree on what should be the result on each cases, I'll refactorize the AssetsCombiner.

modules/system/tests/classes/CombineAssetsTest.php
LukeTowers: I updated my comment with an initial thought of what should happen for each but I'm quite tired right now and can't really think about it too hard, so perhaps @bennothommo can have a read of this PR and add his thoughts.

RomainMazB: To me:
2 css 1 js - css, js not added > agreed, with a warning to notice the dev why the js is not added
1 css 2 js - js, css not added > agreed, with a warning to notice the dev why the css is not added
2 css 2 js - ? > error: can't guess the wanted asset's type
1 combined 1 css - css > agreed
1 combined 1 js - js > agreed
1 combined 1 js 1 css - js > why js? to me it's the same as 2 css 2 js: error, we can't guess the wanted asset's type

bennothommo:
Bit of a shame we didn't pick this up before Winter 1.2 was released, as I think we probably should've made a (potentially) breaking change here to remove the ambiguity.

Unless I'm not understanding this correctly, the issue is that when using an aliased asset (like `@framework.extras`), it's automatically adding this asset to `compileJs` and `compileCss`, then using the extensions of the other assets in the same group to determine what is actually being compiled, even if the aliased asset has an extension included.

The way I see it, the easiest way to remove the ambiguity would be to introduce a parameter for the `| theme` filter, to set a specific combine step. For example, we could have the following to assume that the given paths will be combined as CSS files.

```html
<link rel="stylesheet" href="{{ ['@framework.extras.css', 'my-file.css', 'my-file.js'] | theme('css') }}">
```

Wtih the above, we would consider the framework extras and "my file" CSS files to be combined into one, and simply ignore (or throw a log) for the JS file included.

For the following:

```html
<link rel="stylesheet" href="{{ ['@framework.extras', 'my-file.js'] | theme('js') }}">
```

We would combine the JS file for the framework extras alias, and combine the "my file" JS script as well.

If the extra parameter is not included, then we could keep the current behaviour, even though it's broken. We'd probably just have to update the docs to point out that automatically detecting the correct combination has scenarios (like this one) that it can't handle, and to specifically use one of the parameters above.

LukeTowers:
To be honest I'd prefer we move away from supporting ambiguous aliases entirely.

RomainMazB:
@bennothommo 
Yes, unfortunately we didn't realized this issue before 1.2 so we could have handled it more elegantly, introducing a BC.

But this also means that this issue is not that huge. To be honest I discovered it and then fixed it instantly by using an extra css file which "balanced" correctly the _weak_ type guessing system.

I then just dove into the core's code to understand what happened and discovered all that stuff.

From a WinterCMS user's point of view it's like "oh ok, that's how it works, maybe it should have been documented".
From a core dev's one it's like "what's that crap?".

Now that we discovered it, we should just decide how we could improve this part of the code, but 99.9% of WinterCMS wouldn't have discover this issue.

Most of the use cases could be solved by adding an extra file or using "the typed alias": `framework.extras.css` (or .js) to select one or the other asset, and I don't think that any of us use the `registerAlias` method that could introduce any other ambiguity.

RomainMazB:
Some additional draft work experiencing an alternative, I know that throwing an exception introduce a BC.  

The main goal here is to reverse the type guessing system to simplify it and enforce it at the same time:
Instead of adding the alias name into the `combineCss` and `combineJs` variable, the system now resolves the full asset path.
To me, the only question to solve now is "what to do in the `else` branch of the type guessing system?":
- Throw an exception?
- Keep the dominance of one asset type and just throw a warning?

RoKrypto:
Greetings.

I'm having a similar issue but when adding the Snowboard framework through its aliases...
in CSS...
image

and JS.
image

Is there an estimate time for the rollout of the fix to this issue?

RomainMazB:
@RoKrypto 
ATM you could fix it yourself at project-level, just merging the aliased asset with one of your asset.

As an example in your JS error problem, you can merge the assets from the line 60 which seems to use the `| theme` filter too, with the aliased assets.
For the css one, I guess that you import at some point a css file using the `| theme` filter too.

LukeTowers:
Closing for now as I think @RomainMazB is no longer actively using the project. If anyone else has this or a similar issue and wants this to be merged then just ping me and I'll reopen it.

PR #639, opened 2022-08-06
Getting and storing system boot date value with Parameter model

arvislacis:
This PR changes the logic how the system boot date (appBirthday) parameter value is handled by introducing `system::app.birthday` Parameter. Also additional logic is added to handle parameter value in existing Winter CMS installations:

1. If no `system::app.birthday` Parameter value exists then we try to get the installation date of the first plugin of system (this is the current behavior); if plugin version value exists then we store it as `system::app.birthday` value so we don't need to do this step again in the future;
2. If step 1 fails (system has no plugins installed) then we assume that app birthday is today and we set `system::app.birthday` value.

One additional note - `system::app.birthday` value also could be set in `WinterInstall.php` for all new installations but currently I am not sure where is the best place to add that logic, my current guess is - at the end part of `handle()` function, right before `$this->displayOutro()`.

### DIFF
--- a/modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
+++ b/modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
@@ -5,7 +5,7 @@ use Carbon\Carbon;
 use System\Models\Parameter;
 use System\Models\PluginVersion;
 
-class DbSystemAddAppBirthdayDate extends Migration
+return new class extends extends Migration
 {
     public function up()
     {
--- a/modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
+++ b/modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
@@ -1,0 +1,32 @@
+<?php
+
+use Winter\Storm\Database\Updates\Migration;
+use Carbon\Carbon;
+use System\Models\Parameter;
+use System\Models\PluginVersion;
+
+class DbSystemAddAppBirthdayDate extends Migration
+{
+    public function up()
+    {
+        $appBirthday = Parameter::get('system::app.birthday');
+
+        if (is_null($appBirthday)) {
+            $appBirthdayFromPlugins = PluginVersion::orderBy('created_at')->first()?->created_at;
+
+            if (is_null($appBirthdayFromPlugins)) {
+                $appBirthday = Carbon::now()->timestamp;
+            } else {
+                $appBirthday = Carbon::parse($appBirthdayFromPlugins)->timestamp;
+            }
+
+            Parameter::set('system::app.birthday', $appBirthday);
+        }
+
+        return $appBirthday;
+    }
+
+    public function down()
+    {
+    }
+}
--- a/modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
+++ b/modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
@@ -5,7 +5,7 @@ use Carbon\Carbon;
 use System\Models\Parameter;
 use System\Models\PluginVersion;
 
-return new class extends extends Migration
+return new class extends Migration
 {
     public function up()
     {
@@ -27,4 +27,4 @@ return new class extends extends Migration
     public function down()
     {
     }
-}
+};
--- a/modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
+++ b/modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
@@ -11,17 +11,14 @@ class DbSystemAddAppBirthdayDate extends Migration
     {
         $appBirthday = Parameter::get('system::app.birthday');
 
-        if (is_null($appBirthday)) {
-            $appBirthdayFromPlugins = PluginVersion::orderBy('created_at')->first()?->created_at;
-
-            if (is_null($appBirthdayFromPlugins)) {
-                $appBirthday = Carbon::now()->timestamp;
-            } else {
-                $appBirthday = Carbon::parse($appBirthdayFromPlugins)->timestamp;
-            }
-
-            Parameter::set('system::app.birthday', $appBirthday);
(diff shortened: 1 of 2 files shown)


bennothommo:
@arvislacis I'm wondering if we make it part of a (new) migration, that way, no matter which way the installation takes place, it should theoretically be run and set the birthday of the install.

arvislacis:
@bennothommo Yeah, migration probably would be the best solution for new installations and probably also the existing ones (so the migration and logic is executed once). So you suggest to move content of `getAppBirthday` function (from this PR) to new migration file (probably seed type migration under `winter/modules/system/database/seeds/SeedAppBirthday.php`) in `run()` function of migration?

bennothommo:
Yeah I think that's the best course. Just make sure as part of the seed that it doesn't write a value for the birthday if one already exists.

arvislacis:
@bennothommo Changed PR to migration file format in the last commit.

LukeTowers (approved):

bennothommo:
@arvislacis one more thing, could you do two unit tests:
- One that runs `winter:up` and checks that the birthday is today's date (or maybe use [`Carbon::setTestNow()`](https://carbon.nesbot.com/docs/#api-testing) to set a particular date for the birthday).
- If possible, set up a second that runs all migrations, then removes that the birthday migration has been run, and install a plugin at a set date, then run that migration. That should be a test case for people who have pre-existing installations and run the migration.

arvislacis:
@bennothommo Are you expecting me to add mentioned tests in the source code (in another commit) or make these tests manually, locally?

If we are talking about local tests then yesterday I already did tests with `winter:up` and manually adding/deleting/modifying plugin install dates in `system_plugin_versions` table. Tests worked as expected in situations where there are no installed plugins (sets the today's timestamp) and with plugins installed (sets the timestamp of oldest plugin installed).

ericp-mrel:
I would recommend re-writing the migration file to something like this to avoid the deep nesting of if-else statements which should make it easier to follow the code.

But, that's just my opinion.

```php
<?php

use Winter\Storm\Database\Updates\Migration;
use Carbon\Carbon;
use System\Models\Parameter;
use System\Models\PluginVersion;

class DbSystemAddAppBirthdayDate extends Migration
{
    public function up()
    {
        $appBirthday = Parameter::get('system::app.birthday');

        if ($appBirthday) {
            return $appBirthday;
        }

        $appBirthdayFromPlugins = PluginVersion::query()
            ->orderBy('created_at')
            ->first()
            ?->created_at;

        $appBirthday = $appBirthdayFromPlugins
            ? Carbon::parse($appBirthdayFromPlugins)->timestamp
            : Carbon::now()->timestamp;

        Parameter::set('system::app.birthday', $appBirthday);

        return $appBirthday;
    }

    public function down()
    {
    }
}
```

arvislacis:
@ericp-mrel Done in https://github.com/wintercms/winter/pull/639/commits/6624d2274c502f609234cc8420f066a741e5206b , thanks for the recommendation.

Re-tested scenarios with the new commit - code works as expected.

modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
```diff
@@ -0,0 +1,29 @@
...
+            return $appBirthday;
+        }
+
+        $appBirthdayFromPlugins = PluginVersion::orderBy('created_at')->first()?->created_at;
+        $appBirthday = $appBirthdayFromPlugins ? Carbon::parse($appBirthdayFromPlugins)->timestamp : Carbon::now()->timestamp;
+
+        Parameter::set('system::app.birthday', $appBirthday);
+
+        return $appBirthday;
+    }
```
LukeTowers:
```suggestion
    public function up()
    {
        $appBirthday = Parameter::get('system::app.birthday');
        if ($appBirthday) {
            // Return early if the birthday has already been set
            return;
        }

        $firstPluginCreated = PluginVersion::orderBy('created_at')
            ->first()
            ?->created_at;

        $appBirthday = $firstPluginCreated ?: Carbon::now();

        Parameter::set('system::app.birthday', $appBirthday);
    }
```

LukeTowers: created_at is a Carbon instance when set and Carbon will handle being turned into a string for storage and the System\Helpers\DateTime will handle turning it back into a Carbon instance when rendering.

modules/system/database/migrations/2022_08_06_000026_Db_System_Add_App_Birthday_Date.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+use Winter\Storm\Database\Updates\Migration;
+use Carbon\Carbon;
+use System\Models\Parameter;
+use System\Models\PluginVersion;
+
+class DbSystemAddAppBirthdayDate extends Migration
```
LukeTowers: We can use anonymous classes for migrations now.
```suggestion
return new class extends extends Migration
```

arvislacis:
@LukeTowers Fixed minor typos/bugs you introduced with https://github.com/wintercms/winter/pull/639/commits/249edcb81d900eb6a3c49164f4a3e3a554ed045c by rewriting migration.

I re-tested test cases with and without plugins installed - works fine with latest commit. Speaking about https://github.com/wintercms/winter/pull/639/commits/5caea9dcd54e7a3ab7345d05fbdd4324f9bcbec2 - well, there is actually a minor difference in date string which will be saved now by your changes - for example, if no plugin is installed and `Carbon:now()` case executes then date string will be like `2022-08-10T17:18:37.818388Z` but if there are plugins installed then value retrieved from DB and stored in parameter would be like `2022-08-10 17:18:37` (at least with SQLite DB I tested) - nothing that breaks functionality but just a side note.

LukeTowers:
Either value can be safely parsed by Carbon so it's not a big deal, but if anyone really cares we can probably just call toDateTimeString() before storing it.

LukeTowers:
@bennothommo any final input?

bennothommo:
Sorry I had forgot to answer @arvislacis question about the tests - I was hoping to have unit tests in code for both scenarios, but I figured that might be hard(-ish) to pull off, so I'm happy for it to be merged.

PR #642, opened 2022-08-08
Add support for exit codes in MixCompile

jaxwilko:
This PR adds support for returning exit codes from the `mix:compile` command. This is important for CI/CD pipelines that need to exit early is there is a compilation failure.

What I would like is a flag like phpunit's `--stop-on-error` which would exit the compile when the first error is encountered, but to do that I needed to fix exit code support and add a test case.

The tests will only run if there is a `node_modules` in the base_path, this means these tests wont currently be ran automatically in GitHub actions and we'll need to update the workflows if we want them. I'm not sure if we do, probably, either way these tests can be ran locally.

Added with this to allow for configuring the `mix:compile` without re-writing the root `package.json` file is the `NPM_PACKAGE_MANIFEST` env arg. The reason I did it this way was that env args are passed to the command but I was not sure if `Config::set`s would be. It is not ideal, maybe it would be better as a config option or argument to the command, thoughts?


modules/system/console/MixCompile.php
```diff
@@ -48,6 +48,10 @@ public function handle(): int
         $registeredPackages = $mixedAssets->getPackages();
         $requestedPackages = $this->option('package') ?: [];
 
+        if (is_string($requestedPackages)) {
```
LukeTowers, line 56:
```suggestion
        // Calling commands in unit tests can cause the option casting to not work correctly,
        // ensure that the option value is always an array
        if (is_string($requestedPackages)) {
```

modules/system/console/MixCompile.php
```diff
@@ -124,20 +135,26 @@ protected function canCompilePackage(string $mixJsPath): bool
...
         return in_array($this->getPackagePath($mixJsPath), $workspacesPackages);
     }
 
+    /**
+     * Read the package.json file for the project, path configurable with the
+     * `NPM_PACKAGE_MANIFEST` env option
+     */
+    protected function readNpmPackageManifest(): array
+    {
+        $packageJsonPath = base_path(env('NPM_PACKAGE_MANIFEST', 'package.json'));
```
LukeTowers: Thinking on it more @jaxwilko this would be better as an option rather than an env() variable. Probably `--manifest` would make the most sense.

jaxwilko: Added in https://github.com/wintercms/winter/pull/642/commits/2b0228b45afaa22e7a30427fc0a16464f7365d91

jaxwilko:
I've now added support for `--silent` to make mix / webpack "shhh" and `--stop-on-error` to early exit when the first compile fails

PR #650, opened 2022-08-17
Add support for data-request-parent

LukeTowers:
See https://github.com/octobercms/october/pull/5136

Also see:
- octobercms/october#5084 
- octobercms/october#4925 
- octobercms/october#2727 (Somewhat related, not entirely though)
- https://github.com/octoberrain/test-plugin/issues/73 (Somewhat related)
- octobercms/october#5447 (may be related)


LukeTowers:
See https://github.com/wintercms/wn-test-plugin/commit/2b753f2d0143fbcf2a19b35cbb6f8b1718fa9682 for an applied test case and / or add a recordfinder field to a relationcontroller update popup and attempt to use the search or pagination  on the list of possible records to choose from for testing.

modules/system/assets/js/framework.js
```diff
@@ -479,6 +508,68 @@ if (window.jQuery.request !== undefined) {
...
+            parentDataObjects = [serializeFormToObj($form)],
+            parentFormData = {};
+
+        var findParentForms = function ($form) {
+            // If the form element exists and has a parent element defined
+            if ($form.length && $form.data('request-parent')) {
+                // Identify the owning form of the parent element
+                var $parentEl = $($form.data('request-parent'));
+                if ($parentEl.length) {
+                    var parentEmbeddedData = {};
```
bennothommo, line 543: @LukeTowers unless I'm mistaken, this seems to be a dangling reference. It's extended upon in the next couple of lines, but it's not returned or added to a parent variable?

modules/system/assets/js/framework.js
```diff
@@ -494,6 +585,17 @@ if (window.jQuery.request !== undefined) {
         }
     }
 
+    function serializeFormToObj($form) {
```
bennothommo, line 595:
```suggestion
    function serializeFormToObj ($form) {
```

modules/system/assets/js/framework.js
```diff
@@ -494,6 +585,17 @@ if (window.jQuery.request !== undefined) {
         }
     }
 
+    function serializeFormToObj ($form) {
```
bennothommo: Made a whoops

```suggestion
    function serializeFormToObj($form) {
```

modules/system/assets/js/framework.js
LukeTowers: ?

modules/system/assets/js/framework.js
```diff
@@ -91,6 +91,27 @@ if (window.jQuery.request !== undefined) {
...
+
+        // If the containing form has a request-parent element specified, then we move to that element
+        // and process it's data as if it was the element creating the request, including checking for a
+        // parent element to that request. Once we've gone all the way up until there are no more parent
+        // elements detected then the parent data is merged such that the closest data in the chain to the
+        // original element that triggered the request has the highest priority.
+
+
+        // Loop through every parent element with the data-request-data attribute and attach the data
+        // to the request data object with closer element data overwriting further element data
```
bennothommo: @LukeTowers is this comment needed when we've got an explanatory comment further down, in the `getParentFormData` method?

LukeTowers: Probably not, it was more for my own reference; I'll go through and clean it up tomorrow

modules/system/assets/js/framework.js
LukeTowers: That would be correct, it needs to be added to the merge stack, will figure out what order it needs to be added in tomorrow.

PR #655, opened 2022-08-17
Fix nested tree listsNested method returning unordered hierarchy

RomainMazB:
Hi!
When toying with the `NestedTree` trait, I discovered something which looks to me like an unwanted behavior:
The `listsNested` scope doesn't order properly the tree hierarchy. Actually it doesn't order anything as of today and just returns the collection in the order of the database entries.

In this draft PR, I just mixed the `NestedTreeModelTest` entries to make the test fail and discuss if this is the wanted behavior or if this should be fixed. I just moved the `September` and `October` savings at the end of the seeder method.

I expected that it wouldn't change anything but it does change the order.

Before doing the modification, I'd like to have some thoughts about that.

_There is some other modifications to fix some typos_

### DIFF
--- a/modules/system/tests/plugins/database/NestedTreeModelTest.php
+++ b/modules/system/tests/plugins/database/NestedTreeModelTest.php
@@ -115,12 +115,32 @@ class NestedTreeModelTest extends PluginTestCase
 
         $autumn = $orange->children()->create([
             'name' => 'Autumn Leaves',
-            'description' => 'Disccusion about the season of falling leaves.'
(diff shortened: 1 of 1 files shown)


RomainMazB:
Ok... I think I just been misleaded about what causes the list entries mix on the current tests:

The tests use the `id` as the second parameter for the `listsNested` method call, which have effect to use the id as the list array keys. Then php does the rest of the work and re-order the array using the keys.

As of today, there is just no way to ensure the order of the `listsNeeded` scope due to the actual API.

PR #664, opened 2022-08-30
Add support for overriding phpunit bootstrap

jaxwilko:
This PR allows for testables to implement their own bootstrap file.

Configurable via the `boostrap` attribute in `phpunit.xml`

### DIFF
--- a/modules/system/console/WinterTest.php
+++ b/modules/system/console/WinterTest.php
@@ -141,6 +141,13 @@ class WinterTest extends Command
         // Resolve the configuration path based on the current working directory
         $config = realpath($config);
         $bootstrapPath = base_path('modules/system/tests/bootstrap/app.php');
+        
+        if (
+            ($configBoostrap = (string) simplexml_load_file($config)['bootstrap'])
+            && str_starts_with($configBoostrap, './')
+        ) {
+            $bootstrapPath = dirname($config) . ltrim($configBoostrap, '.');
+        }
 
         $process = new Process(
             array_merge([$this->phpUnitExec, '--configuration=' . $config, '--bootstrap=' . $bootstrapPath], $args),
--- a/modules/system/console/WinterTest.php
+++ b/modules/system/console/WinterTest.php
@@ -142,11 +142,8 @@ class WinterTest extends Command
         $config = realpath($config);
         $bootstrapPath = base_path('modules/system/tests/bootstrap/app.php');
         
-        if (
-            ($configBoostrap = (string) simplexml_load_file($config)['bootstrap'])
-            && str_starts_with($configBoostrap, './')
-        ) {
-            $bootstrapPath = dirname($config) . ltrim($configBoostrap, '.');
+        if ($configBoostrap = (string) simplexml_load_file($config)['bootstrap']) {
+            $bootstrapPath = dirname($config) . DIRECTORY_SEPARATOR . $configBoostrap;
         }
 
         $process = new Process(


modules/system/console/WinterTest.php
```diff
@@ -141,6 +141,13 @@ protected function execPhpUnit(string $config, array $args): int
         // Resolve the configuration path based on the current working directory
         $config = realpath($config);
         $bootstrapPath = base_path('modules/system/tests/bootstrap/app.php');
+        
+        if (
+            ($configBoostrap = (string) simplexml_load_file($config)['bootstrap'])
+            && str_starts_with($configBoostrap, './')
+        ) {
+            $bootstrapPath = dirname($config) . ltrim($configBoostrap, '.');
+        }
```
LukeTowers:
```suggestion
        if ($configBoostrap = (string) simplexml_load_file($config)['bootstrap']) {
            $bootstrapPath = dirname($config) . DIRECTORY_SEPARATOR . $configBoostrap;
        }
```

bennothommo:
@jaxwilko @LukeTowers I think we may have to revert this - this breaks the ability to test plugins pre-1.2 as well as 1.2 (see https://github.com/wintercms/wn-user-plugin/pull/24#issuecomment-1261132275) because now the tests in 1.2 will use the `bootstrap` config in `phpunit.xml`, which needs to be pointing to the old tests bootstrap script to allow the tests pre-1.2 to work.

If people want to use custom bootstraps for tests in 1.2 and above, I would instead suggest we provide an option for the `winter:test` command that allows one to specify the custom bootstrap.

LukeTowers:
@bennothommo phpunit already supports the `--bootstrap` option, so I think we should instead change the tests that need to be run targeting pre-1.2 bootstrap file location to include that flag on their pre-1.2 test runs rather than saying we can't use this XML attribute ever again.

bennothommo:
There's nothing stopping us from adding a `--bootstrap` argument to our own `winter:test` command which can act as an override as opposed to extracting it out of the XML config.

All I'm saying is, the functionality previous to this PR was able to handle both pre-1.2 and 1.2 tests with no issue. This PR stops that from happening.

I had specifically updated the `winter:test` command to override the bootstrap because not only would it be backwards-compatible (and also fix the inherent BC break that was introduced when we moved the whole testing suite out of the `tests` folder in the first place), but because it was one less thing people had to worry about configuring.

jaxwilko:
@bennothommo totally not opposed to reverting the PR and providing a different way to offer the functionality, but might it be better to always pass the default bootstrap to phpunit unless one is specified, then remove it from plugins that are being explicit where they don't need to?

bennothommo:
@jaxwilko the problem is we can't remove it from the `phpunit.xml` as that would break the pre-1.2 testing, where the bootstrap had to be explicitly defined to work.

For example, in the User plugin, we're not using different branches for the different Winter branches, because the plugin should theoretically work with 1.1.x and 1.2.x, so the expectation currently is that any changes going forward need to be able to support both, hence why we test both. But we can't use the same `bootstrap` value for both in `phpunit.xml`, because of the different placements of the test bootstrap script.

That was why I opted to make it so that `winter:test` overrides it for 1.2. It was to ensure that any plugins that have been written for 1.1.x or before could still use the testing suite without making changes to their PHPUnit config or running some form of split of the codebase for the different branches (albeit having to change which class they extend for the test case classes, but that is unavoidable).

We really have to make testing as easy as possible to encourage people to do it, as it's already a complex beast as it is.

I'm sorry I didn't bring this up before - I didn't realise the ramifications of the change until I saw it in action. I understand the reasoning for you needing it in your case, which is why I still want to support it in some fashion, but I don't think it's a scenario that is going to come up all that often, so I feel it's acceptable for this to be delegated to an option in the command or some sort of custom configuration.

LukeTowers:
@bennothommo my concern is that the bootstrap attribute is already an officially supported attribute in phpunit.xml, and not supporting it in winter:test breaks compatibility with phpunit, since now running `phpunit` and `winter:test` have different outcomes.

IMO the plugins should use the `--bootstrap` flag for phpunit to run the pre-1.2 test suites when running in the pre-1.2 environment, is that not possible to achieve in the github actions?

bennothommo:
@LukeTowers

> my concern is that the bootstrap attribute is already an officially supported attribute in phpunit.xml, and not supporting it in winter:test breaks compatibility with phpunit, since now running phpunit and winter:test have different outcomes.

Agreed, but moving the bootstrap in the first place broke every plugin's test suites until they edited their `phpunit.xml` configs. My goal was to gracefully support old test cases. Also, slightly OCD, but I was never a fan of making people have to add `bootstrap="../../modules/system/tests/bootstrap.php"` to their configs - it just didn't look "clean" to me with the relative pathing.

> IMO the plugins should use the --bootstrap flag for phpunit to run the pre-1.2 test suites when running in the pre-1.2 environment, is that not possible to achieve in the github actions?

I think the better argument is if we are going to force people to update their configs, and you want support for custom bootstrap scripts, then we shouldn't be populating the `--bootstrap` argument in `winter:test` at all and restore the default PHPUnit functionality. Essentially, rollback @jaxwilko's changes as well as my own.

LukeTowers:
hmm, what were the original benefits of your original approach again? Would like to consider them one more time before deciding to go back to what it was.

bennothommo:
That was the only benefit - overriding the bootstrap so that people didn't have to change their PHPUnit configs. There was minor cleanup of the `winter:test` command as well, but nothing consequential.

LukeTowers:
@bennothommo I think that it can still be worth keeping, we just only apply it if there is no bootstrap attribute specified, that way we can make life simpler for people while also supporting the PHPUnit standards. Any thoughts on that?

bennothommo:
@LukeTowers it still doesn't resolve the configs created pre-1.2 - they will likely all have `bootstrap` definitions pointing to the wrong location if they begin testing in 1.2.

How about this - we automatically overwrite the `bootstrap` definition if it was pointing to the old location (ie. `tests/bootstrap.php`) when running in 1.2, but otherwise honor the `bootstrap` definition if it's pointing somewhere else?

LukeTowers:
@bennothommo I think that is only a concern for the first party Winter plugins running on GitHub actions since anyone upgrading their plugins for v1.2 is going to be following the upgrade guide to update the base test case classes they extend in the first place so also updating their bootstrap attribute is not unreasonable to require.

Implementing your suggestion will also cause issues with people that upgrade to 1.2 but keep the old /tests folder present and thus none of their setup should actually be changed until they change it on purpose. I personally think the best option is to make it so that we only apply our presumptive logic about where the bootstrap file is located when there is no bootstrap file specified and then for all the first party plugins testing on 1.2 and 1.1 we just include the `--bootstrap` flag in their 1.1 runs.

bennothommo:
Alright, I'll change the tests around on the first-party plugins.

PR #668, opened 2022-09-02
Add support for input range

robertalexa:
This PR adds support for the input range as a form field.

Screenshot:
image

By default, input ranges do not have a way to display their current value, so I have added an extra element that follows the handle.

Here is the PR on the docs repo that should ideally be merged in at the same time: https://github.com/wintercms/docs/pull/91

### DIFF
--- a/modules/backend/widgets/form/partials/_field_range.php
+++ b/modules/backend/widgets/form/partials/_field_range.php
@@ -3,16 +3,16 @@
     <span class="form-control"><?= isset($field->value) ? e($field->value) : '&nbsp;' ?></span>
 <?php else: ?>
     <?php
-        $min = $field->config['min'] ?? 0;
-        $max = $field->config['max'] ?? 100;
-        $step = $field->config['step'] ?? 1;
-        $value = $field->value;
-        if ($min > $max) {
-            $min = $max - $step;
-        }
-        if (is_null($value)) {
-            $value = ($min + $max) / 2;
-        }
+    $min = $field->config['min'] ?? 0;
+    $max = $field->config['max'] ?? 100;
+    $step = $field->config['step'] ?? 1;
+    $value = $field->value;
+    if ($min > $max) {
+        $min = $max - $step;
+    }
+    if (is_null($value)) {
+        $value = ($min + $max) / 2;
+    }
     ?>
 
     <input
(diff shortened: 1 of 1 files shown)


modules/backend/widgets/form/partials/_field_range.php
```diff
@@ -0,0 +1,33 @@
...
+        $(document).ready(function () {
+            const input = document.getElementById("<?= $field->getId() ?>");
+            input.addEventListener("input", function () {
+                this.nextElementSibling.innerHTML = this.value;
+                var pos = ((this.value - <?= $min ?>) / (<?= $max ?> - <?= $min ?>) * 100);
+                this.nextElementSibling.style.left = `calc(${pos}% + ${8 - pos * 0.15}px)`;
+            });
+            input.dispatchEvent(new Event('input'));
+        });
+    </script>
```
LukeTowers, line 39: Does this work if the field is dynamically displayed after being loaded over AJAX? Typically field JS should be added as it's own JS file (ideally written as a Snowboard plugin now) that binds to events at the document root and filters based on the current input, generally with a selector along the lines of `[data-control="range"]`

modules/backend/widgets/form/partials/_field_range.php
```diff
@@ -0,0 +1,33 @@
...
+    <script>
+        $(document).ready(function () {
+            const input = document.getElementById("<?= $field->getId() ?>");
+            input.addEventListener("input", function () {
+                this.nextElementSibling.innerHTML = this.value;
+                var pos = ((this.value - <?= $min ?>) / (<?= $max ?> - <?= $min ?>) * 100);
+                this.nextElementSibling.style.left = `calc(${pos}% + ${8 - pos * 0.15}px)`;
+            });
+            input.dispatchEvent(new Event('input'));
+        });
```
LukeTowers:
```suggestion
        (() => {
            const input = document.getElementById("<?= $field->getId() ?>");
            input.addEventListener("input", function () {
                this.nextElementSibling.innerHTML = this.value;
                var pos = ((this.value - <?= $min ?>) / (<?= $max ?> - <?= $min ?>) * 100);
                this.nextElementSibling.style.left = `calc(${pos}% + ${8 - pos * 0.15}px)`;
            });
            input.dispatchEvent(new Event('input'));
        })();
```

PR #672, opened 2022-09-05
Clean up of base plugin test case class

bennothommo:
- Deprecate `runPluginRefreshCommand` method and replace with `instantiatePlugin`, to better describe the functionality within.
- Run instantiated plugin's `boot` method to ensure all needed functionality is correctly loaded.
- Use namespace to guess plugin being detected, falling back to path if there is no namespace given - this fixes testing for symlinked plugins.
- Add type hints and return types for most methods

This should fix the issues in some PRs ([like this one](https://github.com/wintercms/wn-redirect-plugin/pull/10#issuecomment-1236415903)) where the plugin's `boot` method did not seem to load.

### DIFF
--- a/modules/system/tests/bootstrap/PluginTestCase.php
+++ b/modules/system/tests/bootstrap/PluginTestCase.php
@@ -88,7 +88,10 @@ abstract class PluginTestCase extends TestCase
         $this->pluginTestCaseLoadedPlugins = [];
 
         // Detect the plugin being tested and load it automatically
-        $this->guessAndLoadPlugin();
+        $pluginCode = $this->guessPluginCode();
+        if (!is_null($pluginCode)) {
+            $this->instantiatePlugin($pluginCode, false);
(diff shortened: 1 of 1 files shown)


modules/system/tests/bootstrap/PluginTestCase.php
```diff
@@ -157,49 +159,51 @@ protected function runPluginRefreshCommand($code, $throwException = true)
...
+        if ($firstLoad) {
+            $manager->bootPlugin($plugin);
+        }
     }
 
     /**
      * Returns a plugin object from its code, useful for registering events, etc.
-     * @return PluginBase
+     *
+     * @return PluginBase|null
```
LukeTowers:
```suggestion
```
Docblock type hints are unnecessary when real type hints are used

LukeTowers (changes requested):

PR #686, opened 2022-09-12
Explicitly define an index on component list hidden fields

### LINKED ISSUE #683: [v1.2] CMS editor doesn't save all components added to page/layout/etc
### Winter CMS Build

dev-develop

### PHP Version

8.0

### Database engine

MySQL/MariaDB

### Plugins installed

Winter.Pages

### Issue description

Adding multiple components to a layout/page from the CMS editor does not work, and only the last added component is saved to the actual file itself. All other components are lost.

### Steps to replicate

1. Install fresh winter cms project
2. Install Winter.Pages plugin
3. Add `static page` and `static menu` components to the default layout.
4. Save
5. View the `layouts/default.htm` file from your IDE/editor and see only one component is actually saved to the file.

### Workaround

Add components manually from your editor/IDE without using the backend.

bennothommo:
Fixes #683

It appears that either the browsers, or perhaps the framework, are not picking up component field names with the "append" array syntax in the field name (ie. `component_properties[]`, `component_aliases[]`). The result of this is that it is discarding the properties of earlier components in the component list and only saving the last component when saving a CMS partial.

This PR adds explicit indexes to the field names to ensure that the data is correctly sent across as indexed values, similar to the repeater or data table widget.

I've also included a slight tweak to the alias validation to check for duplicate alias *before* writing the alias to the component, thereby ensuring all aliases are unique.

### DIFF
--- a/modules/cms/assets/js/winter.cmspage.js
+++ b/modules/cms/assets/js/winter.cmspage.js
@@ -399,7 +399,7 @@
         var element = ev.target,
             values = JSON.parse($('[data-inspector-values]', element).val())
 
-        $('[name="component_aliases[]"]', element).val(values['oc.alias'])
+        $('[name^="component_aliases["]', element).val(values['oc.alias'])
         $('span.alias', element).text(values['oc.alias'])
     }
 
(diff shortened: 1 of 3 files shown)


mjauvin:
I confirm this resolves the components not being saved.

I also confirm deleting a component in a layout doesn't work properly (it shows the component is removed, but component is still present when reloading the page after saving it).

bennothommo:
Replaced with https://github.com/wintercms/winter/pull/689.

PR #687, opened 2022-09-12
Harden Snowboard

bennothommo:
This applies some hardening to the core classes in Snowboard, preventing people from accidentally (or intentionally) overwriting or misusing system methods. This also helps to prevent [prototype pollution](https://portswigger.net/daily-swig/prototype-pollution-the-dangerous-and-underrated-vulnerability-impacting-javascript-applications).


bennothommo:
I've tested this on the new website (and Docs theme), which use extensive Snowboard plugins and have not encountered any issues with this hardening in place, so just need the tick of approval on code style. Happy to explain how this works too if need be :stuck_out_tongue:

LukeTowers (approved):

PR #689, opened 2022-09-13
Correctly handle arrayed POST data in AJAX calls

### LINKED ISSUE #683: [v1.2] CMS editor doesn't save all components added to page/layout/etc

bennothommo:
The changes introduced in https://github.com/wintercms/winter/commit/a4cfd0d23f4342f530dc4b425098ee24086cd8d1 have broken arrayed POST values, as the original `serializeFormToObj` method that converts the form data to an object overwrites values with the same key, which includes arrayed data (ie. a field with `name[]` and value `foo` will be overwritten by a following field `name[]` and value `bar` - only `bar` will remain in the object)

This fix leverages the `FormData` class to correctly send back arrayed POST data as part of its `entries` loop, and saves the data as an array as necessary.

Fixes https://github.com/wintercms/winter/issues/683.
Will also fix the issue reported here: https://github.com/wintercms/winter/issues/683#issuecomment-1244682887


bennothommo:
Note that the syntax used here will make the original framework unable to work with Internet Explorer 11 any longer (it works with every other browser version released in the last 5-6 years).

I'm OK with this, especially if this fix is only deployed for the 1.2 branch, but worth noting anyway.

LukeTowers:
@bennothommo I can be fine with dropping IE 11 as well, but can we add a unit test for this change so that it doesn't come up again in the future?

bennothommo:
@LukeTowers I'm happy to take a crack at the unit test in another PR. I need this fix now (the bug it fixes messed up one of my client sites today).

LukeTowers:
Works for me, feel free to merge it

Issue #695, opened 2022-09-15
Mail queue is not instantiable


webbati:
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Issue description

Unable to run the mail queue

Full stack trace:

```
Illuminate\Contracts\Container\BindingResolutionException: Target [Illuminate\Contracts\Mail\Factory] is not instantiable. in /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Container/Container.php:1091
Stack trace:
#0 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Container/Container.php(884): Illuminate\Container\Container->notInstantiable()
#1 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Container/Container.php(756): Illuminate\Container\Container->build()
#2 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(860): Illuminate\Container\Container->resolve()
#3 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Container/Container.php(692): Illuminate\Foundation\Application->resolve()
#4 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(845): Illuminate\Container\Container->make()
#5 /var/www/html/my.domain.tld/vendor/winter/storm/src/Foundation/Application.php(304): Illuminate\Foundation\Application->make()
#6 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(181): Winter\Storm\Foundation\Application->make()
#7 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(124): Illuminate\Container\BoundMethod::addDependencyForCallParameter()
#8 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(36): Illuminate\Container\BoundMethod::getMethodDependencies()
#9 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Container/Util.php(41): Illuminate\Container\BoundMethod::Illuminate\Container\{closure}()
#10 /var/www/html/my.domain.tld/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(93): Illuminate\Container\Util::unwrapIfClosure()
[...]

PR #697, opened 2022-09-15
Aliases for Mail manager and factory

webbati:
Aliases necessaries for resolve and doing initiatable mail service on queue.

see issue https://github.com/wintercms/winter/issues/695


LukeTowers:
Merged in https://github.com/wintercms/storm/commit/9dea96030b6255d6006252bf3c413658dad4f4d6

PR #698, opened 2022-09-16
Add "mix:run" command

bennothommo:
Allows running of scripts defined in the `package.json` through Artisan.

As an example, the System module now has a `test` script defined in its `package.json` to run the JavaScript test cases. The command `php artisan mix:run module-system test` now runs those tests, giving people a simple API to run their Mix-enabled plugins and themes.


modules/system/console/MixRun.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $command = ($this->argument('additionalArgs')) ?? [];
+        if (count($command)) {
+            array_unshift($command, 'npm', 'run', $script, '--');
+        } else {
+            array_unshift($command, 'npm', 'run', $script);
+        }
+
+        echo `pwd`;
+        echo base_path($package['path']);
```
LukeTowers: Debugging statements?

modules/system/tests/js/README.md
```diff
@@ -1,32 +0,0 @@
...
-```bash
-php artisan mix:compile --package snowboard
-```
-
-You can also watch the framework for any changes, which will automatically run a build after a change is made. This will
-make development and testing in parallel much quicker:
-
-```bash
-php artisan mix:watch snowboard
-```
```
LukeTowers, line 32: Where is this file being moved to?

modules/system/console/MixRun.php
```diff
@@ -0,0 +1,104 @@
...
+     */
+    protected static $defaultName = 'mix:run';
+
+    /**
+     * @var string The name and signature of this command.
+     */
+    protected $signature = 'mix:run
+        {package : Defines the package to watch for changes}
+        {script : The name of the script to run, as defined in the package.json "scripts" config.}
+        {additionalArgs?* : Arguments to pass through to the }
```
LukeTowers: to the?

modules/system/console/MixRun.php
bennothommo: Yeah, seems to fail in Actions currently, but works locally. :(

modules/system/tests/js/README.md
bennothommo: It's not, it's deleted. No longer accurate with this new system

.github/workflows/tests.yml
```diff
@@ -3,34 +3,96 @@ name: Tests
 on:
   push:
     branches:
-      - 1.0
-      - 1.1
-      - 1.2
+      - '1.2'
```
LukeTowers, line 6: Are we no longer testing on the other two branches?

bennothommo: The other two branches will use the [old `tests.yml` config](https://github.com/wintercms/winter/blob/1.1/.github/workflows/tests.yml), which runs all the tests within the `tests` folder.

I've split off this config to run only 1.2 and develop as we moved all the tests into the System module.

LukeTowers (approved):
LGTM! Can you add a PR to the docs for this as well?

PR #710, opened 2022-09-25
Fix inspector depends functionality.

mjauvin:
This PR corrects the following problems:

`object` inspector type: properties using `depends` do not refresh.

`objectList` inspector type: `itemProperties` using `depends` do not work.

### DIFF
--- a/modules/system/assets/ui/js/inspector.editor.dropdown.js
+++ b/modules/system/assets/ui/js/inspector.editor.dropdown.js
@@ -313,6 +313,11 @@
             $form = $(this.getSelect()).closest('form'),
             dependents = this.inspector.findDependentProperties(this.propertyDefinition.property)
 
+        if (this.inspector.options.parentContainer) {
+            // get parent container data values instead (ref. objectlist editor)
+            data = this.inspector.options.parentContainer.getValues()
+        }
+
         if (currentValue === undefined) {
             currentValue = this.propertyDefinition.default
         }
@@ -382,7 +387,7 @@
 
         for (var i = 0, len = this.propertyDefinition.depends.length; i < len; i++) {
             var property = this.propertyDefinition.depends[i],
-                value = this.inspector.getPropertyValue(property)
+                value = this.getRootSurface().getPropertyValue(property)
 
             if (value === undefined) {
                 value = '';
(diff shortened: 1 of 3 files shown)


bennothommo:
@mjauvin could you add a test scenario to the Test plugin that demonstrates this?

mjauvin:
@bennothommo I'll try, but I don't garanty anything

mjauvin:
@bennothommo I'm frankly not sure how to approach this. The tests would need to create a inspector widget then simulate trigger events on the dropdown fields... any idea?

I could propably instanciate such a widget with `cms/formwidgets/Components.php`, but how do I programatically interact with it?

LukeTowers:
@mjauvin I assume you're using this functionality in a component in a plugin? We don't need an automated test case for this at the moment, just a reproduction of what the use case is that triggers the original issue so we can test the before / after of the fix. Can be implemented as a component within the test plugin I'd imagine.

mjauvin:
yeah, I think this is a better approach.

mjauvin:
@LukeTowers @bennothommo this commit (https://github.com/wintercms/wn-test-plugin/pull/10/commits/5e93893721059448ca21900319747e337dfc3e61) adds a component to the Test plugin that can be added to a page/layout to demonstrate the problem.

- When selecting a value from the `Reference Dropdown` property, this should update the items in the `Object Test` dropdown property. Currently, this only works after closing/re-opening the Inspector.

- Also, If you open the `ObjectList Test` property and click Add, its dropdown should be populated with values related to the value selected in the `Reference Dropdown` property. This doesn't happen (even after a close/re-open of the Inspector)

This PR fixes both of the above issues.

bennothommo:
I'll have to give this a test in a couple of days - launching that major project I've been working on in the next day or two, so gotta focus on that going smooth :)

mjauvin:
> I'll have to give this a test in a couple of days - launching that major project I've been working on in the next day or two, so gotta focus on that going smooth :)

Did you get time to test this?

bennothommo:
I haven't, no. Thanks for the reminder.

mjauvin:
Is that going to be superseded by the inspector widget rewrite ?

bennothommo:
Most likely but the Inspector rewrite is a ways off still - if this works, I'm happy to merge it. It's all tested, I assume?

EDIT: I realise I was supposed to test it - d'oh! I'll get to that today.

mjauvin:
@bennothommo Yes, it's been tested on my end and there's a procedure/PR to test this with the test plugin above.

PR #725, opened 2022-10-05
Add support for icon picker

robertalexa:
Add support for an icon picker field. This took inspiration from https://github.com/hasinhayder/vue3-icon-picker but was modified greatly in the end.

image
image

Offers support for extending the fonts library dynamically.

A few notes:
- json will be added to the page multiple time if more than 1 icon picker is included.
- there are significant issues with how we handle font awesome 6, 4 (-o), wn- and icon-. We seems to be on FA 6.1.1 with the latest being 6.2.0. An update to this would be required, and maybe at this point re-work how we provide support for all the legacy, so that is also becomes easier to update and maintain.
- because of the above, and because i do not know what FA was used in the forked repo, I am not sure if the icon picker as it stands in this PR covers all the icons in 6.1.1. This should be improved on upon updating FA


bennothommo:
@robertalexa this looks awesome. Do you have any screenshots on how this looks in the current backend theme?

As for upgrading Font Awesome, our current fork is using FA 6.1.1 as I [made the upgrade](https://github.com/wintercms/winter/commit/bd3401f6de6f57cf9f83f71ee6c1b708c88d6cef) in May. There was a manual element to it - I tried to use the LESS edition of the Font Awesome source files, but I believe I had to make substantial edits for us to cover all our legacy bases. If we need to upgrade again, I can have another crack at doing it and either make it more automated or at least document what I did.

modules/backend/formwidgets/IconPicker.php
```diff
@@ -0,0 +1,1635 @@
...
+ * @author Robert Alexa, Jack Wilkinson
+ */
+class IconPicker extends FormWidgetBase
+{
+    /**
+     * @inheritDoc
+     */
+    protected $defaultAlias = 'iconpicker';
+
+    protected array $fontLibraries = [
```
LukeTowers: These should be defined in a separate YAML file / files and be configurable by the user

modules/backend/formwidgets/iconpicker/assets/src/app.js
```diff
@@ -0,0 +1,12 @@
+import { createApp } from 'vue'
+import App from './components/App.vue'
+
+window.addEventListener('load', () => {
+    const elements = document.querySelectorAll("[data-control='iconpicker']:not([data-v-app])");
+    elements.forEach(element => {
+        createApp(App, {
+            ...element.dataset,
+            fontLibraries: JSON.parse(element.parentElement.querySelector("script[type='application/json']").textContent)
```
LukeTowers: Font libraries can be loaded over AJAX to avoid large quantities of data being duplicated unnecessarily on page load

modules/backend/formwidgets/iconpicker/partials/_iconpicker.php
```diff
@@ -0,0 +1,17 @@
...
+         data-control="iconpicker"
+         data-label="Icon Picker"
+         data-prop-value="<?= e($field->getLoadValue()) ?>"
+         data-name="<?= $field->getFieldName() ?>"
+    ></div>
+    <script type="application/json"><?= json_encode($fontLibraries) ?></script>
+<?php endif ?>
+
+
+
```
LukeTowers:
```suggestion
```

LukeTowers (changes requested):

modules/backend/formwidgets/IconPicker.php
robertalexa: How I envisioned this is the following:
- WinterCMS provides a default list of icons based on all available FontAwesome icons at the current version.
- The end user is able to manipulate this default as desired by hooking into the `backend.formwidgets.iconpicker.fontLibraries` event.
- The default list of icons should be moved out of where it currently is. Instead, Font Awesome should be updated to the last known version, all shims be created, and this "library" be generated there and then based on these available icons. This generated library should be included in the iconpicker instead of a hardcoded list. This way, at the next FA update you simply regenerate the "library" and job is good.

In the meantime I have added a comment above the current location.

robertalexa:
@bennothommo The picker looks the same on both default and tailwindui theme.

I had an attempt and thought that it would be great to instead use a variable (the branding colours that the user selects) but I have discovered inconsistencies between the use of those brandings between the default and tailwindui theme. So i have left this for now.

The light blue looks like it belongs on both UIs
image

modules/backend/formwidgets/IconPicker.php
robertalexa: Further changes:
- moved it into a new config file
- retired the event as people can just overload the config by doing their own one in their config folder

jaxwilko:
@LukeTowers any news on this?

LukeTowers:
@jaxwilko it needs to be loaded from a YAML file like a nestedform form config instead of a config file. I can do that sometime this week if @robertalexa can't.

robertalexa:
@LukeTowers What difference does that make if neither of these options is the real good long term one? The real good one is to actually handle this "library" of icons as part of the update process of fontawesome at `winter`'s end and decomissions the hardcoded config/yaml approach.

But to do it the right way now, it requires font awesome updated, the shims done once, in 1 place, and married with an existing FA, with regular, solid and brands categories in mind, and with the update process defined and documented.

LukeTowers:
Regardless of the long term improvements to the icon updating process for the Winter core the API for this particular form widget will remain the same:

```yaml
icon:
    label: Icon
    type: iconpicker
    libraries: 
        fontawesome: ...
        materialui: ...
    # or
    libraries: ~/modules/backend/formwidgets/iconpicker/meta/libraries.yaml
```

It will not be a config file at any point.

Another change that needs to be made is renaming `app.js` to `iconpicker.js` and `App.vue` to `IconPicker.vue` and `let app = createApp(App)` to `let iconPicker = createApp(IconPicker)`

robertalexa:
```yaml
form:
    fields:
        icon:
            label: 'Icon'
            type: 'iconpicker'
            default: 'far icon-angry'
            libraries:
                -
                    id: 'far'
                    title: "Font Awesome Regular"
                    prefix: "far icon-"
                    listicon: "far icon-circle"
                    icons:
                        - "far icon-address-book"
                        - "far icon-address-card"
        icon2:
            label: 'Icon'
            type: 'iconpicker'
            default: 'far icon-angry'
            libraries: ~/modules/backend/formwidgets/iconpicker/meta/libraries.yaml
        icon3:
            label: 'Icon'
            type: 'iconpicker'
```

Use the above to test. These are the current variations supported using the approach suggested by @LukeTowers 

> NOTE: the libraries are not scoped by a library parent. They share equal weight. In the context of FontAwesome; `far`, `fas` and `fab` are "individual" libraries

The above example shows:
- manual library definition
- file based library
- no library defined - it will then default to the default path in `\Backend\FormWidgets\IconPicker::DEFAULT_LIBRARIES`
- no default icon defined - it will then default to 1st available icon
- default icon defined but not available in the library - defaults to 1st available icon

In the grand scheme of things, the libraries.yaml file should be rebuild as part of FA update process. I can't guarantee that the current list contains all the icons available in the current (winter) FA version, but I can guarantee that everything that is included is 100% supported.

Screeshots in the order of the above example:
image

image

image

image

modules/backend/assets/ui/js/build/backend.js
```diff
@@ -1 +1,280 @@
...
...ZW5kLmFqYXguaGFuZGxlciddKCkuYWRkUHJlZmlsdGVyKCk7XG5cbiAgICAvLyBBZGQgcG9seWZpbGwgZm9yIEFzc2V0TWFuYWdlclxuICAgIHdpbmRvdy5Bc3NldE1hbmFnZXIgPSB7XG4gICAgICAgIGxvYWQ6IChhc3NldHMsIGNhbGxiYWNrKSA9PiB7XG4gICAgICAgICAgICBTbm93Ym9hcmQuYXNzZXRMb2FkZXIoKS5sb2FkKGFzc2V0cykudGhlbihcbiAgICAgICAgICAgICAgICAoKSA9PiB7XG4gICAgICAgICAgICAgICAgICAgIGlmIChjYWxsYmFjayAmJiB0eXBlb2YgY2FsbGJhY2sgPT09ICdmdW5jdGlvbicpIHtcbiAgICAgICAgICAgICAgICAgICAgICAgIGNhbGxiYWNrKCk7XG4gICAgICAgICAgICAgICAgICAgIH1cbiAgICAgICAgICAgICAgICB9LFxuICAgICAgICAgICAgKTtcbiAgICAgICAgfSxcbiAgICB9O1xuICAgIHdpbmRvdy5hc3NldE1hbmFnZXIgPSB3aW5kb3cuQXNzZXRNYW5hZ2VyO1xufSkod2luZG93LlNub3dib2FyZCk7XG4iXSwibmFtZXMiOlsiSGFuZGxlciIsInJlYWR5IiwiYWpheEZldGNoT3B0aW9ucyIsImFqYXhVcGRhdGVDb21wbGV0ZSIsIndpbmRvdyIsImpRdWVyeSIsImRvY3VtZW50IiwidHJpZ2dlciIsImFqYXhQcmVmaWx0ZXIiLCJvcHRpb25zIiwiaGFzVG9rZW4iLCJoZWFkZXJzIiwiZ2V0VG9rZW4iLCJ0b2tlbkVsZW1lbnQiLCJxdWVyeVNlbGVjdG9yIiwiaGFzQXR0cmlidXRlIiwiZ2V0QXR0cmlidXRlIiwiU25vd2JvYXJkIiwiU2luZ2xldG9uIiwiQmFja2VuZEFqYXhIYW5kbGVyIiwidW5kZWZpbmVkIiwiRXJyb3IiLCJhZGRQbHVnaW4iLCJhZGRQcmVmaWx0ZXIiLCJBc3NldE1hbmFnZXIiLCJsb2FkIiwiYXNzZXRzIiwiY2FsbGJhY2siLCJhc3NldExvYWRlciIsInRoZW4iLCJhc3NldE1hbmFnZXIiXSwic291cmNlUm9vdCI6IiJ9
```
LukeTowers: All asset files need to be compiled with production mode

modules/backend/formwidgets/IconPicker.php
```diff
@@ -0,0 +1,67 @@
+<?php namespace Backend\FormWidgets;
+
+use Backend\Classes\FormWidgetBase;
+use Config;
+use File;
+use Url;
+use Winter\Storm\Exception\ApplicationException;
```
LukeTowers:
```suggestion
```

modules/backend/formwidgets/IconPicker.php
```diff
@@ -0,0 +1,67 @@
+<?php namespace Backend\FormWidgets;
+
+use Backend\Classes\FormWidgetBase;
+use Config;
```
LukeTowers:
```suggestion
```

LukeTowers:
Looking amazing so far @robertalexa!

Just the few items remaining:

- Need to compile the assets for production

> no default icon defined - it will then default to 1st available icon
> default icon defined but not available in the library - defaults to 1st available icon

This widget should at its core act like an enhanced text input. If a value is present that is not defined in the configured library then it should still load, display, and persist that value as if it was a regular text box, but it will just not show any icon preview. With that in mind, the behaviour for those two cases should be as follows:

- No default icon defined and/or no existing value - text input is empty
- Default icon defined and/or existing value but value is not available in the library - text input contains the value, the preview does not display it.

robertalexa:
@LukeTowers I have implemented changes based on your feedback.

```form:
    fields:
        icon:
            label: 'Icon'
            type: 'iconpicker'
            default: 'far icon-angryyy'
            libraries:
                -
                    id: 'far'
                    title: "Font Awesome Regular"
                    prefix: "far icon-"
                    listicon: "far icon-circle"
                    icons:
                        - "far icon-address-book"
                        - "far icon-address-card"
        icon2:
            label: 'Icon'
            type: 'iconpicker'
            default: 'far icon-angry'
            libraries: ~/modules/backend/formwidgets/iconpicker/meta/libraries.yaml
        icon3:
            label: 'Icon'
            type: 'iconpicker'
```
image

First icon picker open (no value, set default, unavailable in the library):
image

Second icon picker open (no value, set default, available in the library)
image

Third icon picker open (no value, no set default, first icon in library)
image

LukeTowers (approved):

LukeTowers:
Awesome work @robertalexa, thanks! Could you put together a PR to the docs to document it?

PR #726, opened 2022-10-06
Add support for Child Themes

### LINKED ISSUE #22: Add support for child themes and nested layouts
Refs octobercms/october#2886 & https://github.com/octobercms/docs/commit/42d7837d7d755e5eeeb1db644e60b351adbac46e

### LINKED ISSUE #153: Possibility to include partials outside of the theme
I'm currently running an older version that allows including files outside of the theme, but going over newer commits I see that it was patched as a security bug that stops you from going outside your theme to find files.

Usually I would have one theme that is common and than all my other themes would load partials from it
Which could include forms, content, or simple html elements that are exactly the same on ~20 websites.  

Out of all partials, I only have one that has php in it via `onStart()`. 

To test it, you can create new empty theme and try to load a partial from a demo theme

```
{% partial '../../demo/partials/calcresult' %}
```

I have also tried using `~/themes/`, etc

I do understand security concern, but would be good if there was a way to disable it for the whole project.

One of the "workaround" I had in mind is to create a relative symlink in each theme, but was curious if there a better way to do it.

jaxwilko:
This PR adds support for a theme to extend another theme via a `parent: [themeName]` directive in the `theme.yaml`. This PR also supports the `theme.yaml` being added as a `DbDatasource` content meaning that an extending theme can be purely driven by the DB.

Storm PR: https://github.com/wintercms/storm/pull/123


mjauvin:
@jaxwilko Do we want this PR for 1.2.1 since it's still a draft?

jaxwilko:
@mjauvin still got some bugs that need to be fixed before it's ready for review :)

mjauvin:
So it's mostly done, just has a few glitches?

LukeTowers:
@mjauvin this functionality is needed for our project that we're launching in November so what release it's tagged for doesn't really matter since it'll be merged when it's ready. Don't consider it blocking against 1.2.1 though.

modules/cms/classes/Theme.php
```diff
@@ -402,25 +420,63 @@ public function getFormConfig()
         return $config;
     }
 
+    /**
+     * Generates an asset URL for the provided path within the theme, will use the parent theme
+     * if the current theme does not actually have a directory on the filesystem (i.e. is virtual).
+     */
+    public function assetUrl(?string $path): string
+    {
+        $cacheTime = Config::get('cms.urlCacheTtl', 10);
```
LukeTowers: Cache expiration is set in seconds, not minutes, best practice is to always do `now()->addMinutes(10)` instead for most explicit use

jaxwilko: We should probably update this comment then
https://github.com/wintercms/winter/blob/ffa6f99210234f86891def675df69341d30f925c/config/cms.php#L172

And change this value
https://github.com/wintercms/winter/blob/ffa6f99210234f86891def675df69341d30f925c/config/cms.php#L176

jaxwilko: Added this in [1a8fb44](https://github.com/wintercms/winter/pull/726/commits/1a8fb449a9f9989f19ce986b6e442c29cac03ece)

modules/cms/classes/Controller.php
```diff
@@ -1353,22 +1355,51 @@ public function currentPageUrl($parameters = [], $routePersistence = true)
...
      */
-    public function themeUrl($url = null)
+    public function themeUrl($url = null): string
+    {
+        return is_array($url)
+            ? $this->themeUrlArray($url)
+            : $this->getTheme()->assetUrl($url);
+    }
+
+    protected function themeUrlArray(array $url): string
```
```suggestion
    public function themeUrl($url = null): string
    {
        return is_array($url)
            ? $this->themeCombineAssets($url)
            : $this->getTheme()->assetUrl($url);
    }

    /**
     * Generates a URL to the AssetCombiner for the provided array of assets
     */
    protected function themeCombineAssets(array $url): string
```

modules/cms/classes/Controller.php
```diff
@@ -1353,22 +1355,54 @@ public function currentPageUrl($parameters = [], $routePersistence = true)
...
+                themes_path($themeDir)
+            ];
+
+            if ($parentTheme) {
+                $sources[] = themes_path($parentTheme);
             }
-            $_url = Url::asset($_url);
+
+            foreach ($url as $file) {
+                if (str_starts_with($file, '@')) {
```
LukeTowers, line 1387:
```suggestion
                // Leave Combiner Aliases assets unmodified
                if (str_starts_with($file, '@')) {
```

modules/cms/classes/Controller.php
```diff
@@ -1353,22 +1355,55 @@ public function currentPageUrl($parameters = [], $routePersistence = true)
...
+
+                foreach ($sources as $source) {
+                    $asset = $source . DIRECTORY_SEPARATOR . $file;
+                    if (File::exists($asset)) {
+                        $assets[] = $asset;
+                        break 2;
+                    }
+                }
+
+                throw new \Exception('Theme URL File not found: ' . $file);
```
```suggestion
                // Skip combining missing assets and log an error
                Log::error("$file could not be found in any of the theme's sources (" . implode(', ', $sources) . ',');
                continue;
```
also test if correct syntax and import the log class

modules/cms/widgets/AssetList.php
```diff
@@ -405,6 +405,11 @@ protected function getData()
     {
         $assetsPath = $this->getAssetsPath();
 
+        // theme dir does not exist
```
LukeTowers:
```suggestion
        // theme dir does not exist (i.e. in a child theme without an assets directory
```

modules/cms/classes/Theme.php
```diff
@@ -192,43 +167,80 @@ public static function getActiveThemeCode()
...
         }
 
         $theme = static::load(static::getActiveThemeCode());
 
-        if (!File::isDirectory($theme->getPath())) {
-            return self::$activeThemeCache = null;
-        }
+//        if (!File::isDirectory($theme->getPath())) {
+//            return self::$activeThemeCache = null;
+//        }
```
LukeTowers:
```suggestion
```
Allow virtual themes to exist

modules/cms/classes/Theme.php
```diff
@@ -290,28 +300,26 @@ public static function getEditThemeCode()
...
         }
 
         $theme = static::load(static::getEditThemeCode());
 
-        if (!File::isDirectory($theme->getPath())) {
-            return self::$editThemeCache = null;
-        }
+//        if (!File::isDirectory($theme->getPath())) {
+//            return self::$editThemeCache = null;
+//        }
```
LukeTowers:
```suggestion
```
Allow virtual themes

modules/cms/classes/Theme.php
```diff
@@ -332,20 +340,28 @@ public static function all()
...
-        if (!File::exists($path)) {
+        $sources = [
+            'filesystem' => new FileDatasource(themes_path($this->getDirName()), App::make('files'))
+        ];
+
+        if (static::databaseLayerEnabled()) {
+            $sources['database'] = new DbDatasource($this->getDirName(), 'cms_theme_templates');
+        }
+
+        $data = (new AutoDatasource($sources))->selectOne('', 'theme', 'yaml');
```
LukeTowers:
```suggestion
        // Attempt to load the theme's config file from whatever datasources are available.
        $sources = [
            'filesystem' => new FileDatasource(themes_path($this->getDirName()), App::make('files'))
        ];
        if (static::databaseLayerEnabled()) {
            $sources['database'] = new DbDatasource($this->getDirName(), 'cms_theme_templates');
        }
        $data = (new AutoDatasource($sources))->selectOne('', 'theme', 'yaml');
```

modules/cms/classes/Theme.php
```diff
@@ -402,25 +417,63 @@ public function getFormConfig()
...
 
+    /**
+     * Generates an asset URL for the provided path within the theme, will use the parent theme
+     * if the current theme does not actually have a directory on the filesystem (i.e. is virtual).
+     */
+    public function assetUrl(?string $path): string
+    {
+        $cacheTime = now()->addMinutes(Config::get('cms.urlCacheTtl', 10));
+
+        return Cache::remember('winter.cms.assetUrl.' . $path, $cacheTime, function () use ($path) {
```
LukeTowers:
```suggestion
        $expiresAt = now()->addMinutes(Config::get('cms.urlCacheTtl', 10));
        return Cache::remember('winter.cms.assetUrl.' . $path, $expiresAt, function () use ($path) {
```

LukeTowers: Also please refactor this to use the same approach as the themeUrl() from Cms\Classes\Controller with the foreach over the sources.

modules/cms/classes/Theme.php
```diff
@@ -471,55 +523,52 @@ public function writeConfig($values = [], $overwrite = false)
...
      */
-    public static function resetCache()
+    public static function resetCache(bool $memoryOnly = false): void
     {
         self::$activeThemeCache = false;
         self::$editThemeCache = false;
 
-        Cache::forget(self::ACTIVE_KEY);
-        Cache::forget(self::EDIT_KEY);
+        if (!$memoryOnly) {
```
LukeTowers, line 542:
```suggestion
        // Sometimes it may be desired to only clear the local cache of the active / edit themes instead of the persistent cache
        if (!$memoryOnly) {
```

modules/cms/classes/Theme.php
```diff
@@ -528,58 +577,84 @@ public function removeCustomData()
...
-            } else {
-                $datasource = new FileDatasource($this->getPath(), App::make('files'));
+                $sources['parent-database'] = new DbDatasource($config['parent'], 'cms_theme_templates');
             }
 
-            $resolver->addDatasource($this->dirName, $datasource);
+            $sources['parent-filesystem'] = new FileDatasource(themes_path($config['parent']), App::make('files'));
         }
+
+        $resolver->addDatasource($this->dirName, new AutoDatasource($sources, 'halcyon-datasource-auto-' . $this->dirName));
```
LukeTowers:
```suggestion

        $datasource = count($sources) > 1
            ? new AutoDatasource($sources, 'halcyon-datasource-auto-' . $this->dirName)
            : array_shift($sources);
        $resolver->addDatasource($this->dirName, $datasource);
```

LukeTowers (changes requested):

LukeTowers (approved):

PR #730, opened 2022-10-14
Update `failed_jobs` table to add missing `uuid` column.

### LINKED ISSUE #696: Missing uuid field on failed_jobs table
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

When job fail inserting it in the failed_jobs table require an `uuid` field that is not present in the table at the moment, so a pdo exception is thrown

`PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'uuid' in 'field list' in /var/www/html/my.domain.it/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:82`

### Steps to replicate

see Mail queue is not instantiable https://github.com/wintercms/winter/issues/695

### Workaround

Add string field named 'uuid' to the failed_jobs table with a new migration

ericp-mrel:
This also updates the `payload` and `exception` columns to be inline with default the laravel migration.

Fixes #696 

Let me know if you'd prefer I remove the updates to the `payload` and `exception` columns.


LukeTowers:
@ericp-mrel looks good to me, were you able to test it running locally?

ericp-mrel:
@LukeTowers 

I've tested it locally with a blank Winter CMS project, created a job class that fails so it gets added to the `failed_jobs` table and is working correctly. I also tested by importing a database dump from one of my other projects (Winter CMS v1.1) and ran the migrations against that database and the migrations run successfully for existing projects as well.

modules/system/database/migrations/2022_10_17_000028_Db_Jobs_Update_Payload_To_Longtext.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+use Winter\Storm\Database\Schema\Blueprint;
+
+return new class extends \Winter\Storm\Database\Updates\Migration
+{
+    public function up()
+    {
+        Schema::table($this->getTableName(), function (Blueprint $table) {
```
LukeTowers: This can be done in a single migration file, no need to split it into two separate files.

ericp-mrel: @LukeTowers I've merged both changes into one file.

PR #740, opened 2022-10-25
Allow table widget to have a minimum and maximum of items

bennothommo:
This PR adds the config variables `minItems` and `maxItems` for the Table widget and `datatable` field type. The impetus of this is to allow a table to have *no items* by setting the `minItems` to `0` - I have found in my travels that if you have any sort of validation on the table widget, then you must have at least one record, but in my case, I wanted the datatable field to be optional.

I then thought it might be useful to have the opposite, and allow an upper limit for the number of records.

Some rules on the above:
- `minItems` must be `0` or above. By default, it is `1` to match the current functionality. If `minItems` is `2`, then 2 records must be in the datatable at all times (the datatable will have 2 empty records added on load).
- `maxItems` must be either `1` or above, or `null` to indicate no limit.
- `minItems` must be less than or equal to `maxItems`, to allow a datatable to have a fixed number of records.

To allow this change to happen, I've decided to change the datasource objects (client and server-side data handling) to use promises - this way, we have a common way of reading the responses, and it will gracefully handle the server-side taking a little longer to make changes to the data.


modules/backend/widgets/Table.php
```diff
@@ -65,6 +66,16 @@ public function init()
 
         $this->recordsKeyFrom = $this->getConfig('keyFrom', 'id');
 
+        // Validate limits
+        if (!empty($this->getConfig('maxItems')) && (int) $this->getConfig('minItems', 1) > (int) $this->getConfig('maxItems')) {
+            throw new ApplicationException(
```
LukeTowers: This should be a SystemException, ApplicationExceptions are not logged and are meant as validation for the user. In this case the user doesn't control this configuration - the developer does.

modules/backend/widgets/Table.php
```diff
@@ -131,6 +142,8 @@ public function prepareVars()
         $this->vars['deleting'] = $this->getConfig('deleting', true);
         $this->vars['toolbar'] = $this->getConfig('toolbar', true);
         $this->vars['height'] = $this->getConfig('height', false) ?: 'false';
+        $this->vars['minItems'] = (int) $this->getConfig('minItems', 1);
+        $this->vars['maxItems'] = (int) $this->getConfig('maxItems');
```
LukeTowers, line 145: This would convert the null to 0 would it not?

modules/backend/widgets/Table.php
bennothommo: Good point - changed.

modules/backend/widgets/Table.php
bennothommo: That should be fine - zero is considered no limit.

LukeTowers:
@bennothommo what's the status of this?

bennothommo:
I'm gonna say I think it is complete... Definitely haven't tested it quite some time though.

LukeTowers:
@bennothommo do we have a docs PR for it?

bennothommo:
I will test it first, and then I'll put up a docs PR, in case anything needs to change.

LukeTowers:
@bennothommo what's the status on this?

bennothommo:
Same as before. Still needs docs and testing. Haven't had a need for it yet - I think I did this for my old job.

LukeTowers:
@der-On or @WebVPF, either of you bored and looking for something to do 😉?

der-On:
@LukeTowers Bored to the bones :joy_cat: . Sorry I'm currently out on this.

PR #744, opened 2022-10-26
Fix bug when clicking move folder in media manager

quangtrongonline:


modules/backend/widgets/MediaManager.php
```diff
@@ -519,9 +519,9 @@ public function onCreateFolder(): array
     /**
      * Show move item popup AJAX handler
      *
-     * @throws ApplicationException If the exclude input data is not an array
+     * @throws ApplicationException If the exclude input data is not an string
```
LukeTowers:
```suggestion
     * @throws ApplicationException If the exclude input data is not an array
```

PR #745, opened 2022-10-28
Added fix to prevent path duplication

jaxwilko:
Fixed an issue where the asset url path would duplicate the asset file path.


mjauvin:
@LukeTowers @jaxwilko Would that not warrant a test case?

PR #746, opened 2022-10-28
CMS Twig extension `contentFunction()` should return bool like `partialFunction()`

### LINKED ISSUE #731: Test content file exists
### Winter CMS Build

1.2

### PHP Version

8.0

### Database engine

MySQL/MariaDB

### Plugins installed

Pages

### Issue description

Hi,
I would like, in a CMS partial, test if a content file exists.
So I do :
```
                {% if content(<my_file.htm>) %}
                  {% content <my_file.htm> %}
                {% else %}
                  <something else>
                {% endif %}
```
If I do that, I have a twig error on my if line : `An exception has been thrown during the rendering of a template ("The content file '<my_URL>' is not found.").`

How can I do that ?

Thanks

### Steps to replicate

After a fresh install, just add the test

### Workaround

_No response_

mjauvin:
Fixes #731

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -1099,8 +1099,8 @@ class Controller
      * @param string $name The content view to load.
      * @param array $parameters Parameter variables to pass to the view.
      * @param bool $throwException Throw an exception if the content file is not found.
-     * @throws SystemException If the content cannot be found
-     * @return mixed Content file or false if not throwing an exception.
+     * @throws SystemException If the content cannot be found, and `$throwException` is true.
+     * @return string|false Content file, or false if `$throwException` is false.
      */
     public function renderContent($name, $parameters = [], $throwException = true)
     {
@@ -1129,7 +1129,7 @@ class Controller
          */
         elseif (($content = Content::loadCached($this->theme, $name)) === null) {
             if ($throwException) {
-                throw new SystemException(Lang::get('cms::lang.content.not_found_name', ['name'=>$name]));
+                throw new SystemException(Lang::get('cms::lang.content.not_found_name', ['name' => $name]));
             } else {
                 return false;
             }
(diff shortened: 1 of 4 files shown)


modules/cms/classes/Controller.php
```diff
@@ -1098,10 +1098,11 @@ public function renderPartial($name, $parameters = [], $throwException = true)
      *
      * @param string $name The content view to load.
      * @param array $parameters Parameter variables to pass to the view.
+     * @param bool $throwException Throw an exception if the content file is not found.
      * @throws SystemException If the content cannot be found
```
bennothommo:
```suggestion
     * @throws SystemException If the content cannot be found, and `$throwException` is true.
```

modules/cms/classes/Controller.php
```diff
@@ -1098,10 +1098,11 @@ public function renderPartial($name, $parameters = [], $throwException = true)
...
-     * @return string
+     * @return mixed Content file or false if not throwing an exception.
```
bennothommo:
```suggestion
     * @return string|false Content file, or false if `$throwException` is false.
```

modules/cms/classes/Controller.php
```diff
@@ -1127,7 +1128,11 @@ public function renderContent($name, $parameters = [])
          * Load content from theme
          */
         elseif (($content = Content::loadCached($this->theme, $name)) === null) {
-            throw new SystemException(Lang::get('cms::lang.content.not_found_name', ['name'=>$name]));
+            if ($throwException) {
+                throw new SystemException(Lang::get('cms::lang.content.not_found_name', ['name'=>$name]));
```
bennothommo:
```suggestion
                throw new SystemException(Lang::get('cms::lang.content.not_found_name', ['name' => $name]));
```

bennothommo (changes requested):

bennothommo (approved):

PR #755, opened 2022-11-04
Show checkboxlist options when in preview / disabled

jaxwilko:
Previously, the condition `elseif (!$readOnly && count($fieldOptions))` would prevent valid options being rendered when the `$readOnly` var is `true` (`$readOnly = $this->previewMode || $field->readOnly || $field->disabled;`).

This seems to be incorrect, the correct approach should be to show all options but mark them all as disabled.

This changes:

image

To:

image

### DIFF
--- a/modules/backend/widgets/form/partials/_field_checkboxlist.php
+++ b/modules/backend/widgets/form/partials/_field_checkboxlist.php
@@ -41,7 +41,7 @@ $quickselectEnabled = $field->getConfig('quickselect', $isScrollable);
         <?php endforeach ?>
     </div>
 
-<?php elseif (!$readOnly && count($fieldOptions)): ?>
+<?php elseif (count($fieldOptions)): ?>
 
     <div class="field-checkboxlist <?= $isScrollable ? 'is-scrollable' : '' ?>">
         <?php if ($quickselectEnabled): ?>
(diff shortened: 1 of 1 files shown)


bennothommo:
Hmm, might be a bit of a subjective call, but I personally would prefer it to hide options that aren't selected when in preview mode, especially if it's a long checkbox list.

Would your changes also make [this condition block](https://github.com/wintercms/winter/blob/1e0a1ec2a79d212a5c4c9e93a57b8ad4db528111/modules/backend/widgets/form/partials/_field_checkboxlist.php#L9) redundant?

LukeTowers:
@bennothommo our specific use case is showing a preview of the field in a visual formbuilder so we would want all options to be displayed.

bennothommo:
My preference is still the same, but perhaps the preview state can be configurable? (ie. `showOnPreview: all|checked|unchecked`)

mjauvin:
Replaced by #907

mjauvin:
I thought I had made changes to this in #907 but I was wrong.

LukeTowers:
Whats the status on this @jaxwilko @bennothommo?

bennothommo:
As indicated before, I still have the same preference.

jaxwilko:
If nobody is asking for it now maybe we just close it and revisit in the future if there is demand.

From my side of things, I still reckon the most user friendly way to do it would be to show all options but have each input marked as disabled as then people with only preview permission would be able to see all options avaliable even if they could not interact with any of them. Not massively opposed to `showOnPreview`, but it adds more complexity to an already complex system which may not be preferable.

LukeTowers:
I'm fine with merging this and if it becomes a problem in the future that too many options are shown on the preview screen then a `hideWhenReadonly` or similar option can be added at that point.

PR #765, opened 2022-11-11
Prevent previous relations from being unlinked

mjauvin:
When adding a `$belongsToMany` relation with pivot date (without a custom pivot model), the previous pivot records get lost.

This fixes the issue.

The issue was introduced in e9739361bb5aac9ad85865ae6037a86b3b9946fa


LukeTowers (approved):
Good catch and fix, thanks @mjauvin!

PR #771, opened 2022-11-21
Improve multidbconnection support

LukeTowers:
Related to https://github.com/wintercms/storm/pull/132

### DIFF
--- a/modules/backend/ServiceProvider.php
+++ b/modules/backend/ServiceProvider.php
@@ -1,6 +1,5 @@
 <?php namespace Backend;
 
-use App;
 use Backend;
 use BackendMenu;
 use BackendAuth;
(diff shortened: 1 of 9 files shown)


modules/backend/ServiceProvider.php
```diff
@@ -23,12 +22,15 @@ public function register()
         $this->registerConsole();
         $this->registerMailer();
         $this->registerAssetBundles();
-        $this->registerBackendPermissions();
+
+        if (!$this->app->runningUnitTests()) {
+            $this->registerBackendPermissions();
+        }
```
bennothommo, line 28: I think it would be preferable to fix the tests if possible. We don't want to end up with a situation where the tests operate differently to the actual functionality.

modules/backend/behaviors/FormController.php
```diff
@@ -177,7 +177,7 @@ public function initForm($model, $context = null)
          * Detected Relation controller behavior
          */
         if ($this->controller->isClassExtendedWith(\Backend\Behaviors\RelationController::class)) {
-            $this->controller->initRelation($model);
+            $this->controller->initRelation(clone $model);
```
bennothommo, line 180: Just to explore the possibility, could we always operate with a clone in the `RelationController::initRelation()` method? Then this line wouldn't need to change, and that way there's no subtle behavior change when `RelationController` is being used with `FormController` vs. when it's being used on its own.

LukeTowers: My personal preference would be to keep it as it is, if someone wants similar isolation they can initialize it the same way on their own, but if we do the clone in the relationcontroller it makes it impossible for them to choose to have the old behaviour.

modules/backend/ServiceProvider.php
LukeTowers: The tests already operated differently from the core given that this call used to be wrapped in the runningInBackend() check instead.

Issue #19, opened 2021-03-12
Add support for mode option to the MediaFinder


LukeTowers:
Refs: octobercms/october#2726

LukeTowers:
@jaxwilko or @robertalexa, whichever is assigned the bugherd issue for selecting images in the mediafinder, please refer to this comment: https://github.com/octobercms/october/issues/2726#issuecomment-319555427

PR #780, opened 2022-11-30
Add Mediafinder Modes

jaxwilko:
This PR introduces support for all mediamanager filters as modes for the mediafinder formwidget.

This means that the following `mode` values would be available:
- image
- file
- document
- audio
- video
- everything (default)

### DIFF
--- a/modules/backend/formwidgets/mediafinder/assets/js/mediafinder.js
+++ b/modules/backend/formwidgets/mediafinder/assets/js/mediafinder.js
@@ -42,7 +42,7 @@
         }
 
         if (this.options.mode === null) {
-            this.options.mode = this.$el.data("mediafinder-mode") || "all";
+            this.options.mode = this.$el.data('mediafinder-mode') || 'all';
         }
 
         this.$el.one('dispose-control', this.proxy(this.dispose))
--- a/modules/backend/formwidgets/mediafinder/assets/js/mediafinder.js
+++ b/modules/backend/formwidgets/mediafinder/assets/js/mediafinder.js
@@ -99,7 +99,7 @@
 
         new $.wn.mediaManager.popup({
             alias: 'ocmediamanager',
-            cropAndInsertButton: ["image", "all"].indexOf(self.options.mode) > -1,
+            cropAndInsertButton: ['image', 'all'].includes(self.options.mode),
             mode: self.options.mode,
             onInsert: function(items) {
                 if (!items.length) {
--- a/modules/backend/formwidgets/mediafinder/assets/js/mediafinder.js
+++ b/modules/backend/formwidgets/mediafinder/assets/js/mediafinder.js
@@ -41,6 +41,10 @@
             this.options.isImage = this.$el.hasClass('is-image')
         }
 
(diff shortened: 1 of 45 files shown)


LukeTowers:
See https://github.com/wintercms/winter/issues/19#issuecomment-1312826732 & https://github.com/octobercms/october/issues/2726#issuecomment-319555427.

@jaxwilko @bennothommo any thoughts on `all` vs `everything` for the mode that shows everything?

jaxwilko:
@LukeTowers I literally had it as `all` everywhere, then checked the php const and change it to `everything`. Yeah I'm okay with that

LukeTowers:
Sigh, we should probably leave it as everything then, unless we update all of the references (which includes lang keys) to FILTER_ALL instead. Perhaps if we also introduced FILTER_FILE, FILTER_FOLDER we could justify switching it to FILTER_ALL?

bennothommo:
I'd prefer `all`, it's more inline with the terminology we use elsewhere.

jaxwilko:
@LukeTowers @bennothommo the filter const and value have now been updated to reference `all` instead of `everything`

LukeTowers:
@jaxwilko is there a PR for the docs?

modules/backend/formwidgets/mediafinder/assets/js/mediafinder.js
```diff
@@ -41,6 +41,10 @@
             this.options.isImage = this.$el.hasClass('is-image')
         }
 
+        if (this.options.mode === null) {
+            this.options.mode = this.$el.data("mediafinder-mode") || "all";
```
bennothommo: Our JS style guide says single quotes for strings.
```suggestion
            this.options.mode = this.$el.data('mediafinder-mode') || 'all';
```

modules/backend/formwidgets/mediafinder/assets/js/mediafinder.js
```diff
@@ -95,7 +99,8 @@
 
         new $.wn.mediaManager.popup({
             alias: 'ocmediamanager',
-            cropAndInsertButton: true,
+            cropAndInsertButton: ["image", "all"].indexOf(self.options.mode) > -1,
```
bennothommo: We can use ES6 code since we don't support IE11 anymore ;)
```suggestion
            cropAndInsertButton: ['image', 'all'].includes(self.options.mode),
```

bennothommo (changes requested):

modules/backend/widgets/MediaManager.php
```diff
@@ -32,7 +32,7 @@ class MediaManager extends WidgetBase
     const SELECTION_MODE_FIXED_RATIO = 'fixed-ratio';
     const SELECTION_MODE_FIXED_SIZE = 'fixed-size';
 
-    const FILTER_EVERYTHING = 'everything';
```
jaxwilko, line 35: @LukeTowers not sure if anybody is using this but it might be a breaking change

jaxwilko:
Docs PR: https://github.com/wintercms/docs/pull/129

PR #782, opened 2022-12-01
Added alias to list of configs to transfer from ListController to List widget

jaxwilko:
This allows you to specify a list widget alias when defining the config for the list controller.

In my use case it would be useful as I'm using one list with different columns dependant upon scopes, and this change allows me to modify the alias to include the scope key, meaning that the user preferences stored against the alias are unique to each scope.

Very open to thoughts on this.


jaxwilko:
The list alias widget can be modified by providing a key to the list config such as:
```php
$this->listConfig = [
    'aliasName' => [
        // ....
    ]
];
```
As this is possible already, it makes the change in this PR redundent

PR #785, opened 2022-12-02
Trigger a change event on repeater item remove

jaxwilko:
Currently when an item is removed from a repeater widget, no change event is triggered, this means that anything watching for changes is not informed of item removals. This PR implements a change trigger on the repeater when an item is removed.


LukeTowers:
Does this send the updated repeater data to the server when another field dependsOn the repeater and the remove is triggered?

bennothommo:
Ping @jaxwilko - any chance you tested @LukeTowers [scenario](https://github.com/wintercms/winter/pull/785#issuecomment-1336125288)?

PR #790, opened 2022-12-08
Allow removeField() on contextual field names

damsfx:
On a field defined with context:
```yaml
dates@update:
    label: Other dates
   ...
```
Then in `backend.form.extendFields` event, the `$widget->removeField('dates')` method doesn't work.  

This PR allow `removeField()` method to work on contextualized fields name.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -736,13 +736,13 @@ class Form extends WidgetBase
 
             switch (strtolower($addToArea)) {
                 case FormTabs::SECTION_PRIMARY:
-                    $this->allTabs->primary->addField($name, $fieldObj, $fieldTab);
+                    $this->allTabs->primary->addField($fieldObj->fieldName, $fieldObj, $fieldTab);
                     break;
                 case FormTabs::SECTION_SECONDARY:
-                    $this->allTabs->secondary->addField($name, $fieldObj, $fieldTab);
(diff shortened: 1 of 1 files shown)


mjauvin:
To add more context for this PR, the following code snippet shows the actual issue:
https://github.com/wintercms/winter/blob/develop/modules/backend/widgets/Form.php#L735-L747

Let's take a field named `dates@update` shown in the PR description above:

The `$fieldObj` was assigned to the $allFields array using the context-less field name (dates):
```
$this->allFields[$fieldObj->fieldName] = $fieldObj;
```

But `$fieldObj` was assigned to the different tabs using the full name (dates@update):
```
$this->allTabs->primary->addField($name, $fieldObj, $fieldTab);
```

That creates a problem here when removing fields:
https://github.com/wintercms/winter/blob/develop/modules/backend/widgets/Form.php#L779-L781

since $allFields array will have a different key (dates) for the field than the tabs objects will (dates@update)

So the field does not get removed if `removeField()` gets called with "dates@update" since `$allFields` doesn't have a key for this name, and if "dates" is used as name, the tabs don't have fields of that name (since they have dates@update).

bennothommo (approved):
Looks good to me!

bennothommo:
If you could add an example to the Test plugin, that would be super awesome. :)

damsfx:
> If you could add an example to the Test plugin, that would be super awesome. :)

@bennothommo Example of `removeField()` method?  
In a dedicated controller?

PR #791, opened 2022-12-08
[FIX] Added fix to allow nested AutoDatasource instances to manage their own populateCache

jaxwilko:
This PR adds a fix to support nested `AutoDatasource` instances. The issue with nesting an `AutoDatasource` inside an `AutoDatasource` is that the `populateCache()` method would not proxy the call to the nested instance and instead force the caching against the `AutoDatasource` instead of the child Datasource instances, this would mean that cache would not be correctly invalidated and after issuing `insert`, `update` or `delete` requests the parent cache would be updated but the instance cache would not be. 

This change allows the `populateCache()` request to be passed into the nested `AutoDatasource` to allow for recursive behaviour.


LukeTowers:
@jaxwilko the tests are more relevant now, can you fix it?

jaxwilko:
> @jaxwilko the tests are more relevant now, can you fix it?

Yeah, so the issue is not caused by this PR, but just becomes apparent because of it. The fix for the issue is here: https://github.com/wintercms/storm/pull/135

PR #802, opened 2022-12-15
[FIX] iconpicker not being registered after load

jaxwilko:
This PR changes how the icon picker is inited to ensure it can be used after page load, i.e. on static pages. There is also a fix here to provide the elements form to the snowboard request if available

### DIFF
--- a/modules/backend/formwidgets/iconpicker/assets/src/iconpicker.js
+++ b/modules/backend/formwidgets/iconpicker/assets/src/iconpicker.js
@@ -1,20 +1,28 @@
 import { createApp } from 'vue'
 import IconPicker from './components/IconPicker.vue'
 
-window.addEventListener('load', () => {
-    const elements = document.querySelectorAll("[data-control='iconpicker']:not([data-v-app])");
-    elements.forEach(element => {
-        element.querySelector(".input-group").addEventListener("click", () => {
-            snowboard.request(null, element.dataset.alias + "::onLoadIconLibrary", {
-                success: (data) => {
-                    let iconPicker = createApp(IconPicker, {
-                        ...element.dataset,
-                        fontLibraries: JSON.parse(data.result)
-                    }).mount(element);
+(() => {
+    const init = () => {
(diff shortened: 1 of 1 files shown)


jaxwilko:
@bennothommo any thoughts on this?

LukeTowers:
> i.e., if it's in "complete" state), then call init() immediately instead of waiting for load events to fire later on down the line when they might not be needed anymore because everything has already been initialized by that point anyway and we don't want any unnecessary work being done at all times just in case something else happens later on down the line after this initial loading process completes).

Good grief @what-the-diff, I know I sometimes have long sentences but that one is just ridiculous

LukeTowers:
@jaxwilko does this help with loading it on partial page refresh?

jaxwilko:
@LukeTowers that's the idea, i found some more issues that i'm now working on though, i'll move it out of draft when i get those fixed

bennothommo:
@jaxwilko @LukeTowers I've written a [Widget Handler](https://github.com/wintercms/winter/blob/wip/monaco-editor/modules/backend/assets/ui/js/ui/WidgetHandler.js) class in with the Monaco Editor changes, which should streamline attaching widgets on load or when a partial update occurs.

See also https://github.com/wintercms/winter/blob/wip/monaco-editor/modules/backend/formwidgets/codeeditor/assets/js/index.js#L580-L581

It tracks elements using the `data-control` attribute, which all our widgets do, and initialises the given widget with that `data-control` value.

I also plan to handle disposing widgets through the same class.

bennothommo:
This should be fixed by the following:

https://github.com/wintercms/winter/commit/2dff8adf25701b6f3def34a005297a7bea07625d - adds the Widget Handler, refactors the Icon Picker to use this to initialise, and moves the embedded Vue library (which was making the Icon Picker JS super large) into a shared vendor build so that we can use Vue for other widgets.
https://github.com/wintercms/winter/commit/64e1e31fb8b6654b593bc0a22fd2f8303576c317 - adds a Mutation observer to watch when elements are removed from DOM. If those elements contain widgets, the widget will go through the destructing / disposing process.

This should make things a lot more graceful and performant.

PR #806, opened 2022-12-17
[Fix] TagList type error

Denoder:
This is causing an error when using relation mode and you have an empty relation and try to save. the method expects an array, but is null so we either put the null value in an array or we can change the type to a nullable type hint. I'm fine with either and will change accordingly.


LukeTowers (approved):

LukeTowers:
Thanks @Teranode, this had been popping up on a client site from time to time for a while and I hadn't gotten around to diving in to the problem so I appreciate the PR 😄

PR #835, opened 2023-02-08
Update composer.json

hecc127:
Minor fix, change "merge-plugin"  configuration


LukeTowers:
It is the way it is right now on purpose, including everything by default causes issues with plugins that are already pulled in via composer. You should be using this section to include only the specific plugins in your project that you include as a part of your project but have their own composer.json files.

PR #836, opened 2023-02-10
Add support for extending head block on dashboard

jaxwilko:
By default, `Block::put('key')` & `Block::endPut()` defaults to overwriting instead of appending.

This change means calls to `Block::append('head', $content)` will function correctly for the dashboard.


jaxwilko:
> * The endPut() method was changed to accept a boolean parameter.
> * If the value of this parameter is true, then it will render the block immediately instead of waiting for all blocks to be rendered at once in Block::end().

Bad bot

PR #841, opened 2023-02-19
Return correct URLs from File->getPublicPath for non-local disks

wpjscc:
when driver is cloud driver, the url  is dependent on ASSET_URL。

it should dependent on cloud ENDPOINT

### DIFF
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -71,7 +71,7 @@ class File extends FileBase
             $uploadsPath .= '/protected';
         }
 
-        return Url::asset($uploadsPath) . '/';
+        return $this->getDisk()->url($uploadsPath) . '/';
     }
 
     /**


bennothommo:
I think there's a specific rationale for why it's like that - @jaxwilko or @LukeTowers might know more.

jaxwilko:
@bennothommo I vaguly remember it being important, not sure why though. In theory this change looks okay but maybe @LukeTowers knows more

LukeTowers:
Hmm, on my phone at the moment but check the git blame. Hopefully there's a decent commit message

bennothommo:
https://github.com/wintercms/winter/commit/5aa351f67c0f9b4d7b1b5c5cb6c2199d19164c79

Not sure about that...

LukeTowers:
This is totally fine then @bennothommo, that commit is from before I added the getDisk() method to the File model in https://github.com/wintercms/storm/commit/48fed9b7e4c4c22ccc6b3c2e5b1273da3e2c38dd, just leftover code from the old non-disk focused way of doing things.

PR #842, opened 2023-02-19
custom uploads and media and resized  filesystem driver

wpjscc:


LukeTowers:
We want to refactor how those disks are configured in general by making them real disk instances instead of pseudo "sub disks", so this PR isn't going to be approved at this time. Feel free to implement it this way in your own Winter projects though, the config files are not modified by composer updates.

PR #844, opened 2023-02-21
Use native disk methods instead of the pseudo scoped disks

LukeTowers:
WIP, requires https://github.com/wintercms/storm/pull/141


LukeTowers:
@jaxwilko this was the WIP PR for breaking out those disks, I don't think it would be too difficult to finish. Perhaps we should also use properties instead of class constants for storing the disks so that we can more easily switch away from the singleton pattern later for the medialibrary etc.

Do you want to take a look and see what's left to push it across the finish line?

jaxwilko:
@LukeTowers Happy to do so but it's going to be middle of next week before i can start :)

PR #849, opened 2023-02-28
Improve French translations

damsfx:


modules/backend/lang/fr/lang.php
```diff
@@ -29,9 +35,24 @@
         ],
     ],
     'partial' => [
-        'not_found_name' => 'Le modèle partiel ":name" est introuvable.'
+        'not_found_name' => 'Le modèle partiel ":name" est introuvable.',
+        'invalid_name' => 'Le nom du modèle partiel est invalide: :name.',
+    ],
+    'ajax_handler' => [
+        'invalid_name' => 'Nom du gestionnaire AJAX non valide: :name.',
+        'not_found' => "Le gestionnaire AJAX ':name' n'a pas été trouvé..",
```
mjauvin:
```suggestion
        'not_found' => "Le gestionnaire AJAX ':name' n'a pas été trouvé.",
```

modules/system/lang/fr/lang.php
```diff
@@ -12,11 +12,6 @@
     'file' => [
         'create_fail' => "Impossible de créer le fichier : :name",
     ],
-    'page' => [
-        'invalid_token' => [
-            'label' => 'La clé de sécurité est invalide',
-        ],
-    ],
```
mjauvin, line 19: Why was this removed?

mjauvin (changes requested):
Please address these.

modules/system/lang/fr/lang.php
damsfx: The `page` exist twice in the file.
https://github.com/wintercms/winter/blob/b1d28d9d3ef3e5889bf93a38dfbcfbabb723f256/modules/system/lang/fr/lang.php#L15-L19
And 
https://github.com/wintercms/winter/blob/b1d28d9d3ef3e5889bf93a38dfbcfbabb723f256/modules/system/lang/fr/lang.php#L410-L423

mjauvin: Yes, but we're now missing the invalid_token key under page...

modules/backend/lang/fr/lang.php
```diff
@@ -29,9 +35,24 @@
         ],
```
mjauvin, line 35:
```suggestion
        ],
        'invalid_token' => [
            'label' => 'La clé de sécurité est invalide',
        ],
```

damsfx, line 35: @mjauvin Why do you want to add this entry?
It doesn't exist in the EN file.

mjauvin:
Sorry @damsfx, I was looking in the wrong file...

mjauvin:
Thanks @damsfx !

PR #853, opened 2023-03-02
fix when useKey is true do not save selected data

wpjscc:
when useKey is true 

the selected is  `$relationModel->getKeyName()` not name

```
 $existingTags = $relationModel
            ->whereIn($this->nameFrom, $names)
            ->lists($this->nameFrom)
        ;
```

`whereIn($this->nameFrom, $names)`  is not get data

change to this when useKey is true
```
 $existingTags = $relationModel
            ->whereIn($this->useKey ?  $relationModel->getKeyName() : $this->nameFrom, $names)
            ->lists($this->nameFrom, $relationModel->getKeyName())
        ;
```

it work for me

### DIFF
--- a/modules/backend/formwidgets/TagList.php
+++ b/modules/backend/formwidgets/TagList.php
@@ -120,6 +120,7 @@ class TagList extends FormWidgetBase
     /**
      * Returns an array suitable for saving against a relation (array of keys).
      * This method also creates non-existent tags.
+     * @return array
      */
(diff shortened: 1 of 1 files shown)


LukeTowers:
@wpjscc could you provide an example of the field definition that you were trying to use that was broken and an explanation of how this fixes it? I'm not sure I fully understand your description as it is right now.

Issue #855, opened 2023-03-02
Problem in Cms\Classes\Theme::getDatasource()


webmaxx:
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

Wintercms.Pages, Wintercms.Sitemap, Wintercms.Notify, Wintercms.Redirect

### Issue description

```php
/**
 * Get the theme's datasource
 */
public function getDatasource(): AutoDatasource
{
    $resolver = App::make('halcyon');
    return $resolver->datasource($this->getDirName());
}
```
Cms\Classes\Theme::getDatasource(): Return value must be of type Cms\Classes\AutoDatasource, Winter\Storm\Halcyon\Datasource\FileDatasource returned

### Steps to replicate

1. Create new theme in console `artisan create:theme example`
2. Activate new theme in backend (and DO NOT change anything in "Customize theme" or click on the button "Reset to default")
3. Go to main page

webmaxx:
Made a video demonstration: https://youtu.be/EHFv70Csqs0

jaxwilko:
Hi @webmaxx could you do me a favour and try https://github.com/wintercms/winter/pull/856 to see if that fixes the issue for you? I couldn't actually get it to fail on my end but that's probably a configuration issue for me

webmaxx:
> Hi @webmaxx could you do me a favour and try #856 to see if that fixes the issue for you? I couldn't actually get it to fail on my end but that's probably a configuration issue for me

Yes. It's working

LukeTowers:
Fixed by #856

PR #856, opened 2023-03-02
Fix typehint

jaxwilko:
This is a fix for https://github.com/wintercms/winter/issues/855

For efficency, if the theme only registers 1 `Datasource` we're now not injecting that `Datasource` into an `AutoDatasource` and instead registering the single `Datasource`.

This breaks the `AutoDatasource` type hinting, updating the return type to reference the `DatasourceInterface` should resolve the issue in all cases.

### DIFF
--- a/modules/cms/classes/Theme.php
+++ b/modules/cms/classes/Theme.php
@@ -666,7 +666,7 @@ class Theme extends CmsObject
     /**
      * Get the theme's datasource
      */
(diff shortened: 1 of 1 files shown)


jaxwilko:
Fix confirmed by @webmaxx https://github.com/wintercms/winter/issues/855#issuecomment-1452022438

LukeTowers:
Fixes #855

PR #858, opened 2023-03-05
Fix Class "Cms\Classes\Log" not found error

xitara:


bennothommo:
@xitara the convention would be to include a `use Log;` call at the top of the file where the other use cases are. Would you mind changing the change to that? I will then merge it.

xitara:
no problem :)

PR #860, opened 2023-03-06
Add traits support to Snowboard

bennothommo:
Implements support for traits in Snowboard, allowing additional inheritance and code re-use. 

Traits can be added to any Snowboard plugin class using the `traits()` method, which must return an array of classes that you wish to use as traits.

```js
class ClassWithTrait extends Snowboard.PluginBase {
    traits() {
        return [
            MyTraitClass
        ];
    }
}
```

A trait class can be a simple JavaScript class with methods and properties.

```js
class MyTraitClass {
    constructor() {
        this.traitProperty = 'Value';
    }

    traitMethod() {
        return 'Method return';
    }
}
```

In the above example, the `ClassWithTrait` class will now contain a `traitProperty` property and a `traitMethod` method, derived from the trait.

Classes can override methods and properties in the trait by defining them in the class itself - if I were to add `traitMethod` into the `ClassWithTrait` class, that method will be used over the one defined in the trait.

Traits will also be retrieved up the class hierarchy, so any traits defined in parent classes will also be loaded in the child classes. The only difference between this and PHP traits is that a locally-defined trait will *not* overwrite a method defined in parent classes.

#### Differences from PHP traits:
- Traits are classes in JavaScript.
- Locally defined traits do *not* overwrite methods from parent classes.
- Two or more traits can have the same method name - precedence is first-come-first-serve, (ie. the first trait that provides the method will be used), since aliasing does not exist.
- Calling `instanceof` of a class with the trait does not return `true` if the testing against the trait. You will have to run checks against the `traits()` method directly. This is because classes are transpiled to functions and objects on build.


LukeTowers:
@bennothommo is this ready to merge in?

bennothommo:
@LukeTowers I'll can the PR here. I've slightly changed how the traits work in the standalone version that I feel will work better. I will document them once I've got everything ready.

PR #876, opened 2023-03-29
Remove left-over debug info

AIC-BV:

### DIFF
--- a/modules/cms/models/ThemeLog.php
+++ b/modules/cms/models/ThemeLog.php
@@ -77,12 +77,6 @@ class ThemeLog extends Model
         $newContent = $template->toCompiled();
         $oldContent = $template->getOriginal('content');
 
-        if ($newContent === $oldContent && $templateName === $oldTemplateName && !$isDelete) {
-            traceLog($newContent, $oldContent);
-            traceLog('Content not dirty for: '. $template->getObjectTypeDirName().'/'.$template->fileName);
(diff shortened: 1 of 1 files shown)


modules/cms/models/ThemeLog.php
```diff
@@ -77,12 +77,6 @@ public static function add(HalcyonModel $template, $type = null)
         $newContent = $template->toCompiled();
         $oldContent = $template->getOriginal('content');
 
-        if ($newContent === $oldContent && $templateName === $oldTemplateName && !$isDelete) {
-            traceLog($newContent, $oldContent);
-            traceLog('Content not dirty for: '. $template->getObjectTypeDirName().'/'.$template->fileName);
-            return;
```
LukeTowers, line 83: This would change the behaviour by removing the short circuit that would occur when no changes are present; thus causing this method to always add an entry even when there are no changes.

AIC-BV: Didn't see the return statement

AIC-BV: Only removing the traceLog would be the correct way to go?

LukeTowers: Yeppers

AIC-BV: I'm derping around, can't find how to edit this

Issue #881, opened 2023-04-03
Backend reorder not working correctly


ericp-mrel:
### Winter CMS Build

1.1

### PHP Version

8.0

### Database engine

MySQL/MariaDB

### Issue description

There is an issue where sorting from the backend reorder list view does not work when records all contain the same/duplicate sort order value since the `winter.reorder.js` file does not recalculate the sort order value and re-uses the existing sort order values from the DB and just switches what record that value is applied to. This is a problem if you add sorting functionality to an existing table with records already created and then the migration assigns a default sort order value of 0 to every record and when you create a new record, it will contain the sort order of 0 which will also cause sorting for that record to stop working if it conflicts with another.

![Screenshot 2023-04-03 at 10 10 52 AM](https://user-images.githubusercontent.com/63664924/229552071-d50548e7-68e1-4bed-ab1f-9b4acae976ab.png)

<img width="764" alt="Screenshot 2023-04-03 at 10 11 33 AM" src="https://user-images.githubusercontent.com/63664924/229552093-073fe882-7fb0-4a49-a575-8894ad5d81a8.png">

### Steps to replicate

1. Create records in a table without sorting functionality.
2. Add new `sort_order` column to table and have the migration assign default `sort_order` value of 0 to every record.
3. Implement Reorder Controller functionality
4. Try to re-order list of records from backend, but notice the new order of the records is not reflected in the DB / after reloading the page.

### Workaround

1. During migration, loop through every record and adjust the `sort_order` to be the same as the primary key.
2. When creating new records, override default `sort_order` so it matches the primary key as well so we don't run into duplicates.

bennothommo:
@ericp-mrel the documentation does state there is an expectation that the developer has to assign sort order values for existing records - see the note at the bottom of the Sortable chapter here: https://wintercms.com/docs/database/traits#sortable

Given that they would need to create a migration in the first place (to add the `sort_order` column to the model's table), that would be the place to put the query to assign sort orders.

ericp-mrel:
Yeah, sorry I see it now that you point it out. I still personally feel it would be a bit of a nicer experience to not have to do a migration like that assigning sort order values in order for it to actually work from the backend UI.

I'll let you guys decide on if that's something you'd be willing to consider changing, otherwise you can close this ticket since as you pointed out in the documentation this isn't really a bug, but could be more of an enhancement.

[...]

PR #882, opened 2023-04-03
Re-calculate new sort order values.

ericp-mrel:
Fixes https://github.com/wintercms/winter/issues/881

Re-calculate new sort order values so we no longer have an issue with duplicate sort order values causing records not to be sorted correctly.

This probably needs to be fixed in v1.2 as well, but I haven't tested that version yet.

### DIFF
--- a/modules/backend/behaviors/reordercontroller/assets/js/winter.reorder.js
+++ b/modules/backend/behaviors/reordercontroller/assets/js/winter.reorder.js
@@ -69,8 +69,8 @@
         this.initSortingSimple = function () {
             var sortOrders = []
 
-            $('#reorderTreeList li').each(function(){
-                sortOrders.push($(this).data('recordSortOrder'))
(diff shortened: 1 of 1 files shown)


bennothommo:
@ericp-mrel would you mind applying this fix towards 1.2? Technically, 1.1 is in "security-only" fixes at this point.

LukeTowers:
@ericp-mrel are you still interested in getting this merged?

ericp-mrel:
@LukeTowers Sorry, I don't get to spend as time much working on WInterCMS as much these days. Would it be easier if I submit a new PR against the 1.2 branch rather than trying to update this PR to a new branch? I've never tried switching a PR to a different branch, so I'm not sure what may happen / break when doing so.

Thanks

LukeTowers:
Sure, whatever's easiest for you :)

PR #891, opened 2023-04-23
Add inputmode attribute to datepicker formwidget

mjauvin:
On mobile, when the DatePicker formwidget opens, you don't want the keyboard to popup.

In order to prevent this behavior, we add inputmode=none to the input field. (ref. https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/#aa-none)

This can still be overriden in the field config like this:
```
date:
    type: datepicker
    attributes:
      inputmode: text
```

### DIFF
--- a/modules/backend/formwidgets/DatePicker.php
+++ b/modules/backend/formwidgets/DatePicker.php
@@ -102,6 +102,14 @@ class DatePicker extends FormWidgetBase
                 ? Carbon::createFromTimestamp($this->maxDate)
                 : Carbon::parse($this->maxDate);
         }
+
+        // Add inputmode attribute (can still be overriden in field config)
+        $fieldAttributes = array_merge(
+            ['inputmode' => 'none'],
+            $this->formField->getAttributes('field', false),
+        );
+
+        $this->formField->attributes($fieldAttributes, 'field');
     }
 
     /**
--- a/modules/backend/formwidgets/DatePicker.php
+++ b/modules/backend/formwidgets/DatePicker.php
@@ -102,14 +102,6 @@ class DatePicker extends FormWidgetBase
                 ? Carbon::createFromTimestamp($this->maxDate)
                 : Carbon::parse($this->maxDate);
         }
-
-        // Add inputmode attribute (can still be overriden in field config)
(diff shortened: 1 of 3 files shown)


bennothommo:
@mjauvin I'm going to bump this to v1.2.3 as v1.2.2 is locked from further issues/PRs so that we can get it out the door. :)

mjauvin:
What about these errors @bennothommo ?

bennothommo:
Probably just needs to be re-run. Doing that now :)

EDIT: This branch needs to pull in the changes from `develop` to fix the issue.

LukeTowers:
@mjauvin the only concern I have with this is if the datepicker widget fails to load. Perhaps we could instead set the attribute as the final step of the datepicker initializing so that it will only take effect when we know that the datepicker JS will be able to handle the date selection before we prevent the users from being able to enter data manually?

mjauvin:
An alternative would be to simply document this and let people enable this themselves?

But frankly, most if not all our formwidgets would be broken if their js didn't load/run

LukeTowers:
Not necessarily, it's happened before that only specific widgets broke 😜.

bennothommo:
@mjauvin still need something like what @LukeTowers [suggested](https://github.com/wintercms/winter/pull/891#issuecomment-1529155407) to be implemented. I agree most widgets would not work with JS, but where we can, we should gracefully degrade.

mjauvin:
Allright, hope this is satisfactory now!

LukeTowers:
@mjauvin any chance you could provide a before / after screen recording?

mjauvin:
Are you serious?

Before: when you click on a date/time picker field, the keyboard pops up along with the date/time selector

After: only the date/time selector pops up.

LukeTowers:
@mjauvin I understand that, I just want verification that the JS approach works correctly 😜

mjauvin:
> @mjauvin I understand that, I just want verification that the JS approach works correctly stuck_out_tongue_winking_eye

It works, I tested it.

LukeTowers:
Good enough for me :P

LukeTowers:
Will merge after releasing 1.2.2

PR #899, opened 2023-04-29
Improve docs for checking for the correct Form widget instance

mjauvin:

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -496,6 +496,7 @@ class Form extends WidgetBase
         /**
          * @event backend.form.extendFieldsBefore
          * Called before the form fields are defined
+         * >**NOTE**: You should test `$formWidget->isNested` to prevent adding fields to nested elements
          *
          * Example usage:
          *
@@ -569,6 +570,7 @@ class Form extends WidgetBase
         /**
          * @event backend.form.extendFields
          * Called after the form fields are defined
+         * >**NOTE**: You should test `$formWidget->isNested` to prevent adding fields to nested elements
          *
          * Example usage:
          *
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -522,8 +522,12 @@ class Form extends WidgetBase
          * Or
          *
          *     $formWidget->bindEvent('form.extendFieldsBefore', function () use ((\Backend\Widgets\Form $formWidget)) {
-         *         // You should always check to see if you're extending correct model/controller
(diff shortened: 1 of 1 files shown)


bennothommo:
Just a thought - should we even call the `extendFields` events for nested forms? `isNested` is only `true` for nested form and repeater widgets, and it stands to reason that the user can still extend the primary form to add fields to these widgets, if necessary.

mjauvin:
> Just a thought - should we even call the `extendFields` events for nested forms? `isNested` is only `true` for nested form and repeater widgets, and it stands to reason that the user can still extend the primary form to add fields to these widgets, if necessary.

How would that work exactly? The user would get the nested/repeater widgets from the main formWidget and extend its fields directly ?

LukeTowers:
> Just a thought - should we even call the `extendFields` events for nested forms? `isNested` is only `true` for nested form and repeater widgets, and it stands to reason that the user can still extend the primary form to add fields to these widgets, if necessary.

That seems like a can of worms I don't want to open.

bennothommo:
This requirement for checking `isNested` has tripped up a few people now, and I don't think it's a good experience for form changes to spill into the nested forms unless its explicit, nor is it ideal that we even have to document this whole "you need to add some condition checks on `isNested` if you don't want your nested forms to be affected."

LukeTowers:
The problem is how people are attaching the listener. The global event fires at all times on all instances of a Form widget and it must stay that way regardless of whether or not a Form instance is nested.

It is already standard practice to check for the model class match, the controller class match, and the isNested call. I would recommend that we just provide a better example in the code example section rather than having a separate call out specifically for the isNested property.

mjauvin:
Done @LukeTowers

modules/backend/widgets/Form.php
```diff
@@ -500,8 +500,18 @@ protected function defineFormFields()
...
+         *             return;
+         *         }
+         *
+         *         // Only for the User model
+         *         if (!$formWidget->model instanceof \Winter\User\Models\User) {
+         *             return;
+         *         }
+         *
+         *         // Only for main form fields
+         *         if ($formWidget->isNested) {
```
LukeTowers:
```suggestion
         *         // Check that we're extending the correct Form widget instance
         *         if (
         *             !($formWidget->getController() instanceof \Winter\User\Controllers\Users)
         *             || !($formWidget->model instanceof \Winter\User\Models\User)
         *             || $formWidget->isNested
         *         ) {
```

mjauvin:
@bennothommo @LukeTowers should that be moved to 1.2.2 since it has been merged already?

PR #902, opened 2023-05-15
Provide the current Form instance to partials

mjauvin:
There is currently no way for a partial to know the form's previewMode, especially when in a repeater or nestedform firmwidget.

### DIFF
--- a/modules/backend/widgets/form/partials/_field_partial.php
+++ b/modules/backend/widgets/form/partials/_field_partial.php
@@ -4,5 +4,6 @@
     'formValue' => $field->value,
     'model'     => $formModel,
     'field'     => $field,
-    'value'     => $field->value
+    'value'     => $field->value,
+    'previewMode' => $this->previewMode,
 ]) ?>
--- a/modules/backend/widgets/form/partials/_field_partial.php
+++ b/modules/backend/widgets/form/partials/_field_partial.php
@@ -1,9 +1,9 @@
 <?= $this->controller->makePartial($field->path ?: $field->fieldName, [
-    'formModel' => $formModel,
-    'formField' => $field,
-    'formValue' => $field->value,
-    'model'     => $formModel,
-    'field'     => $field,
-    'value'     => $field->value,
-    'previewMode' => $this->previewMode,
+    'formWidget' => $this,
+    'formModel'  => $formModel,
+    'formField'  => $field,
+    'formValue'  => $field->value,
+    'model'      => $formModel,
+    'field'      => $field,
+    'value'      => $field->value
 ]) ?>


bennothommo:
@mjauvin what about `$this->formGetContext() === 'preview'`?

LukeTowers:
Why does it need to?

AIC-BV:
> Why does it need to?

Because when you are in update context and your repeater is disabled,
All fields in the repeater are disabled except the partials (they don't get saved but they are visually enabled)
image
```
_mypartial.htm
<?php if ($this->previewMode): ?>
    <span class="form-control"><?= isset($field->value) ? e($field->value) : '&nbsp;' ?></span>
<?php else: ?>
    ... This code is executed
<?php endif ?>
```

bennothommo:
@LukeTowers I spoke with @AIC-BV on Discord and the issue is that `previewMode` is set to `true` for a repeater or nested form if they are disabled (regardless of the context), so it's expected that all fields within should be shown in a preview context in that case, but with partials there doesn't seem to be a way to get that `previewMode` variable.

bennothommo (approved):
All good to be merged in with v1.2.3

AIC-BV:
> 

It looks fine, but it is not working 😅 
Its `true` in `_field_partial.php`, but still `null` in my partial

bennothommo:
@AIC-BV are you using `$this->previewMode` or `$previewMode` in your partial? Because you should be using the second one ;)

AIC-BV:
> @AIC-BV are you using `$this->previewMode` or `$previewMode` in your partial? Because you should be using the second one ;)

Yes 🤦‍♂️ 
`<?php if ($this->previewMode || $previewMode): ?>`
That works! 😁

mjauvin:
Can we merge this one with 1.2.2 as well @LukeTowers it's definitely not going to break anything, it just add a variable for the partial.

LukeTowers:
Not yet @mjauvin, I'm still not convinced of it and it adds to the API surface. Why is previewMode necessary and not just looking at the disabled state of the current field?

mjauvin:
> Not yet @mjauvin, I'm still not convinced of it and it adds to the API surface. Why is previewMode necessary and not just looking at the disabled state of the current field?

That information is NOT available elsewhere when within a repeater/nestedform @LukeTowers

mjauvin:
@LukeTowers the only information sent to the partial is this:
```php
<?= $this->controller->makePartial($field->path ?: $field->fieldName, [
    'formModel' => $formModel,
    'formField' => $field,
    'formValue' => $field->value,
    'model'     => $formModel,
    'field'     => $field,
    'value'     => $field->value
]) ?>       
```
`model` is not useful for previewMode
`field` does not contain this information (unlike other formwidgets).

Now it would be possible to set the previewMode in the field as well (from backend/widgets/Form.php) but I find it more intrusive.

mjauvin:
@LukeTowers I can also do the following if you prefer:
```diff
diff --git a/modules/backend/widgets/form/partials/_field_partial.php b/modules/backend/widgets/form/partials/_field_partial.php
index ae545876a..5b6f54277 100644
--- a/modules/backend/widgets/form/partials/_field_partial.php
+++ b/modules/backend/widgets/form/partials/_field_partial.php
@@ -1,3 +1,5 @@
+<?php $field->previewMode = $this->previewMode ?>
+
 <?= $this->controller->makePartial($field->path ?: $field->fieldName, [
     'formModel' => $formModel,
     'formField' => $field,
@@ -5,5 +7,4 @@
     'model'     => $formModel,
     'field'     => $field,
     'value'     => $field->value,
-    'previewMode' => $this->previewMode,
 ]) ?>
```

mjauvin:
That works too, but how is it different about the API surface ?

LukeTowers:
Because it's more purposeful. Adding the variable `previewMode` to partials is non-standard and solves only this issue, adding the variable `formWidget` follows the convention of the other variables that are already provided, solves the current issue, and makes the partial field have a more similar set of variables to what is available to other field type partials; therefore hopefully solving other potential future issues.

PR #904, opened 2023-05-16
Added RelationManager FormWidget

mjauvin:
Example usage:

```yaml

myRelationField:
  label: My Relation
  type: relationmanager
```

The following options can also be used in the field definition:

- readOnly - will default to parent (form, nestedform, repeater) previewMode/disabled/readOnly
- recordUrl
- recordOnClick

Docs PR for this: https://github.com/wintercms/docs/pull/131


bennothommo:
@mjauvin we should probably put some checks in the initialisation to make sure that certain requirements are met - namely, that the RelationController behaviour is in use on the controller that the form is connected to, and that a valid relation name is being used also.

Otherwise, we'll probably have a flood of issue reports from people who dove straight into using this field without knowing it's extra setup steps.

FWIW, I suspect that the above is the reason why it has never been made into a core field - it's entirely reliant on a different behaviour and configuration file.

mjauvin:
> @mjauvin we should probably put some checks in the initialisation to make sure that certain requirements are met - namely, that the RelationController behaviour is in use on the controller that the form is connected to, and that a valid relation name is being used also.
> 
> Otherwise, we'll probably have a flood of issue reports from people who dove straight into using this field without knowing it's extra setup steps.
> 
> FWIW, I suspect that the above is the reason why it has never been made into a core field - it's entirely reliant on a different behaviour and configuration file.

The following basic test can verify the field is a relation:
```php
$relation = $this->model->{$this->formField->fieldName}();
if (! $relation instanceof \Illuminate\Database\Eloquent\Relations\Relation) {
    ...
}
```
And we test if the `relationRender()` method exists on the controller:
```php
if (!method_exists($this->controller, 'relationRender')) {
    ...
}
```

mjauvin:
I'd like to merge this as well, anything missing ?

bennothommo:
@mjauvin I feel my comment still stands - `$this->controller->relationRender` is going to fail if someone uses this new widget type in their forms and they haven't included the `RelationController` behaviour. I still feel we need a check and a descriptive exception message if someone does this, so they know they *need* the `RelationController`.

mjauvin:
@bennothommo I added an exception with a proper message if the behavior is not implemented.

bennothommo:
@mjauvin perfect - one final tweak, could the exception also provide the field name - maybe `Field ":field" requires the ":controller" controller to implement the "RelationController" behavior`?

mjauvin:
> @mjauvin perfect - one final tweak, could the exception also provide the field name - maybe `Field ":field" requires the ":controller" controller to implement the "RelationController" behavior`?

done!

WebVPF:
@mjauvin You have a duplicate in your localization. You added the `missing_behavior` key twice.

mjauvin:
> @mjauvin You have a duplicate in your localization. You added the `missing_behavior` key twice.

Thanks!
I copied the line initially to have a reference and forgot to delete it...

bennothommo (approved):

bennothommo:
@LukeTowers anything to add here?

modules/backend/formwidgets/RelationManager.php
```diff
@@ -0,0 +1,70 @@
...
+     * @var string custom JavaScript code to execute when clicking on a record.
+     */
+    protected $recordOnClick;
+
+    public function init(): void
+    {
+        $this->fillFromConfig([
+            'readOnly',
+            'recordUrl',
+            'recordOnClick',
```
LukeTowers, line 42: Let's also add a `relation` property in case someone wants to have a different field name from the relationship definition name.

mjauvin: @LukeTowers do you mean to make this call configurable for its first argument ?
```
return $this->controller->relationRender($this->formField->fieldName, $options);
```

modules/backend/formwidgets/RelationManager.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public function render()
+    {
+        if (!$this->controller->isClassExtendedWith(\Backend\Behaviors\RelationController::class)) {
+            $error = Lang::get('backend::lang.relation.missing_behavior', [
+                'field' => $this->formField->fieldName,
+                'controller' => get_class($this->controller),
+            ]);
+            throw new ApplicationException($error);
```
LukeTowers: Should this be a SystemException instead?

modules/backend/formwidgets/RelationManager.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @var string custom JavaScript code to execute when clicking on a record.
+     */
+    protected $recordOnClick;
+
+    /**
+     * @var string relation name if different from the field name.
+     */
+    protected $relation;
```
LukeTowers: Can we add native types / default values to these properties?

modules/backend/formwidgets/RelationManager.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace Backend\FormWidgets;
+
+use SystemException;
+use Backend\Classes\FormField;
+use Backend\Classes\FormWidgetBase;
+use Lang;
```
LukeTowers:
```suggestion
use Backend\Classes\FormField;
use Backend\Classes\FormWidgetBase;
use Lang;
use SystemException;
```

modules/backend/formwidgets/RelationManager.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @var string custom JavaScript code to execute when clicking on a record.
+     */
+    protected string $recordOnClick;
+
+    /**
+     * @var string relation name if different from the field name.
+     */
+    protected string $relation;
```
LukeTowers:
```suggestion
    /**
     * Disables the ability to add, update, delete or create relations.
     */
    protected bool $readOnly;

    /**
     * Path to controller action to open a record.
     */
    protected string $recordUrl;

    /**
     * Custom JavaScript code to execute when clicking on a record.
     */
    protected string $recordOnClick;

    /**
     * Relation name if different from the field name.
     */
    protected string $relation;
```

mjauvin:
@LukeTowers what are all those errors suddenly !?

LukeTowers:
While I wait for the tests to finish running, do you think it would be possible for this FormWidget to dynamically inject & configure the RelationController if it isn't present already? That would be pretty cool :)

mjauvin:
> While I wait for the tests to finish running, do you think it would be possible for this FormWidget to dynamically inject & configure the RelationController if it isn't present already? That would be pretty cool :)

I guess it would be possible, the relationConfig being the trickiest part.

I'll look into this.

LukeTowers:
@mjauvin any luck?

mjauvin:
Initial investigation says it's not so easy

PR #905, opened 2023-05-17
Support trigger actions for dropdown fields in preview mode

### LINKED ISSUE #845: Trigger events not working in preview mode
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

Winter.Sitemap, Winter.Pages, Winter.TinyPNG, Winter.Translate, Winter.Redirects, Aic.Globals, Aic.ExtendWinterPages, Aic.Blog, Aic.Team, Aic.Account, Aic.Contact, Aic.Invoice, Aic.Discount, Aic.Aftersales, Aic.Reviews, Aic.Faq, StudioBosco.BackendNotifications, StudioBosco.BackendComments, ...

### Issue description

[Trigger events](https://wintercms.com/docs/backend/forms#field-trigger-events) don't seem to happen in preview mode, only in update.

Meaning the fields with `action: hide` are shown in preview
And the fields with `action: show` are hidden in preview

### Steps to replicate

Add trigger event and preview context

### Workaround

Not using preview context

mjauvin:
When in preview mode, if a dropdown field is used as another field's trigger, the trigger won't happen because the js code does not find the trigger field.

Fix #845

### DIFF
--- a/modules/backend/widgets/form/partials/_field_dropdown.php
+++ b/modules/backend/widgets/form/partials/_field_dropdown.php
@@ -9,12 +9,7 @@ $emptyOption = $field->getConfig('emptyOption', $field->placeholder);
     <div class="form-control" <?= $field->readOnly ? 'disabled="disabled"' : ''; ?>>
         <?= (isset($fieldOptions[$field->value])) ? e(trans($fieldOptions[$field->value])) : '' ?>
     </div>
-    <?php if ($field->readOnly): ?>
-        <input
-             type="hidden"
-             name="<?= $field->getName() ?>"
-             value="<?= $field->value ?>">
-    <?php endif; ?>
+    <input name="<?= $field->getName() ?>" value="<?= $field->value ?>">
 <?php else: ?>
     <select
         id="<?= $field->getId() ?>"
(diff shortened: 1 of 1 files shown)


bennothommo:
I'm curious as to how this fixes the issue. AFAICT, no trigger attributes are being populated here and in some other field types when in preview mode, so there will be a disconnect between triggers in preview vs. create/update.

mjauvin:
The trigger js code is searching for the trigger field and its value and couldn't find it without this hidden input element.

mjauvin:
> I'm curious as to how this fixes the issue. AFAICT, no trigger attributes are being populated here and in some other field types when in preview mode, so there will be a disconnect between triggers in preview vs. create/update.

Those trigger attributes you refer to are the element BEING triggered, not the trigger field itself.

bennothommo:
Alright, fair enough, this fixes dropdowns. But I'd say the issue will remain for other field types (ie. the text/email/number/etc. fields where the input is swapped with a span). So on face value, I don't think this entirely fixes the linked issue.

mjauvin:
The trigger attributes are added to the field container, I think this is what you're looking for.

This is done in the field_container partial

mjauvin:
> Alright, fair enough, this fixes dropdowns. But I'd say the issue will remain for other field types (ie. the text/email/number/etc. fields where the input is swapped with a span). So on face value, I don't think this entirely fixes the linked issue.

I see your point though, if the trigger field was a text field, that would still be broken, most likely. I'll revisit this for the other field types as well. But dropdown and switch/checkboxes are the most common trigger field types.

PR #907, opened 2023-05-18
Fix all remaining field partials in preview mode

mjauvin:

### DIFF
--- a/modules/backend/widgets/form/partials/_field_checkbox.php
+++ b/modules/backend/widgets/form/partials/_field_checkbox.php
@@ -1,18 +1,14 @@
 <!-- Checkbox -->
 <div class="checkbox custom-checkbox" tabindex="0">
     <input
-        type="hidden"
-        name="<?= $field->getName() ?>"
-        value="0"
-        <?= $this->previewMode ? 'disabled="disabled"' : '' ?>>
-    <input
         type="checkbox"
         id="<?= $field->getId() ?>"
         name="<?= $field->getName() ?>"
         value="1"
-        <?= $this->previewMode ? 'disabled="disabled"' : '' ?>
+        <?= $this->previewMode ? 'disabled' : '' ?>
         <?= $field->isSelected() ? 'checked="checked"' : '' ?>
-        <?= $field->getAttributes() ?>>
+        <?= $field->getAttributes() ?>
+    >
 
     <label for="<?= $field->getId() ?>">
         <?= e(trans($field->label)) ?>
--- a/modules/backend/widgets/form/partials/_field_checkbox.php
+++ b/modules/backend/widgets/form/partials/_field_checkbox.php
@@ -6,7 +6,7 @@
         name="<?= $field->getName() ?>"
         value="1"
         <?= $this->previewMode ? 'disabled' : '' ?>
-        <?= $field->isSelected() ? 'checked="checked"' : '' ?>
+        <?= $field->isSelected() ? 'checked' : '' ?>
         <?= $field->getAttributes() ?>
     >
 
--- a/modules/backend/widgets/form/partials/_field_checkbox.php
+++ b/modules/backend/widgets/form/partials/_field_checkbox.php
@@ -1,10 +1,10 @@
 <!-- Checkbox -->
 <div class="checkbox custom-checkbox" tabindex="0">
     <input
-       type="hidden"
-       name="<?= $field->getName() ?>"
-       value="0"
-       <?= $this->previewMode ? 'disabled' : '' ?>
+        type="hidden"
+        name="<?= $field->getName() ?>"
+        value="0"
+        <?= $this->previewMode ? 'disabled' : '' ?>
     >
     <input
         type="checkbox"
--- a/modules/backend/widgets/form/partials/_field_checkbox.php
+++ b/modules/backend/widgets/form/partials/_field_checkbox.php
@@ -1,5 +1,11 @@
 <!-- Checkbox -->
 <div class="checkbox custom-checkbox" tabindex="0">
+   <input
+       type="hidden"
+       name="<?= $field->getName() ?>"
+       value="0"
+       <?= $this->previewMode ? 'disabled' : '' ?>
+    >
     <input
         type="checkbox"
         id="<?= $field->getId() ?>"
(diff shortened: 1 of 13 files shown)


modules/backend/widgets/form/partials/_field_checkbox.php
```diff
@@ -1,18 +1,14 @@
 <!-- Checkbox -->
 <div class="checkbox custom-checkbox" tabindex="0">
-    <input
-        type="hidden"
-        name="<?= $field->getName() ?>"
-        value="0"
-        <?= $this->previewMode ? 'disabled="disabled"' : '' ?>>
```
LukeTowers, line 7: Pretty sure this breaks this field, if a checkbox isn't selected then the field name will be absent from the request entirely which is why we need this hidden field with a value of 0 here before the checkbox input itself.

mjauvin: @LukeTowers Wouldn't the field be missing from the post since its type is "hidden" ?

LukeTowers: No, the entire reason for hidden fields is so that they are hidden from the user but included in the POST

mjauvin: Fair enough, it's back in.

modules/backend/widgets/form/partials/_field_checkbox.php
```diff
@@ -1,18 +1,20 @@
...
+   <input
```
LukeTowers:
```suggestion
    <input
```

modules/backend/widgets/form/partials/_field_checkbox.php
```diff
@@ -1,18 +1,20 @@
...
 <div class="checkbox custom-checkbox" tabindex="0">
     <input
-        type="hidden"
-        name="<?= $field->getName() ?>"
-        value="0"
-        <?= $this->previewMode ? 'disabled="disabled"' : '' ?>>
+       type="hidden"
+       name="<?= $field->getName() ?>"
+       value="0"
+       <?= $this->previewMode ? 'disabled' : '' ?>
```
LukeTowers:
```suggestion
        type="hidden"
        name="<?= $field->getName() ?>"
        value="0"
        <?= $this->previewMode ? 'disabled' : '' ?>
```

modules/backend/widgets/form/partials/_field_dropdown.php
```diff
@@ -5,35 +5,29 @@
 ?>
 
 <!-- Dropdown -->
-<?php if ($this->previewMode || $field->readOnly): ?>
-    <div class="form-control" <?= $field->readOnly ? 'disabled="disabled"' : ''; ?>>
-        <?= (isset($fieldOptions[$field->value])) ? e(trans($fieldOptions[$field->value])) : '' ?>
-    </div>
-    <input type="hidden" name="<?= $field->getName() ?>" value="<?= $field->value ?>">
```
LukeTowers, line 12: Why did this file need changes? It already had a hidden element.

modules/backend/widgets/form/partials/_field_range.php
```diff
@@ -1,40 +1,37 @@
...
+$min = $field->config['min'] ?? 0;
+$max = $field->config['max'] ?? 100;
+$step = $field->config['step'] ?? 1;
+$value = $field->value;
+if ($min > $max) {
+    $min = $max - $step;
+}
+if (is_null($value)) {
+    $value = ($min + $max) / 2;
+}
```
LukeTowers, line 12:
```suggestion
    $min = $field->config['min'] ?? 0;
    $max = $field->config['max'] ?? 100;
    $step = $field->config['step'] ?? 1;
    $value = $field->value;
    if ($min > $max) {
        $min = $max - $step;
    }
    if (is_null($value)) {
        $value = ($min + $max) / 2;
    }
```

LukeTowers (changes requested):

modules/backend/widgets/form/partials/_field_dropdown.php
mjauvin: Because this PR tries to stay as close as possible to the regular markup when viewed in previewMode. Unless the markup I removed causes issue (like it did for the checkbox field), I think it's better to stay lean (easier to maintain)

LukeTowers: Arguably the easiest to maintain is the unchanged code 😜 What's the visual difference between the before / after?

mjauvin: > Arguably the easiest to maintain is the unchanged code stuck_out_tongue_winking_eye

Yes, I should have said "future" maintainability...

> What's the visual difference between the before / after?

The after shows a disabled select dropdown with the current value selected.

bennothommo:
One thing this is going to have a big impact on is how the forms look in preview mode now - effectively, it's just going to look like a form in which all fields are not editable, whereas before, it tried to simplify the view to only show text or simplified widget, like an actual "view screen" for a record.

I'm not saying it needs to be done in this PR, but I would propose that we'll need to follow up perhaps with some CSS that tries to go back to that look, because I personally liked that look.

LukeTowers:
Yeah, I'll need to see a before / after screenshot, preferably with the "Records" update / preview views from the test plugin

mjauvin:
@LukeTowers Any idea why we're getting this error in Code Quality check?
image

bennothommo:
@mjauvin the issue is that whole PHP block at the top of the `field-range` partial is indented, when it should not be - it should be aligned to the same column as the `<?php` tag.

mjauvin:
> @mjauvin the issue is that whole PHP block at the top of the `field-range` partial is indented, when it should not be - it should be aligned to the same column as the `<?php` tag.

@LukeTowers please don't fix things that aren't broken... :stuck_out_tongue_winking_eye:

LukeTowers:
@mjauvin I disagree with that coding style rule :P

mjauvin:
> @mjauvin I disagree with that coding style rule :P

Ok, feel free to change it! :)

bennothommo:
@LukeTowers technically it's not a rule specific to the templates but to PHP in general.

ie. You can't do this.

```php
<?php
    namespace My\Plugin;

    class MyClass
    {
        // ...
    }
```

LukeTowers:
@mjauvin is there still any interest in getting this merged?

mjauvin:
Personally, nope!

PR #908, opened 2023-05-19
Allow scoped and local extensions on core controllers.

mjauvin:
Complement https://github.com/wintercms/storm/pull/134

### DIFF
--- a/modules/backend/classes/BackendController.php
+++ b/modules/backend/classes/BackendController.php
@@ -2,6 +2,7 @@
 
 use Str;
 use App;
+use Closure;
 use File;
 use View;
 use Event;
@@ -11,7 +12,6 @@ use Response;
 use Illuminate\Routing\Controller as ControllerBase;
 use Winter\Storm\Router\Helper as RouterHelper;
 use System\Classes\PluginManager;
-use Closure;
 
 /**
  * This is the master controller for all back-end pages.
@@ -312,7 +312,7 @@ class BackendController extends ControllerBase
             return \Closure::fromCallable($params[0])->call($this, $params[1] ?? $this);
         }
 
-        return parent::__call($name, $params);
+        return $this->extendableCall($name, $params);
     }
 
     public static function __callStatic($name, $params)
--- a/modules/backend/classes/BackendController.php
+++ b/modules/backend/classes/BackendController.php
@@ -312,7 +312,7 @@ class BackendController extends ControllerBase
             return \Closure::fromCallable($params[0])->call($this, $params[1] ?? $this);
         }
 
-        return $this->extendableCall($name, $params);
+        return parent::__call($name, $params);
     }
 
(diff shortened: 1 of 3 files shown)


mjauvin:
the `__call()` and `__callStatic()` methods could also be added to `ExtendableTrait`, not sure if this is a good idea though.

LukeTowers:
Traits can't have magic methods AFAIK which is why they aren't there already.

LukeTowers:
Is this taking into account the existing __call & __callStatic methods on these classes / their parents?

mjauvin:
> Is this taking into account the existing __call & __callStatic methods on these classes / their parents?

It is in ExtendableTrait's `extendableCall()` & `extendableCallStatic()` methods.

LukeTowers:
@mjauvin I mean this PR is going to overwrite the existing base class' __call() and __callStatic() methods, we need to make sure the behavior remains the same.

mjauvin:
@LukeTowers I adjusted CmsController & BackendController's `__call()` method to replicate the previous behavior.

What should we do for the the `__callStatic()` methods of the same two classes which previously did not exist ? Throw a badMethodCall exception  like below?
```php
    public static function __callStatic($name, $params)
    {   
        if ($name === 'extend') {
            if (empty($params[0])) {
                throw new \InvalidArgumentException('The extend() method requires a callback parameter or closure.');
            }
            self::extendableExtendCallback($params[0], $params[1] ?? false, $params[2] ?? null);
            return;
        }
        
        throw new BadMethodCallException(sprintf(
            'Method %s::%s does not exist.', static::class, $method
        ));
    }
```

mjauvin:
Looking into this again, and the behavior is not changed without calling the parent's `__call()` / `__callStatic()` methods. If we don't call the ExtendableTrait's methods (extendableCall() and extendableCallStatic()), we loose the extensibility (other than the "extend()" method itself).

What do you think @bennothommo ?

mjauvin:
The base class for CmsController & BackendController is Illuminate\Routing\Controller which has no special handling other than returning BadMethodCallException().

So behavior will not change.

bennothommo (approved):
As far as I can tell, this should be OK. I had to do the same thing for the database models.

LukeTowers:
Does it still throw the exact same BadMethodCall Exception if the resolution fails?

mjauvin:
> Does it still throw the exact same BadMethodCall Exception if the resolution fails?

Depends which class and whether it's static or not...

Backend\Classes\Controller throws the exact same exception.

Backend\Classes\BackendController & Cms\Classes\CmsController used to throw a php Error exception since they didn't have any `__callStatic()` method defined, now they will throw the same BadMethodCall exception. Also, their `__call()` method will now thrown the same exception as before (BadMethodException), but the error message is slightly different: "Call to undefined method %s::%s()" vs "Method %s::%s does not exist."

PR #913, opened 2023-06-01
Improve responsive views

gergo85:
Current on Mobile:
before-mobile

Current on Tablet:
before-tablet

Current on PC:
before-pc

New design on Mobile:
after-mobile

New design on Tablet:
after-tablet

New design on PC:
after-pc


LukeTowers:
Thanks @gergo85, looks great!

Issue #42, opened 2021-03-16
Repeater maxItems doesn't take effect for the first item


LukeTowers:
When using repeater with maxItems=1, repeater does not work correctly. It adds one record and still shows the button to add one more. When you remove all records and start adding them again, the repeater works as expected.

Test with the Test Plugin using the Countries > Related repeater and setting maxItems: 1

`plugins\winter\test\models\country\fields.yaml`

Initial report by @GinoPane in octobercms/october#5533

bennothommo:
@LukeTowers is this is still occurring? I cannot replicate on a fresh copy of Winter.

LukeTowers:
No idea. Will close.

lex0r:
I'm afraid the issue is still there and happens exactly as it was described by @LukeTowers on the latest Winter release

lex0r:
A possible fix 
https://github.com/wintercms/winter/pull/919

LukeTowers:
Fixed by #919

lex0r:
The issue is happening again because of wrong jQuery selector for the "Add" button. The fix was suggested here: 
https://github.com/wintercms/winter/pull/1199

FYI @LukeTowers  @bennothommo

PR #919, opened 2023-06-09
Introduce a delay to give the widget enough time to show the just add…

lex0r:
…ed item before maxItem limit check is run.

This should address https://github.com/wintercms/winter/issues/42

### DIFF
--- a/modules/backend/formwidgets/repeater/assets/js/repeater.js
+++ b/modules/backend/formwidgets/repeater/assets/js/repeater.js
@@ -144,8 +144,11 @@
     }
 
     Repeater.prototype.onAddItemSuccess = function(ev) {
-        this.togglePrompt()
-        $(ev.target).closest('[data-field-name]').trigger('change.oc.formwidget')
+        var self = this;
+        setTimeout(function () {
+            self.togglePrompt();
+            $(ev.target).closest('[data-field-name]').trigger('change.oc.formwidget')
+        }, 10);
     }
 
     Repeater.prototype.togglePrompt = function () {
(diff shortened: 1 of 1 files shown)


mjauvin:
@lex0r even using a timeout of ZERO works... not sure what difference that makes

mjauvin:
Interesting explanation:
https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful

mjauvin:
I think this is an acceptable workaround.

Ultimately, it would be nice to find a more "global" fix for this race condition, if that's even possible.

modules/backend/formwidgets/repeater/assets/js/repeater.js
```diff
@@ -144,8 +144,11 @@
     }
 
     Repeater.prototype.onAddItemSuccess = function(ev) {
-        this.togglePrompt()
-        $(ev.target).closest('[data-field-name]').trigger('change.oc.formwidget')
+        var self = this;
+        setTimeout(function () {
+            self.togglePrompt();
+            $(ev.target).closest('[data-field-name]').trigger('change.oc.formwidget')
+        }, 10);
```
bennothommo: Would someone be willing to give this is a test?
```suggestion
        window.requestAnimationFrame(() => {
            this.togglePrompt();
            $(ev.target).closest('[data-field-name]').trigger('change.oc.formwidget')
        });
```

mjauvin: @bennothommo That works as well.

Issue #448, opened 2022-02-14
Improve form repeaters


jaxwilko:
This is a list of notes for things I would like to add, if anybody has any suggestions please add a comment :)

- Add new item under each item (rather than at the bottom and dragging it up)
- Speed up item generation
- Allow items to be dragged between scopes if acceptable
- More span options? (maybe)

AIC-BV:
I noticed that the lazy loading does not work in fields.yaml for a repeater field.
It returns an array but the lazy loading expects a string. Could be an improvement! 😄

EDIT: Hmm forget what I said, its more a bug, it works but in some cases it doesn't. Can't find the reason I'll have to dig deeper.

AIC-BV:
> I noticed that the lazy loading does not work in fields.yaml for a repeater field. It returns an array but the lazy loading expects a string. Could be an improvement! 😄
> 
> EDIT: Hmm forget what I said, its more a bug, it works but in some cases it doesn't. Can't find the reason I'll have to dig deeper.

So I did some deeper digging and found out the following:

Error: `htmlspecialchars(): Argument #1 ($string) must be of type string, array given`
Triggered by: I think it happens when you have a repeater like this in the database.
```
[{
  "name": "My name",
  "remarks": "My remarks",
  "image": "data:image\/jpeg;base64, ...",
  "production": {
      "drawing": "file.pdf",
      "approved_status": 1,
      "remarks": "Ok."
  }
}]
```

I can understand it errors on a base64.
The base64 was only temporary and will be removed and changed to an image path (no worries! 😅)
**But I'm pretty sure it happens on the json object**.

I can read the JSON object like this:
Note that I am using this inside a repeater

```
production[drawing]:
  type: mediafinder

production[approved_status]:
  type: dropdown

production[remarks]:
  type: textarea
```

This is definitely not a required improvement, could work around by splitting the JSON object or not using lazy 😄
Just wanted to let you guys know the issue!

AIC-BV:
Finally found this again! 😌
@jaxwilko I ran into the issue described above again, and it took me a while to figure it out. When I found the solution, I realised I had this problem in the past too...

I am using repeater groups, in which I created a pretty block using a Nestedform because I wanted tabs to make the repeater more compact:
![image](https://user-images.githubusercontent.com/89913092/177555508-6a08bd1b-a437-4ae1-99d2-508e78c49eb7.png)

In fields.yaml
```
blocks:
            tab: aic.blog::lang.form.blocks
[...]

PR #920, opened 2023-06-09
Add configuration attribute to force showing an empty repeater form a…

lex0r:
…utomatically if there are no existing repeater items yet.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -29,6 +29,11 @@ class Repeater extends FormWidgetBase
     public $sortable = true;
 
     /**
+     * @var bool Show an empty repeater form when it's first rendered and has no items yet.
+     */
+    public $showEmpty = false;
+
+    /**
      * @var string Field name to use for the title of collapsed items
      */
     public $titleFrom = false;
@@ -113,9 +118,10 @@ class Repeater extends FormWidgetBase
             'style',
             'prompt',
             'sortable',
+            'showEmpty',
             'titleFrom',
             'minItems',
-            'maxItems',
+            'maxItems'
         ]);
 
         if ($this->formField->disabled) {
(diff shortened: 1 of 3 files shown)


lex0r:
See https://github.com/wintercms/winter/issues/448

LukeTowers:
You should be able to accomplish the same end result in user space right now by setting the value of the attribute to an array with one empty array inside of it.

Something like this:

```php
Event::listen('backend.form.extendFields', function ($formWidget) {
    // Check that we're extending the correct Form widget instance
    if (
        !($formWidget->getController() instanceof \Winter\User\Controllers\Users)
        || !($formWidget->model instanceof \Winter\User\Models\User)
        || $formWidget->isNested
    ) {
        return;
    }

    if (empty($widget->model->my_attribute) && empty($widget->data->my_attribute)) {
        $widget->data->my_attribute = [[]];
    }
});
```

lex0r:
@LukeTowers do you think it's better to use the solution you suggested or make the widget's behaviour more explicit and understandable to users?

LukeTowers:
I'm not opposed to making the FormWidget's behaviour more explicit and understandable but I'm not a fan of the proposed implementation as it sits right now. `showEmpty` isn't clear to me what exactly it does and having a configuration value that then just applies a client side "hack" of clicking a button using JS isn't a great experience.

Do you have any alternative suggestions for an interface / configuration property that might be more clear / intuitive for developers that are trying to accomplish what you propose; alternatively do you have more to say about why your proposed option is the best one?

lex0r:
Naming is a notoriously difficult problems in software development :)

My solution is not the best one, it just does the job. Definitely can be improved, so anyone willing to do so is welcome.

LukeTowers:
> Naming is a notoriously difficult problems in software development :)
> 
> My solution is not the best one, it just does the job. Definitely can be improved, so anyone willing to do so is welcome.

Then in that case I guess my answer to the original question of "do you think it's better to use the solution you suggested" is "yes, unless you or anyone else can propose a better API / developer experience" 😜 For anyone's future reference looking to do something similar:

```php
Event::listen('backend.form.extendFields', function ($formWidget) {
    // Check that we're extending the correct Form widget instance
    if (
        !($formWidget->getController() instanceof \Winter\User\Controllers\Users)
        || !($formWidget->model instanceof \Winter\User\Models\User)
        || $formWidget->isNested
    ) {
        return;
    }

    // Ensure that there is always at least one item ready to be populated
    if (empty($widget->model->my_attribute) && empty($widget->data->my_attribute)) {
        $widget->data->my_attribute = [[]];
    }
});
```

PR #926, opened 2023-06-20
Add support for "grid" repeater

bennothommo:
Test plugin example: https://github.com/wintercms/wn-test-plugin/pull/16

This adds support for formatting a repeater as a "grid", allowing items to be horizontally listed and moving to a new row when the columns are filled. This was done mainly for the Blocks plugin, but can also be useful for other applications, such as creating a gallery or staff list, or any type of content that might be better demonstrated in a row as opposed to as a list.

### New configuration options:

- `mode`: Defines the repeater mode. Either `list` (default, same as current repeater functionality) or `grid`.
- `columns`: Defines the number of columns for the repeater. Can be between 2 and 6.
- `rowHeight`: Defines the height, in pixels, for each row. This is treated as a minimum height and is useful to give a visual indicator of how big each item will be. If a grid item ends up being larger than this value, the rest of the row will resize to accommodate.

### Example

image


LukeTowers:
Can you add a PR to the docs?

LukeTowers:
Screenshot 2023-06-23 at 1 37 37 AM
Needs better support for mobile / smaller screens. 

Perhaps https://stackoverflow.com/a/49502351/6652884 can provide some pointers?

bennothommo:
@LukeTowers I've sorted out some responsiveness.

PR #927, opened 2023-06-22
Apply filterFields before saving the form

### LINKED ISSUE #950: Visible fields are not saved correctly when filtering form fields
### Winter CMS Build

all

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

None

### Issue description

There is a bug in the Form widget : 
When you define a hidden field you want to show in the `filterFields()` model method, the later is not correctly saved.
This behavior occurs because the model form fields filtering is not applied in the `getSaveData()` widget method.

### Steps to replicate

1. Create a backend controller with its related model
2. Define a hidden field in the form fields definition
```yaml
#  Form Field Definitions

fields:
    value:
        label: Value
        hidden: true
        type: text
```
4. Define a `filterFields()` method in your test model that make your hidden field visible
```php
/**
 * Filtering form fields.
 *
 * @param object $fields
 * @param string $context
 */
 public function filterFields($fields, $context = null)
{
    $fields->value->hidden = false;
}
```
6. Trying to save the form in the backend

The value is not saved.

### Workaround

The workaround is to show your hidden fields in the form fields definition and hide them in your `filterFields` method.
That is to say, invert your visibility logic.

mjauvin:
Fixes #950

Restore form fields as they were before saving the form in case hidden/disabled fields were modified in `filterFields()` / `model.form.filterFields` event handler.

In Forms::getSaveData() hidden/disabled fields are skipped:
 ```php
foreach ($this->allFields as $field) {
    /*
     * Disabled and hidden should be omitted from data set
     */
    if ($field->disabled || $field->hidden) {
        continue;
    }
    ...
```

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1177,6 +1177,7 @@ class Form extends WidgetBase
     public function getSaveData()
     {
         $this->defineFormFields();
+        $this->applyFiltersFromModel();
 
         $result = [];
 
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -488,6 +488,8 @@ class Form extends WidgetBase
      */
     protected function defineFormFields()
     {
+        $this->applyFiltersFromModel();
+
         if ($this->fieldsDefined) {
             return;
         }
@@ -677,7 +679,6 @@ class Form extends WidgetBase
         }
 
         $this->fieldsDefined = true;
-        $this->applyFiltersFromModel();
     }
 
     /**
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -306,7 +306,6 @@ class Form extends WidgetBase
     protected function prepareVars()
     {
         $this->defineFormFields();
-        $this->applyFiltersFromModel();
         $this->vars['sessionKey'] = $this->getSessionKey();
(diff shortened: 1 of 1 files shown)


bennothommo:
@mjauvin just for cleanliness, should we just place one `$this->applyFiltersFromModel();` call at the end of the `defineFormFields()` method and remove it from everywhere else? That should cover your fix, and all other cases.

mjauvin:
> @mjauvin just for cleanliness, should we just place one `$this->applyFiltersFromModel();` call at the end of the `defineFormFields()` method and remove it from everywhere else? That should cover your fix, and all other cases.

Done @bennothommo

bennothommo:
@mjauvin looks good to me. Have you got an example to test it with? If so and you've confirmed it working, all good to merge.

mjauvin:
@bennothommo I tested it when called at the end of defineFormFields() and it only works when the form loads, not when the dependsOn triggers... calling it at the start of defineFormFields() now works as expected.

mjauvin:
@bennothommo it doesn't work when called at the end of defineFormFields() because that gets bypassed once the fields have been defined once... it' better to leave it as it was.

Thanks.

LukeTowers:
I'm concerned that this could be a BC break if someone was previously not expecting these fields to be skipped, any thoughts?

LukeTowers (approved):

PR #935, opened 2023-07-04
Populate migration file with actual model fields

mjauvin:
Populate the migration file with fields from the fields definition file and model properties (relations, timestamps, etc.)

Associated docs PR: https://github.com/wintercms/docs/pull/165


modules/system/console/CreateMigration.php
```diff
@@ -157,11 +158,13 @@ protected function prepareVars(): array
...
         if ($this->option('create') && $this->option('update')) {
-            $this->error('The create & update options cannot both be set at the same time');
-            return false;
+            throw new InvalidArgumentException('The create & update options cannot both be set at the same time');
         }
 
         if ($this->option('create')) {
+            if (empty($model)) {
+                throw new InvalidArgumentException('The create options requires the --model option');
+            }
```
LukeTowers: This shouldn't be required, you can have a creation migration without a model so long as you have a table name, and that is checked in line 176. 

The auto scaffolding of table structure based on model fields should be a separate flag that can work for create or update contexts alike.

modules/system/console/CreateMigration.php
```diff
@@ -199,6 +203,120 @@ protected function prepareVars(): array
...
+            default:
+                $dbType = 'string';
+        }
+        $required = $fieldConfig['required'] ?? false;
+
+        return [
+            'type' => $dbType,
+            'required' => $required,
+            'index' => in_array($name, ["slug"]) or str_ends_with($name, "_id"),
+        ];
```
LukeTowers, line 353: Can we move the mapping logic from fields.yaml to table structure into a separate helper class that can be more easily unit tested? Additionally, is there any overlap here with the Builder plugin?

modules/system/console/scaffold/migration/migration.create.stub
```diff
@@ -14,8 +14,19 @@ return new class extends Migration
     public function up()
     {
         Schema::create('{{ table }}', function (Blueprint $table) {
-            $table->id();
+            $table->increments({{ primaryKey }});
+{% for field,config in fields %}
+            $table->{{ config.type }}('{{ field }}'){{ config.required == false ? '->nullable()'}}{{ config.index ? '->index()' }};
+{% endfor %}
+{% for field in jsonable %}
+            $table->text('{{ field }}')->nullable();
```
LukeTowers: Any thoughts on text vs mediumtext for jsonable columns? I tend to lean more towards mediumText now instead of just text

mjauvin: last time I checked, text is huge already...

modules/system/console/CreateMigration.php
mjauvin: > This shouldn't be required, you can have a creation migration without a model so long as you have a table name, and that is checked in line 176.

It currently doesn't extract the model from the table name... the model was not used in the basic migration prior to this.

> The auto scaffolding of table structure based on model fields should be a separate flag that can work for create or update contexts alike.

Updating fields is a whole different story, this PR doesn't addess this.

LukeTowers: Right, but I'm saying that this check shouldn't be targeting the `--create` option, that's overloading the option to mean both create stub and autocreate from model stub, when instead it should really be a separate option altogether with its own validation to handle the autogeneration from the model.

We don't have to be perfect, but in theory we could apply the same logic for an update context, it would just be up to the developer to edit the results.

modules/system/console/scaffold/migration/migration.create.stub
LukeTowers: You'd think, but it's not impossible to run into limitations with it. It can handle roughly 5.9k to 13k words depending on whether the characters are multibyte or not (see https://stackoverflow.com/a/35785869/6652884), but if you start adding in overhead for the JSON characters and cramming a lot of fields in there you can run out of space relatively quickly.

LukeTowers: @bennothommo @jaxwilko any thoughts?

bennothommo: I agree with `mediumText`. I know it's probably overkill for some JSON objects, but it will mean it should cover all bases. And I don't think it has a huge ramification on storage space - it's an extra byte per row for storage.

mjauvin: ok, changed

modules/system/console/CreateMigration.php
mjauvin: > Right, but I'm saying that this check shouldn't be targeting the `--create` option, that's overloading the option to mean both create stub and autocreate from model stub, when instead it should really be a separate option altogether with its own validation to handle the autogeneration from the model.

Ok, autocreate from model now only triggers if --model option is supplied.

> We don't have to be perfect, but in theory we could apply the same logic for an update context, it would just be up to the developer to edit the results.

Ok, the fields are now included in the update stub if --update and --model options are used.

modules/system/console/CreateMigration.php
```diff
@@ -192,13 +192,130 @@ protected function prepareVars(): array
...
+            case 'number':
+            case 'range':
+                $dbType = 'integer';
+                break;
+            case 'datepicker':
+                $dbType = 'timestamp';
+                break;
+            case 'markdown':
+            case 'textarea':
+                $dbType = 'mediumText';
```
mjauvin, line 331: Should we use `mediumText` here as well or is `text` enough?

modules/system/console/CreateMigration.php
mjauvin: > Can we move the mapping logic from fields.yaml to table structure into a separate helper class that can be more easily unit tested?

Any suggestion ?

> Additionally, is there any overlap here with the Builder plugin?

Not really, from what I could see, this is done at a different level in Builder.

mjauvin: > Can we move the mapping logic from fields.yaml to table structure into a separate helper class that can be more easily unit tested?

Done!

modules/system/console/CreateMigration.php
bennothommo: Let's go `text` for textarea and perhaps `mediumText` for markdown, I reckon. Some markdown documents are mighty large and then you have to include formatting syntax.

mjauvin: Done

mjauvin:
Can this be merged or is there anything left to address?

modules/system/console/CreateMigration.php
```diff
@@ -192,13 +193,104 @@ protected function prepareVars(): array
...
+                }
+
+                $type = $config['type'] ?? 'text';
+
+                if (str_starts_with($field, '_')
+                    or $field === $model->getKeyName()
+                    or str_contains($field, '[')
+                    or in_array($type, ['fileupload','relation','relationmanager','repeater','section','hint'])
+                    or in_array($field, $model->purgeable ?? [])
+                    or $model->getRelationType($field)
```
LukeTowers: Can we add a style rule to only allow `||` as the or syntax?

modules/system/console/CreateMigration.php
```diff
@@ -192,13 +193,104 @@ protected function prepareVars(): array
...
+                if (str_contains($field, '@')) {
+                    list($field, $context) = explode('@', $field);
+                }
+
+                $type = $config['type'] ?? 'text';
+
+                if (str_starts_with($field, '_')
+                    or $field === $model->getKeyName()
+                    or str_contains($field, '[')
+                    or in_array($type, ['fileupload','relation','relationmanager','repeater','section','hint'])
```
LukeTowers:
```suggestion
                    or in_array($type, ['fileupload', 'relation', 'relationmanager', 'repeater', 'section', 'hint'])
```
Could we add a style rule for this as well?

modules/system/console/scaffold/migration/migration.create.stub
```diff
@@ -14,8 +14,23 @@ return new class extends Migration
     public function up()
     {
         Schema::create('{{ table }}', function (Blueprint $table) {
+{% if primaryKey %}
+            $table->increments('{{ primaryKey }}');
+{% else %}
             $table->id();
+{% endif %}
+{% for field,config in fields %}
+            $table->{{ config.type }}('{{ field }}'){{ config.required == false ? '->nullable()'}}{{ config.index ? '->index()' }};
```
LukeTowers:
```suggestion
            $table->{{ config.type }}('{{ field }}'){{ config.required == false ? '->nullable()' }}{{ config.index ? '->index()' }};
```

modules/system/console/scaffold/migration/migration.update.stub
```diff
@@ -14,7 +14,15 @@ return new class extends Migration
     public function up()
     {
         Schema::table('{{ table }}', function (Blueprint $table) {
-            //
+{% for field,config in fields %}
+            $table->{{ config.type }}('{{ field }}'){{ config.required == false ? '->nullable()'}}{{ config.index ? '->index()' }};
```
LukeTowers:
```suggestion
            $table->{{ config.type }}('{{ field }}'){{ config.required == false ? '->nullable()' }}{{ config.index ? '->index()' }};
```

modules/system/helpers/Migration.php
```diff
@@ -0,0 +1,65 @@
...
+
+namespace System\Helpers;
+
+use Winter\Storm\Database\Model;
+
+/**
+ * This helper class is used in migration scaffolding console scripts
+ *
+ */
+class Migration
```
LukeTowers: I'm not sold on the idea of a helper class for this, at least not without me putting some more thought into the API I'd like to see for it, so can we just make this a protected method of the console command for now and then make the test look at the actual generated output of the console command?

LukeTowers (dismissed):

modules/system/helpers/Migration.php
mjauvin: Dude, you're the one that specifically asked me to create a helper class for this...

ref. https://github.com/wintercms/winter/pull/935#discussion_r1252204195

LukeTowers: 🤦‍♂️ welp I guess that proves my memory is shot 😂 

The original intention was to have a helper class that could be used to translate between forms, columns, migrations, etc for a given form/column/migration schema but I realized that was a bit overly ambitious for now so I wanted to switch back to a protected method so that we could get this PR merged without having to nail the public API the first time. Apologies for my confusion; hopefully that explanation makes sense and you agree - if not let me know.

modules/system/console/CreateMigration.php
```diff
@@ -192,13 +193,104 @@ protected function prepareVars(): array
...
+            return $vars;
+        }
+
+        $vars['fields'] = [];
+
+        try {
+            $fields_path = '$/' . $vars['plugin_url'] . '/models/' . $vars['lower_model'] . '/fields.yaml';
+            $fields = Yaml::parseFile(File::symbolizePath($fields_path));
+        } catch (\Exception $e) {
+            die($e->getMessage());
```
LukeTowers: Is there anything else we can use other than die()? Why aren't we just passing up the exception at this point?

mjauvin: You mean remove thr try/catch statement entirely ?

LukeTowers: Or rethrow the exception, or provide a more helpful message when rethrowing.

LukeTowers:
@mjauvin do you know if it's possible to write a unit test for this?

mjauvin:
> @mjauvin do you know if it's possible to write a unit test for this?

What exactly would you like to test ?

bennothommo:
@mjauvin this should be easy to test - simply write out a couple of different form definitions and ensure that they are within the migration after the command is run. Or, if you don't want to parse the migration, you could run the migration and then check the table schema of the generated table.

mjauvin:
I wrote tests for the migration file generator.

It's not perfect as the test uses the SQLite db which does not support all field types MySQL/PostgreSQL do, but it should be good enough.

modules/system/tests/fixtures/plugins/winter/tester/models/testmodel/fields.yaml
```diff
@@ -0,0 +1,57 @@
+fields:
+    cb:
+      label: Checkbox
```
LukeTowers: What's with the switch from 4 to 2 space indentation here?

LukeTowers:
@mjauvin final thing (apart from the test yaml indentation), is there a related PR to the docs?

mjauvin:
Done @LukeTowers

PR #936, opened 2023-07-05
Support mass enable / disable in PermissionEditor

marvindurot:
This PR allows administrators to check or uncheck all permission checkboxes by clicking on the table head row.

When all permissions are checked, it will uncheck all permission checkboxes. Otherwise, it will check all permission checkboxes.

### DIFF
--- a/modules/backend/formwidgets/permissioneditor/assets/js/permissioneditor.js
+++ b/modules/backend/formwidgets/permissioneditor/assets/js/permissioneditor.js
@@ -13,6 +13,7 @@
     PermissionEditor.prototype.constructor = PermissionEditor
 
     PermissionEditor.prototype.init = function() {
+        $(document).on('click', '.permissioneditor table th.permission-type', this.proxy(this.onPermissionTypeClick))
         $(document).on('click', '.permissioneditor table td.permission-name', this.proxy(this.onPermissionNameClick))
         $(document).on('click', '.permissioneditor table tr.mode-checkbox input[type=checkbox]', this.proxy(this.onPermissionCheckboxClick))
         $(document).on('click', '.permissioneditor table tr.mode-radio input[type=radio]', this.proxy(this.onPermissionRadioClick))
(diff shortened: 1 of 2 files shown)


bennothommo:
Thanks for the contribution @durotm.

Which screen did you test this out on? Unfortunately it doesn't work for the permission list when creating/editing administrators, since that permission list uses radio inputs, not checkboxes.

modules/backend/formwidgets/permissioneditor/assets/js/permissioneditor.js
```diff
@@ -21,6 +22,30 @@
...
+                }
+            }
+        }
+
+        for (let i = 0; i < $rows.length; i++) {
+            var $row = $rows.eq(i)
+            var $checkbox = $row.find("input[type=checkbox]").eq(0)
+            if ($checkbox.length > 0) $checkbox.prop("checked", !allChecked)
+        }
+    };
```
bennothommo: This will make it work with both checkboxes and radio inputs:

```suggestion
    PermissionEditor.prototype.onPermissionTypeClick = function (ev) {
        var $rows = $(ev.target).closest('tr').nextAll()
        var index = $(ev.target).index()

        var allChecked = true

        for (let i = 0; i < $rows.length; i++) {
            var $row = $rows.eq(i)
            var $check = $row.find('td:nth-child(' + (index + 1) + ') input[type=radio], td:nth-child(' + (index + 1) + ') input[type=checkbox]').eq(0)

            if ($check.length > 0) {
                if (!$check[0].checked) {
                    allChecked = false
                    break
                }
            }
        }

        for (let i = 0; i < $rows.length; i++) {
            var $row = $rows.eq(i)
            var $check = $row.find('td:nth-child(' + (index + 1) + ') input[type=radio], td:nth-child(' + (index + 1) + ') input[type=checkbox]').eq(0)
            if ($check.length > 0) {
                if ($check.is('input[type=checkbox]')) {
                    $check.prop('checked', !allChecked)
                } else {
                    $check.prop('checked', true)
                }
            }
        }
    };
```

bennothommo (changes requested):

marvindurot:
Thanks for your review @bennothommo !

I tried a custom form definition wich uses the permission editor widget in checkboxgroup mode but I didn't test the radiogroup mode...

LukeTowers:
@durotm can I see before / after screenshots of the radiogroup mode and the checkboxgroup mode?

bennothommo:
@LukeTowers this is just a functionality change - there's no visual change to the permissions editor.

modules/backend/formwidgets/permissioneditor/assets/js/permissioneditor.js
```diff
@@ -21,6 +22,37 @@
...
+            var $row = $rows.eq(i)
+            var $check = $row.find('td:nth-child(' + (index + 1) + ') input[type=radio], td:nth-child(' + (index + 1) + ') input[type=checkbox]').eq(0)
+            if ($check.length > 0) {
+                if ($check.is('input[type=checkbox]')) {
+                    $check.prop('checked', !allChecked)
+                } else {
+                    $check.prop('checked', true)
+                }
+            }
+        }
```
LukeTowers, line 53: These look like they're duplicates of each other, can this logic not be simplified into one loop?

marvindurot: It can't be simplified because we don't know if there was at least one option which is not checked before processing the whole options.

Another approach could be unchecking all options until there was an unchecked option and loop back to check previous unchecked options. But this approach is less easy to read.

PR #940, opened 2023-07-09
Fix installation script with default env file

marvindurot:
This PR makes the installation script work again with the default environment file. 

Currently the installation script writes prompted configuration variables as second parameter of  **env** function call :
```php
'mysql' => [
            'charset' => 'utf8mb4',
            'collation' => 'utf8mb4_unicode_ci',
            'database' => env('DB_DATABASE', 'winter'),
            'driver' => 'mysql',
            'engine' => 'InnoDB',
            'host' => env('DB_HOST', '127.0.0.1'),
            'options' => extension_loaded('pdo_mysql') ? array_filter([
                PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),
            ]) : [],
            'password' => env('DB_PASSWORD', ''),
            'port' => env('DB_PORT', '3306'),
            'prefix' => '',
            'prefix_indexes' => true,
            'strict' => true,
            'unix_socket' => env('DB_SOCKET', ''),
            'url' => env('DATABASE_URL'),
            'username' => env('DB_USERNAME', 'winter'),
        ],
```

As the new default environment file is copied at composer installation, the prompted variables are overwritten by the environment file.

The solution is to write prompted variables to environment file too in the installation script thanks to this wonderful package : 
- https://github.com/wintercms/laravel-config-writer/tree/main#environment-files

### DIFF
--- a/modules/system/console/WinterInstall.php
+++ b/modules/system/console/WinterInstall.php
@@ -11,6 +11,7 @@ use System\Classes\UpdateManager;
 use Winter\Storm\Config\ConfigWriter;
 use Illuminate\Console\Command;
 use Illuminate\Encryption\Encrypter;
+use Winter\LaravelConfigWriter\EnvFile;
 use Symfony\Component\Console\Input\InputOption;
 use Exception;
 
@@ -44,6 +45,11 @@ class WinterInstall extends Command
     protected $configWriter;
 
     /**
+     * @var Winter\Storm\Config\EnvFile
+     */
+    protected $envFileWriter;
+
+    /**
      * Create a new command instance.
      */
     public function __construct()
@@ -51,6 +57,8 @@ class WinterInstall extends Command
         parent::__construct();
 
         $this->configWriter = new ConfigWriter;
+
+        $this->envFileWriter = new EnvFile(base_path('.env.example'));
 
         // Register aliases for backwards compatibility with October
         $this->setAliases(['october:install']);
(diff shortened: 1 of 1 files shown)


modules/system/console/WinterInstall.php
```diff
@@ -442,6 +456,17 @@ protected function writeToConfig($file, $values)
         $this->configWriter->toFile($configFile, $values);
     }
 
+    protected function writeToEnv($values)
+    {
+        $path = base_path('.env');
+
+        foreach ($values as $key => $value) {
+            $this->envFileWriter->set($key, $value);
```
LukeTowers, line 464: This should probably instead happen in the writeToConfig method with a mapping of file / key values to env key values and a check to see if the .env file exists to decide if values are written to `.env` or if they're written to the config files. If the .env file exists, then values should only be written there, not in both locations.

@bennothommo @jaxwilko any thoughts on that?

bennothommo: I don't necessarily mind it going in both locations as long as the `.env` file acts as an override (which it should given we have `env()` calls in the config). It would at least allow people who don't want to use `.env` files to still be able to install and use Winter.

LukeTowers: My point with that is to avoid writing config values into source control unintentionally.

LukeTowers:
@marvindurot is this something you're still interested in?

PR #943, opened 2023-07-11
Add index_onCopy method to ListController behavior

marvindurot:
This PR adds new method `index_onCopy` method to `ListController` behavior.

This method can be used to mass copy checked list elements like mass deletion works.

I also added a list column configuration variable to define which column is `protected` from mass copy.

For more documentation about replicating models, see : 
- https://laravel.com/docs/10.x/eloquent#replicating-models


bennothommo:
Oooo, I like this. Any chance you could hook this up to a list in the Backend, or perhaps the [Test plugin](https://github.com/wintercms/wn-test-plugin) so we can see it in action?

Some review comments to follow.

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +362,103 @@ public function index_onDelete()
...
+        }
+
+        $listConfig = $this->controller->listGetConfig($definition);
+
+        /*
+         * Validate checked identifiers
+         */
+        $checkedIds = post('checked');
+
+        if (!$checkedIds || !is_array($checkedIds) || !count($checkedIds)) {
```
bennothommo: Minor tidy up

```suggestion
        $checkedIds = post('checked', []);

        if (!is_array($checkedIds) || !count($checkedIds)) {
```

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +362,103 @@ public function index_onDelete()
...
+        $columnConfig = $this->makeConfig($listConfig->list);
+
+        $protectedColumns = [];
+
+        foreach ($columnConfig as $column => $definition) {
+            if (array_key_exists('protected', $definition)) {
+                if ($definition['protected']) {
+                    $protectedColumns[] = $column;
+                }
+            }
```
bennothommo: Minor tidy up

```suggestion
            if (array_key_exists('protected', $definition) && $definition['protected'] === true) {
                $protectedColumns[] = $column;
            }
```

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +362,103 @@ public function index_onDelete()
...
+         * Copy records
+         */
+        $records = $query->get();
+
+        if ($records->count()) {
+            foreach ($records as $record) {
+                $copy = $record->replicate($protectedColumns);
+                $this->controller->listBeforeCopy($record);
+                $copy->save();
+                $this->controller->listAfterCopy($record);
```
bennothommo: There might be a need to have some error checking or a try / catch block here, in case a record cannot be copied due to validation rules or uniqueness checks?

bennothommo (changes requested):

marvindurot:
> Oooo, I like this. Any chance you could hook this up to a list in the Backend, or perhaps the [Test plugin](https://github.com/wintercms/wn-test-plugin) so we can see it in action?
> 
> Some review comments to follow.

Thank you for your review!

I will hook this up to a list in the Test plugin :wink:

marvindurot:
I made an example on Test plugin : https://github.com/wintercms/wn-test-plugin/pull/17 :wink: 
screenshot

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +364,103 @@ public function index_onDelete()
...
+         * Create the query
+         */
+        $query = $model->newQuery();
+        $this->controller->listExtendQueryBefore($query, $definition);
+
+        $query->whereIn($model->getKeyName(), $checkedIds);
+        $this->controller->listExtendQuery($query, $definition);
+
+        /*
+         *  Retrieve protected columns from list column configuration
```
LukeTowers:
```suggestion
         * Retrieve protected columns from list column configuration
```

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +364,103 @@ public function index_onDelete()
         return $this->controller->listRefresh($definition);
     }
 
+    /**
+     * Bulk copy records.
+     * @return void
+     * @throws \Winter\Storm\Exception\ApplicationException when the parent definition is missing.
+     */
+    public function index_onCopy()
```
LukeTowers: This should either use "Duplicate" or "Replicate" terminology in the code to match Laravel's method and should use "Duplicate" terminology in the UI to make it more explicit to the user what is going on.

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +364,103 @@ public function index_onDelete()
...
+         */
+        $columnConfig = $this->makeConfig($listConfig->list);
+
+        $protectedColumns = [];
+
+        foreach ($columnConfig as $column => $config) {
+            if (array_key_exists('protected', $config) && $config['protected'] === true) {
+                $protectedColumns[] = $column;
+            }
+        }
```
LukeTowers, line 450: I would prefer if the configuration was available under `$listConfig->list` and supported either an explicit allow list or an explicit ignore list (example below):

`config_list.yaml`
```yaml
replication:
    enabled: true
    # one of the following, if both present then the most restrictive combination should be used
    allowed: [name, type, status]
    ignored: [id, slug, password]
```

The allowed functionality can probably be implemented by doing `$protected = array_diff($allowed, array_keys($record->getAttributes())`

And if replication is enabled and neither is provided then we should default to looking at the model's fillable / guarded definitions (allowing explicit configuration in the config_list.yaml to overrule those values if desired).

@bennothommo any thoughts?

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +364,103 @@ public function index_onDelete()
...
+                    $copy->save();
+                    $this->controller->listAfterCopy($record);
+                }
+                Flash::success(Lang::get(
+                    (!empty($listConfig->copyMessage))
+                        ? $listConfig->copyMessage
+                        : 'backend::lang.list.copy_selected_success'
+                ));
+            } catch (ValidationException $e) {
+                Flash::error($e->getErrors()->first());
```
LukeTowers: This should probably be an ApplicationException instead of a flash message

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +364,103 @@ public function index_onDelete()
         return $this->controller->listRefresh($definition);
     }
 
+    /**
+     * Bulk replicate records.
+     * @return void
+     * @throws \Winter\Storm\Exception\ApplicationException when the parent definition is missing.
```
LukeTowers:
```suggestion
     * @throws \Winter\Storm\Exception\ApplicationException when the parent definition is missing or when replication fails due to validation errors
```

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +364,103 @@ public function index_onDelete()
         return $this->controller->listRefresh($definition);
     }
 
+    /**
+     * Bulk replicate records.
+     * @return void
```
LukeTowers:
```suggestion
```

modules/backend/behaviors/ListController.php
```diff
@@ -362,6 +364,103 @@ public function index_onDelete()
...
+     * @throws \Winter\Storm\Exception\ApplicationException when the parent definition is missing.
+     */
+    public function index_onReplicate()
```
LukeTowers:
```suggestion
    public function index_onReplicate(): array
```

modules/backend/lang/fr/lang.php
```diff
@@ -226,8 +227,15 @@
         'check' => 'Sélectionner',
         'delete_selected' => 'Supprimer la sélection',
         'delete_selected_empty' => 'Il n\'y a aucun enregistrement à supprimer',
-        'delete_selected_confirm' => 'Confirmer la suppression des enregistrements sélectionnés ?',
+        'delete_selected_confirm' => 'Confirmer la suppression des eanregistrements sélectionnés ?',
```
LukeTowers, line 230: Was this intentional?

modules/backend/behaviors/ListController.php
LukeTowers: @bennothommo further idea, should model replication be controlled solely through properties on the model itself, so we can then more easily reuse this logic for a "Duplicate" button in the form context for duplicating records one by one?

bennothommo: I think we try and stick close to what Laravel has done, for feature parity. Laravel's `replicate()` method accepts fields as an "exclude" list, ie. exclude these fields from being replicated - I guess the assumption is that you don't want to have to have yet another array of fields you have to add new fields to. I also think that tying it to the `guarded` / `fillable` arrays might dilute what `replicate` is supposed to do - it's supposed to replicate an entire model after all.

bennothommo: To answer your question about keeping the control within the model - I think that would be OK, but I would: a) allow the list controller to override this if a list of fields is provided, and b) stick with Laravel and make it an exclude list.

bennothommo: We should also have an event fired on the model on replication, if there isn't one already.

LukeTowers: What concern do you have about also supporting allow lists rather than just ignore lists? Laravel's replicate() as implemented is a low level feature where developers have to explicitly call it and know what they're doing, this feature in Winter is meant for inclusion in backend lists (potentially all of them by default) and use by end users

LukeTowers: @bennothommo also what is your proposed implementation for the event? I'm fine leaving it out for now if we don't have a specific design for it yet.

bennothommo: I don't see a purpose for an allow list. This is a copy function - in most cases, it will copy everything. If we add an allow list, that's an extra bit of functionality on top of Laravel that we need to track in the future for breakages on major updates.

marvindurot: In conclusion what I am supposed to do in this pull request ?

LukeTowers:
@marvindurot are you still interested in this?

modules/backend/behaviors/ListController.php
LukeTowers: @marvindurot if you're still interested in having this merged then just ping me and I'll reopen. If you could submit a PR to the docs that should be all that's needed at this point.

PR #944, opened 2023-07-12
Limit the default amount of returned records

AIC-BV:
Makes sure that there is pagination and thus prevents all records being fetched and rendered at once in a single list

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -62,7 +62,7 @@ class Lists extends WidgetBase
     /**
      * @var int Maximum rows to display for each page.
      */
-    public $recordsPerPage;
+    public $recordsPerPage = 20;
 
     /**
      * @var array Options for number of items per page.


LukeTowers:
Is it still possible to disable pagination with this change? Why wouldn't you just make sure you set recordsPerPage on your list config?

AIC-BV:
I understand the argument!
In my case I didn't set the property, thus it was using 'all' as the default value.
It was returning all my records at once and I was having bad performance.

I think UX wise it should have a default value?
But yes, setting recordsPerPage solves my issue, but its seems to be missing the bigger picture? Which is why the PR..

Why would you allow users to fetch 1000 records at once? I mean, sure, it is the 'developers fault' but then again, with a default value of 20 or 100 or whatever, you can limit or prevent the 'developers fault' altogether?

For me it doesn't matter anymore, I added recordsPerPage, but for future users?

LukeTowers:
Most users will use the scaffolding commands or the builder plugin to build their controllers, which does have a default value of 20 set for that configuration: https://github.com/wintercms/winter/blob/develop/modules/backend/console/scaffold/controller/config_list.stub#L21.

AIC-BV:
I guess I'm the exception 😂
I just followed the documentation

LukeTowers:
@AIC-BV ah, you like doing things the hard way I see 😂 I've edited the docs just for you 😜 https://github.com/wintercms/docs/commit/d840ed2b7011fda4cb31b5b03635ff13c8837477

PR #951, opened 2023-07-25
Put mail template style tags in the head

RomainMazB:

### DIFF
--- a/modules/system/views/mail/layout-default.php
+++ b/modules/system/views/mail/layout-default.php
@@ -7,13 +7,12 @@ name = "Default layout"
 <head>
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
-</head>
-<body>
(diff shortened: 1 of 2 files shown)


LukeTowers:
Out of curiosity was the old location causing any issues?

mjauvin:
Technically, style tag have to be within the head tag... Strict mail client might not honor it within the body tag.

LukeTowers:
Fair enough, I'm just wondering if there any real world issues with it. Not opposed to merging, just wondering if there was a real world issue that prompted this because there wasn't anything in the PR description.

RomainMazB:
Nope @LukeTowers , I did not see any issue, I just worked on some mail layout and figured out it was misplaced.

LukeTowers:
According to https://www.campaignmonitor.com/css/style-element/style-in-body/ vs https://www.campaignmonitor.com/css/style-element/style-in-head; putting it in the head has slightly better support across some gmail clients, but loses support for yahoo mail on android. Works for me :)

PR #954, opened 2023-07-27
Create partials for the relation controller form footers

mjauvin:
We can now override the different Relation Controller form footers (buttons).


LukeTowers:
@mjauvin it would be better to replace the contents of the footer with a call to make a relation partial as that would allow for overriding the footer contents in the same way as the other behaviors, placing a properly named partial in your controller folder.

mjauvin:
> @mjauvin it would be better to replace the contents of the footer with a call to make a relation partial as that would allow for overriding the footer contents in the same way as the other behaviors, placing a properly named partial in your controller folder.

Perhaps, but would it also allow to replace this dynamically when extending a third party plugin controller?

LukeTowers:
@mjauvin sure, because plugins can extend controller's viewPaths arrays. Same way the Tailwind UI plugin works.

mjauvin:
This is ready to review.

LukeTowers:
@mjauvin can the partials include their containing divs as well? Also can you make a PR to the docs to accompany this?

mjauvin:
> @mjauvin can the partials include their containing divs as well? Also can you make a PR to the docs to accompany this?

Why ?

LukeTowers:
I don't have a good enough reason so fair enough 😜 

Can we rename it to `manage_form_footer_create` and `manage_form_footer_update` though? That way the relevant partials will be right next to each other in the file list

mjauvin:
> I don't have a good enough reason so fair enough stuck_out_tongue_winking_eye
> 
> Can we rename it to `manage_form_footer_create` and `manage_form_footer_update` though? That way the relevant partials will be right next to each other in the file list

Done!

LukeTowers:
Just need a docs PR to mention the new partials now

mjauvin:
> Just need a docs PR to mention the new partials now

In here ?
https://wintercms.com/docs/v1.2/docs/backend/relations#overriding-relation-partials

We don't document any of the partials specifically, why these ?

LukeTowers:
@mjauvin TBH I just didn't check what we had documented already or not, the current docs are fine for this change. Thanks! 😂

PR #955, opened 2023-07-28
Include extended refresh results on create or update pivot events.

hecc127:
Include extended refresh results on create or update pivot events.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1410,7 +1410,13 @@ class RelationController extends ControllerBehavior
             }
         });
 
-        return ['#'.$this->relationGetId('view') => $this->relationRenderView()];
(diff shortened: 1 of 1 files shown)


LukeTowers:
Could you give me an example (ideally the specific one for your use case) where this is used and what for?

hecc127:
Yes, I must create a product sale form, in this case the "price" is in the pivot table, in the view there is a partial that shows the summary data of the sale, including the total price, this change allows me to use the relationExtendRefreshResults(), to refresh the data when adding data in the relation, as in the management of simple updates.

LukeTowers:
You should be able to achieve this already via the following:

```php
public function onRelationManageCreate()
{
    $result = $this->asExtension('RelationController')->onRelationManageCreate();

    // Apply customizations
    $result = $result + $myCustomResult;

    return $result;
}
```

PR #956, opened 2023-08-02
Add video preview image for Media Finder form widget

bennothommo:
This PR adds a video preview image for `mediafinder` fields that have their `mode` config set to `video`. It renders the video screenshot by loading the first 10 seconds of the video and uses a HTML5 canvas element to take a snapshot as an image (stopping the video load afterwards).

image

In the course of this, I have also rewritten Media Finder to a Snowboard widget.

Only things left to do are:
- Make sure it works in `preview` and `update` contexts
- Perhaps cache or save the preview image so it doesn't have to load the video.
- Put something in place if the video cannot load (placeholder image perhaps?)
- Add preview for audio mode.


LukeTowers:
Something that I'd like to do as a general pattern across all our frontend development is to ensure that all assets live under either dist/ or src/ folders, with the objective being that /src folders aren't expected to exist on production servers or be persisted to CDNs.

So `assets/dist/js` vs `assets/js/dist`

modules/backend/formwidgets/mediafinder/partials/_mediafinder.php
```diff
@@ -7,17 +7,15 @@
     <?php
     switch ($mode) {
         case 'image':
-            echo $this->makePartial('image_single');
-            break;
         case 'video':
         case 'audio':
```
LukeTowers, line 11: What will the preview be for an audio file? Was this missed here?

LukeTowers:
With this and #958 it's looking more and more like we need to get #16 done as the thumbnail is definitely something that should stored in the item's metadata.

modules/backend/formwidgets/mediafinder/partials/_mediafinder.php
bennothommo: Ah, I do intend to add an audio player to the backend field preview if an audio file is selected.

LukeTowers:
@bennothommo what's the status of this?

bennothommo:
I've held this work until we have media library metadata in place.

PR #958, opened 2023-08-03
Add Mediafinder dimensions

AIC-BV:
This PR adds the dimensions to an image in the Mediafinder
image


modules/system/classes/MediaLibrary.php
```diff
@@ -636,9 +636,20 @@ protected function initLibraryItem($item, $itemType)
...
+        /*
+         * set image dimensions,
+         * folders dimensions is always NULL.
+         */
+        if ($itemType === MediaLibraryItem::TYPE_FILE) {
+            $imageDimensions = getimagesize($this->getStorageDisk()->path($item['path']));
+            $dimensions = "$imageDimensions[0] x $imageDimensions[1]";
+        } else {
+            $dimensions = null;
+        }
```
jaxwilko, line 648: `TYPE_FILE` would also include videos, audio, etc.

It this could be limited by only applying this logic to `$item->getFileType() === MediaLibraryItem::FILE_TYPE_IMAGE`, however that will not be available until after the `MediaLibraryItem` has been inited.

modules/system/classes/MediaLibraryItem.php
```diff
@@ -68,14 +73,16 @@ class MediaLibraryItem
...
     /**
      * @param string $path
+     * @param string $dimensions
      * @param int $size
      * @param int $lastModified
      * @param string $type
      * @param string $publicUrl
      */
-    public function __construct($path, $size, $lastModified, $type, $publicUrl)
+    public function __construct($path, $dimensions, $size, $lastModified, $type, $publicUrl)
```
LukeTowers, line 82: We should be passing this as the last argument for maximum BC, and since it's only relevant to specific types of media items we should start moving towards metadata support, so the parameter should actually be `array $metadata` and contain `'dimensions'` as a an array with `'width'` and `'height'` values.

modules/system/classes/MediaLibrary.php
```diff
@@ -636,9 +636,20 @@ protected function initLibraryItem($item, $itemType)
             $size = $this->getFolderItemCount($item['path']);
         }
 
+        /*
+         * set image dimensions,
+         * folders dimensions is always NULL.
+         */
+        if ($itemType === MediaLibraryItem::TYPE_FILE) {
+            $imageDimensions = getimagesize($this->getStorageDisk()->path($item['path']));
```
LukeTowers, line 644: This won't work for files stored on disks that aren't the local disk.

PR #969, opened 2023-09-07
Share variables shared via View::share() with Twig templates

### LINKED ISSUE #630: Twig not recovered View::share()
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

Source : https://wintercms.com/docs/cms/content#content-global-variables

There seems to be a problem with twig to get the values in the View::share() function 

The View::share() function return the good value but can not be recovered by twig on the front 

### Steps to replicate

In the boot function juste add 
`       
 View::share('site_name', 'Winter CMS');
`

and in the layout or any page / partial dump the "site_name" variable, its returning null 

### Workaround

_No response_

bennothommo:
Fixes https://github.com/wintercms/winter/issues/630

This allows global variables populated by the `View::share('name', 'value')` method to be used in Twig templates. The values will only be populated if they have not been specified as a global previously, to prevent overriding system globals such as `this.page`. It will also allow the pages themselves to still override the values if need be.

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -324,10 +324,9 @@ class Controller
          */
         $globalVars = ViewHelper::getGlobalVars();
         if (!empty($globalVars)) {
-            $specified = array_keys($this->getTwig()->getGlobals());
-
+            $existingGlobals = array_keys($this->getTwig()->getGlobals());
             foreach ($globalVars as $key => $value) {
-                if (!in_array($key, $specified)) {
+                if (!in_array($key, $existingGlobals)) {
                     $this->getTwig()->addGlobal($key, $value);
                 }
             }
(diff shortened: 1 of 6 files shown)


modules/cms/classes/Controller.php
```diff
@@ -319,6 +319,20 @@ public function runPage($page, $useAjax = true)
...
 
+        /*
+         * Add global vars defined by View::share() into Twig, only if they have yet to be specified.
+         */
+        $globalVars = ViewHelper::getGlobalVars();
+        if (!empty($globalVars)) {
+            $specified = array_keys($this->getTwig()->getGlobals());
+
+            foreach ($globalVars as $key => $value) {
+                if (!in_array($key, $specified)) {
```
```suggestion
            $existingGlobals = array_keys($this->getTwig()->getGlobals());
            foreach ($globalVars as $key => $value) {
                if (!in_array($key, $existingGlobals)) {
```

PR #974, opened 2023-09-13
Moved to Laravel 10

wverhoogt:
Laravel 10 updates for WinterCMS

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         }
     ],
     "require": {
-        "php": "^8.0.2",
+        "php": "^8.1",
         "winter/storm": "dev-develop as 1.2",
         "winter/wn-system-module": "dev-develop",
         "winter/wn-backend-module": "dev-develop",
--- a/composer.json
+++ b/composer.json
@@ -28,13 +28,35 @@
         "discord": "https://discord.gg/D5MFSPH6Ux",
         "source": "https://github.com/wintercms/winter"
     },
+    "repositories": [
+        {
+            "type": "path",
+            "url": "../winter-dev/modules/system",
+            "options": { "symlink": false  }
+        },
+        {
+            "type": "path",
+            "url": "../winter-dev/modules/backend",
+            "options": { "symlink": false  }
+        },
+        {
+            "type": "path",
+            "url": "../winter-dev/modules/cms",
+            "options": { "symlink": false  }
+        },
+        {
+            "type": "path",
+            "url": "../winter-dev/vendor/winter/storm",
+            "options": { "symlink": false  }
+        }
+    ],
     "require": {
         "php": "^8.0.2",
         "winter/storm": "dev-develop as 1.2",
         "winter/wn-system-module": "dev-develop",
         "winter/wn-backend-module": "dev-develop",
         "winter/wn-cms-module": "dev-develop",
-        "laravel/framework": "^9.1",
+        "laravel/framework": "^10.1",
         "wikimedia/composer-merge-plugin": "~2.1.0"
     },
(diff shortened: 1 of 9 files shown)


modules/backend/tests/classes/NavigationManagerTest.php
```diff
@@ -26,14 +26,14 @@ public function testRegisterMenuItems()
...
         $item = $items['WINTER.TEST.DASHBOARD'];
-        $this->assertObjectHasAttribute('code', $item);
-        $this->assertObjectHasAttribute('label', $item);
-        $this->assertObjectHasAttribute('icon', $item);
-        $this->assertObjectHasAttribute('url', $item);
-        $this->assertObjectHasAttribute('owner', $item);
-        $this->assertObjectHasAttribute('order', $item);
-        $this->assertObjectHasAttribute('permissions', $item);
-        $this->assertObjectHasAttribute('sideMenu', $item);
+        $this->assertObjectHasProperty('code', $item);
```
bennothommo, line 29: @wverhoogt with PHPUnit changes like this where it's a simple name change, I like to create stubs in the root `TestCase` class ([see here for example](https://github.com/wintercms/winter/blob/develop/modules/system/tests/bootstrap/TestCase.php#L61-L76)) so that the old names continue to work.

I'm happy for us to keep the changed names here, I'd just still like the old method name to still work so people don't have to update their tests.

composer.json
```diff
@@ -28,13 +28,35 @@
...
+            "type": "path",
+            "url": "../winter-dev/modules/cms",
+            "options": { "symlink": false  }
+        },
+        {
+            "type": "path",
+            "url": "../winter-dev/vendor/winter/storm",
+            "options": { "symlink": false  }
+        }
+    ],
```
bennothommo, line 52: Reminder that we'll have to drop this before merging.

bennothommo (changes requested):

modules/backend/tests/classes/NavigationManagerTest.php
wverhoogt: I´ve added the stub method.

modules/system/tests/bootstrap/TestCase.php
```diff
@@ -91,4 +92,22 @@ public static function assertRegExp(string $pattern, string $string, string $mes
...
+
+    /**
+     * Stub for `assertObjectHasAttribute` to allow compatibility with both PHPUnit 9 and 10.
+     *
+     * @param string $propertyName
+     * @param object $object
+     * @param string $message
+     * @return void
+     */
+    public static function assertObjectHasAttribute(string $propertyName, $object, string $message = ''): void
```
LukeTowers, line 104: Won't this fail to work as a drop in replacement as it is now a static method?

wverhoogt: Interestingly it was always a static method, just like assertRegExp, `$this->assertObjectHasProperty` calls the static method. I copied the stub from there, and it seems to work. Kinda strange indeed.

wverhoogt:
I noticed that the tests workflow also tests with PHP 8.0. Laravel 10 requires PHP 8.1+. Shall I change this workflow?

modules/system/tests/bootstrap/TestCase.php
bennothommo: @LukeTowers I think PHPUnit has a magic method that routes calls to the static methods.

bennothommo:
@wverhoogt yes please, we'll need to drop PHP 8.0 from the workflows on here and on your Storm PR too.

Issue #988, opened 2023-10-11
Some installation directory names lead to failed install


singular0:
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

None

### Issue description

Some particular target installation directory names (like `"/test"`) will lead to `compose create-project` failure.

### Steps to replicate

1. Run `composer create-project wintercms/winter /test`
2. Get the following error

```
> @php artisan winter:version
*** Detecting Winter CMS build...

In FileManifest.php line 139:

  Undefined array key "systems"


Script @php artisan winter:version handling the post-update-cmd event returned with error code 1
```

### Workaround

Use paths that are not substring of system file tree.

PR #989, opened 2023-10-11
Fix install error for some paths

singular0:
Instead of `str_replace`, which may, instead of truncating lead directory from the path, replace some random inner part of it, use `substr` to remove it.

### DIFF
--- a/modules/system/tests/classes/FileManifestTest.php
+++ b/modules/system/tests/classes/FileManifestTest.php
@@ -73,6 +73,6 @@ class FileManifestTest extends TestCase
         $class = new ReflectionClass('System\Classes\FileManifest');
         $method = $class->getMethod('getFilename');
 
-        $this->assertEquals('/test', $method->invoke($fileManifest, '/test/test'));
+        $this->assertEquals('/test', $method->invoke($fileManifest, '/modules/test/file1.php'));
     }
 }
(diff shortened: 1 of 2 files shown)


LukeTowers:
@singular0 would you be willing to add a test case for this?

singular0:
@LukeTowers well, this will not fit into `modules/system/tests/classes/FileManifestTest.php`, as to test you have to initialize `FileManifest` with `$root` path that is substring of inner files paths (like `"/test"`). not sure how to proceed here. create separate `TestCase` for this in a separate file?

LukeTowers:
@singular0 you can put it in the same test, you would just initialize a new local instance of FileManifest instead of using the pre-initialized instance.

singular0:
@LukeTowers ok, so i was unable to create direct test of the #988 itself as github runner (as well as my local macos system integrity protection) does not allow to create directories in `/`. the best thing i could come up with is the test that proves changed `FileManifest->getFilename()` to work correctly.

modules/system/tests/classes/FileManifestTest.php
```diff
@@ -61,4 +63,16 @@ public function testSingleModule()
             'test' => 'c0b794ff210862a4ce16223802efe6e28969f5a4fb42480ec8c2fef2da23d181',
         ], $this->fileManifest->getModuleChecksums());
     }
+
+    public function testGetFilename()
+    {
+        link(base_path('modules/system/tests/fixtures/manifest/1_0_1'), '/test');
```
LukeTowers: Does this clean up after itself?

PR #990, opened 2023-10-12
Add support for PDF 'image'

AIC-BV:
If the file is a PDF, there is an 'img' tag shown but that is not possible for the PDF src


AIC-BV:
image
With the right styling discussed in https://github.com/wintercms/winter/pull/985 we can, for example, center this

LukeTowers:
This should be solved by returning an icon from the thumbUrl instead and can be done with your own custom file models. If we add this specific icon for a specific file type why not add all file types in existence? It would be better to define that behaviour directly in the model class that's returning the thumbUrl so it's more consistent across the entire codebase.

AIC-BV:
> It would be better to define that behaviour directly in the model class that's returning the thumbUrl so it's more consistent across the entire codebase.

Sounds great to me!

PR #994, opened 2023-10-16
Use temporary rewriting of curly brackets when rendering Markdown content blocks

### LINKED ISSUE #992: Markdown parser fails in content file when using a content variable as the url
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

The new markdown parser fails in a contend file when using a content variable in place of a static url.

### Steps to replicate

Using this markup in a CMS page:
```twig
{% content "test-content.md" url="https://wintercms.com" %}
```
And this content file:
```markdown
[click here]({url})
```

### Workaround

Using `<a href="{url}">click here</a>` tag in the content file works.

bennothommo:
Fixes https://github.com/wintercms/winter/issues/992

The new Markdown parser (CommonMark) applies some escaping logic for links and some HTML attributes. This prevents the curly bracket variables working in these, as they are escaped upon parsing. Since the Commonmark library is very much locked down with `final` classes and the like, we can't directly override this behaviour, so this is the next best thing (although feels kinda dirty).

This replaces the curly brackets with another placeholder while the Markdown parsing takes place and then switches it back to curly brackets afterwards. The characters chosen should be safe from being parsed, I believe. It should also ensure that the curly brackets are still cached as it was with the previous behaviour.

### DIFF
--- a/modules/cms/classes/Content.php
+++ b/modules/cms/classes/Content.php
@@ -78,9 +78,6 @@ class Content extends CmsCompoundObject
      *
      * This method replaces curly variables in the content temporarily while Markdown rendering takes place, to
      * circumvent the escaping that Commonmark does on curly brackets in links.
-     *
-     * @param string $markup
-     * @return string
      */
     protected function parseMarkdownMarkup(string $markup): string
     {
--- a/modules/cms/classes/Content.php
+++ b/modules/cms/classes/Content.php
@@ -1,7 +1,9 @@
-<?php namespace Cms\Classes;
+<?php
 
-use File;
-use Markdown;
+namespace Cms\Classes;
+
+use Winter\Storm\Support\Facades\File;
+use Winter\Storm\Support\Facades\Markdown;
 
 /**
  * The CMS content file class.
@@ -62,12 +64,46 @@ class Content extends CmsCompoundObject
                 $result = htmlspecialchars($this->markup);
                 break;
             case 'md':
-                $result = Markdown::parse($this->markup);
+                $result = $this->parseMarkdownMarkup($this->markup);
                 break;
             default:
                 $result = $this->markup;
         }
 
         return $result;
+    }
+
+    /**
+     * Parse Markdown content files.
+     *
+     * This method replaces curly variables in the content temporarily while Markdown rendering takes place, to
+     * circumvent the escaping that Commonmark does on curly brackets in links.
+     *
+     * @param string $markup
+     * @return string
+     */
+    protected function parseMarkdownMarkup(string $markup): string
+    {
+        /**
+         * Replace variables temporarily for Markdown parsing, as the Commonmark library escapes curly brackets in
+         * links.
+         * @see https://github.com/wintercms/winter/issues/992
+         */
+        $variables = [];
+        $markup = preg_replace_callback('/\{([^\n \}]+)\}/', function (array $matches) use (&$variables) {
+            $signature = hash('sha256', $matches[1]);
+            if (!array_key_exists($signature, $variables)) {
+                $variables[$signature] = $matches[1];
+            }
+            return ".=VAR={$signature}=.";
+        }, $markup);
+
+        $markup = Markdown::parse($markup);
+
+        foreach ($variables as $signature => $variable) {
+            $markup = str_replace(".=VAR={$signature}=.", "{{$variable}}", $markup);
+        }
+
+        return $markup;
     }
 }


modules/cms/classes/Content.php
```diff
@@ -62,12 +64,46 @@ public function parseMarkup()
...
     }
+
+    /**
+     * Parse Markdown content files.
+     *
+     * This method replaces curly variables in the content temporarily while Markdown rendering takes place, to
+     * circumvent the escaping that Commonmark does on curly brackets in links.
+     *
+     * @param string $markup
+     * @return string
```
LukeTowers:
```suggestion
```
Docblock typehinting is not required when native typehints are present.

LukeTowers:
Can we have some unit tests added for this? If they could include attempts at abusing this new logic to somehow perform some sort of malicious action (if you can think of any) that would be great. The proposed fix could open the door for that type of potential attack vector.

bennothommo:
@LukeTowers the only specific opening this would provide would be if one were to be accepting any form of user input into a content block - they could put in the replacement signature (ie. `.=VAR=[some variable]=.`) into their input, and as long as the variable is actually passed into the rendering pipeline, it would be replaced.

I figured the likelihood that people would be accepting input into content blocks was extremely low, and at best, they would simply get access to a variable that's already made available to the page anyway (and can only be provided by the backend, so it would need to be a privileged user).

I will set up a test case or two for it though.

LukeTowers:
@bennothommo this is a horrible idea, but we could prevent it from happening by pre-replacing the replace target with something else and then changing it back afterwards.

So:
- replace `.=VAR=.` with `.=NOTVAR=.`
- replace `{}` with `.=VAR=.`
- markdown parse
- replace `.=VAR=.` with `{}`
- replace `.=NOTVAR=.` with `.=VAR=.`

I could see some potential attack vectors around the use of template fields and safe mode, so we should at least cover this case to avoid introducing any vectors unnecessarily.

bennothommo:
@LukeTowers slightly less horrible, but the other idea is we could add that sequence to the escaping functionality within Twig, so if it's fed through a Twig variable, it's escaped, but should be fine if it's directly in the template.

mjauvin:
Maybe a stupid question, but why can't we expand the content variables to their values BEFORE parsing the markup with Markdown::parse ?

bennothommo:
@mjauvin because the content blocks are cached on render. Putting in the variables before rendering would lock that set of variables in, and you'd lose the ability to re-use the content block with different values without clearing and re-rendering each time.

mjauvin:
Right, that couldn't be so simple... It was worth a shot!

LukeTowers:
@bennothommo what would that look like? Cause it sounds pretty horrible 😂

LukeTowers:
Also is there any flexibility to just changing the markdown parser to not cause us this issue in the first place?

bennothommo:
@LukeTowers it basically requires us to completely replace one of the core Commonmark extensions with our own copy of the same extension (plus any overrides to make the links work with the variables).

While I'm not adverse to that, it does mean that we then have to ensure that we're monitoring the Commonmark library for changes and implementing any fixes into our own.

LukeTowers:
Argh, how annoying. Let's give the twig escaping a try then I guess? Would that actually stop the theoretical attacks I was proposing?

bennothommo:
It should. I'll have to circle around to this in a couple of weeks though unfortunately. Currently knee-deep in a very time-critical project.

bennothommo:
@LukeTowers just circling back to this - I'm wondering if we even need Twig escaping? The particular issue here is with content blocks that are Markdown-formatted. Content blocks aren't put through the Twig parser at all, as far as I'm aware.

LukeTowers:
> @LukeTowers just circling back to this - I'm wondering if we even need Twig escaping? The particular issue here is with content blocks that are Markdown-formatted. Content blocks aren't put through the Twig parser at all, as far as I'm aware.

@bennothommo didn't you say that they are?

> @mjauvin because the content blocks are cached on render. Putting in the variables before rendering would lock that set of variables in, and you'd lose the ability to re-use the content block with different values without clearing and re-rendering each time.

LukeTowers:
@bennothommo what's the status on this?

bennothommo:
I really don't like this fix, TBH. I'm more inclined to can it and think of something else.

PR #996, opened 2023-10-18
Add support for PDF 'image'

AIC-BV:
See https://github.com/wintercms/winter/pull/990
PDF thumb path needs to be provided

### DIFF
--- a/modules/backend/formwidgets/FileUpload.php
+++ b/modules/backend/formwidgets/FileUpload.php
@@ -505,7 +505,13 @@ class FileUpload extends FormWidgetBase
         }
 
         $file->pathUrl = $path;
-        $file->thumbUrl = $thumb;
+        switch($file->extension) {
+            case 'pdf':
+                $file->thumbUrl = 'https://avatars.githubusercontent.com/u/7253840?v=4';
+                break;
+            default:
+                $file->thumbUrl = $thumb;
+        }
 
         return $file;
     }


AIC-BV:
@LukeTowers
@bennothommo
Can you help me finish this?

First file is saved.
Second file has just been uploaded and not yet saved.
All it needs is a path to a PDF image/icon and support for other types if you want
image

bennothommo:
@AIC-BV sorry, just so I'm clear, is this a `fileupload` type field that you have set to `mode: image` and are uploading a PDF to it?

AIC-BV:
> @AIC-BV sorry, just so I'm clear, is this a `fileupload` type field that you have set to `mode: image` and are uploading a PDF to it?

It might be 🤭😇

AIC-BV:
I tried to file mode and I don't like it at all, there is no preview and thus you have to open each file/image individually to actually know what it is. This significantly lowers efficiency.
You have to check each file manually. I really need the thumbnails.

It is possible that a client uploads a PDF or IMAGE
The mode IMAGE supports it perfectly fine. Only thing missing is a thumbnail for file types.

bennothommo:
Arguably, a PDF is not an image. It is a file, and yes, we don't have previews of files because it's very likely we *can't* generate a preview of some file types.

AIC-BV:
Its up to you!
I'm fine with not having a PDF icon/Word icon/... it just feels more complete.
Having support for the most common filetypes is better than none at all?
Anyways, it works either way!

bennothommo:
@AIC-BV I do like the way that the `image` type file uploads present, so I'm not adverse to us using the same "style" for standard file uploads too, but we would need some sort of preview for every file type (with a fallback, I would image, of a file icon with perhaps the extension within, kinda like what you suggested earlier).

Might be worth mocking that up first so we can see how it would look.

LukeTowers:
Will continue in #1045

PR #997, opened 2023-10-19
Add new multilingual button css support

AIC-BV:
https://github.com/wintercms/wn-translate-plugin/pull/73

### DIFF
--- a/modules/backend/formwidgets/markdowneditor/assets/js/markdowneditor.js
+++ b/modules/backend/formwidgets/markdowneditor/assets/js/markdowneditor.js
@@ -576,7 +576,6 @@
         }
 
         this.editor.focus()
-        
         $('[data-button-code="fullscreen"]', this.$toolbar).toggleClass('active')
     }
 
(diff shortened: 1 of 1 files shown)


AIC-BV:
Looking for a friendly soul to add this in the mlmarkdowneditor.js

LukeTowers:
I'm not really a fan of requiring this in the core just to make a plugin work, why can't the plugin handle this?

AIC-BV:
I guess it can, I was looking for a fix and the  required events are just not there (yet)

LukeTowers:
@mjauvin can you chime in on this?

mjauvin:
> @mjauvin can you chime in on this?

I didn't look deeply in this, but it should be possible to override the necessary code in `mlmarkdowneditor.js` in Winter.Translate

AIC-BV:
@mjauvin can you get me started so we can finish this?
The button event is defined in `markdowneditor.js` and so are the required methods (`toggleFullscreen()` & `togglePreview()`).
It does not exist in `mlmarkdowneditor.js`

AIC-BV:
I think I found it

PR #1004, opened 2023-11-07
Change 'save' to 'confirm'

AIC-BV:
The uploaded image isn't being saved, but the title and description are added.
This is confusing because it doesn't save/upload the image even though the button says 'save'.

My collegues tought it 'saves' the image as well but their work is gone because they didn't press the actual form submit button. What do you think?

Once the image is actually saved/uploaded, it does edit the title and you don't have to submit the form anymore.
But if the image isn't saved/uploaded, adding the title/description does NOT save/upload the image, which is why I think the button shouldn't be labeled 'save' (or it should save/upload the image)

### DIFF
--- a/modules/backend/formwidgets/fileupload/partials/_config_form.php
+++ b/modules/backend/formwidgets/fileupload/partials/_config_form.php
@@ -32,7 +32,7 @@
                 class="btn btn-primary"
                 data-request="<?= $this->getEventHandler('onSaveAttachmentConfig') ?>"
                 data-popup-load-indicator>
-                <?= e(trans('backend::lang.form.save')) ?>
+                <?= e(trans('backend::lang.form.confirm')) ?>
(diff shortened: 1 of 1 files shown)


LukeTowers:
@AIC-BV the form button does "save" the image, the image is uploaded when it is selected and a record for it exists in the database. What is actually not being "saved" is the relationship between the uploaded file and the current record being edited. If you would like for that to happen as soon as the image is uploaded because that works better for your workflow / users, then simply set `attachOnUpload: true` on the fileupload field configuration: https://wintercms.com/docs/v1.2/docs/backend/forms#file-upload.

Note that this still won't help if the parent record hasn't been created yet.

AIC-BV:
Thank you @LukeTowers sounds like exactly what we require!
Should've read the docs better 🤪

PR #1014, opened 2023-11-28
Remove importance="high"

AIC-BV:
This property does not exist (anymore?)

### DIFF
--- a/modules/backend/layouts/auth.php
+++ b/modules/backend/layouts/auth.php
@@ -29,13 +29,13 @@
         ?>
 
         <?php foreach ($styles as $style): ?>
-            <link href="<?= $style . '?v=' . $coreBuild; ?>" rel="stylesheet" importance="high">
-            <link href="<?= $style . '?v=' . $coreBuild; ?>" rel="preload" as="style" importance="high">
+            <link href="<?= $style . '?v=' . $coreBuild; ?>" rel="stylesheet">
+            <link href="<?= $style . '?v=' . $coreBuild; ?>" rel="preload" as="style">
         <?php endforeach; ?>
 
         <?php foreach ($scripts as $script): ?>
-            <script data-cfasync="false" src="<?= $script . '?v=' . $coreBuild; ?>" importance="high"></script>
-            <link href="<?= $script . '?v=' . $coreBuild; ?>" rel="preload" as="script" importance="high">
+            <script data-cfasync="false" src="<?= $script . '?v=' . $coreBuild; ?>"></script>
+            <link href="<?= $script . '?v=' . $coreBuild; ?>" rel="preload" as="script">
         <?php endforeach; ?>
 
         <?php if (!Config::get('cms.enableBackendServiceWorkers', false)): ?>
--- a/modules/backend/layouts/auth.php
+++ b/modules/backend/layouts/auth.php
@@ -30,7 +30,7 @@
 
         <?php foreach ($styles as $style) : ?>
             <link rel="preload" href="<?= $style . '?v=' . $coreBuild; ?>" as="style">
-            <link rel="stylesheet" href="<?= $style . '?v=' . $coreBuild; ?>"/>
+            <link rel="stylesheet" href="<?= $style . '?v=' . $coreBuild; ?>">
         <?php endforeach; ?>
 
         <?php foreach ($scripts as $script) : ?>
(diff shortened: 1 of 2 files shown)


LukeTowers:
It was originally added in https://github.com/octobercms/october/pull/4285 and it looks like there's one other place where it might still be present in the codebase, could you take a look?

Looks like the spec was renamed and adjusted since then and now exists as [Fetch Priority](https://web.dev/articles/fetch-priority#history). Would you be willing to take a look at that and make any necessary adjustments to retain the original intention of the PR while following the newest standard?

AIC-BV:
@LukeTowers I added fetchpriority how I interpreted it.
I opened a really large order which takes around **4 seconds** to load according to the network tab.
Using the modified code, it took significantly shorter: **2.5 seconds** to load.
Don't know if it is coincidence or not.

BUT
I kept getting the 'invalid security token' error very often while refreshing so this definitly needs some investigation of the maintainers to optimise this PR.

AIC-BV:
Also low support
https://caniuse.com/?search=fetchpriority

bennothommo:
Some thoughts on this:

- `async` attribute on scripts will introduce issues with loading scripts out of order on some browsers. This will cause issues for some widgets that need to load base scripts before others, such as the Inspector and Media finder.
- The best way of loading scripts in way that prevents stalling still remains loading them just before the closing `</body>` tag. I recall the only reason we haven't done that is because some widgets (and possibly user plguins) load JavaScript inline that requires certain scripts to be available immediately.
- My understanding is that we can use `rel="preload"` for all stylesheets without the need for the `rel="stylesheet"` line before it for all browsers now.

AIC-BV:
Must've done something wrong since its broken now 😅

AIC-BV:
Should work fine
https://caniuse.com/?search=preload

modules/backend/layouts/auth.php
```diff
@@ -28,14 +28,14 @@
...
-        <?php foreach ($styles as $style): ?>
-            <link href="<?= $style . '?v=' . $coreBuild; ?>" rel="stylesheet" importance="high">
-            <link href="<?= $style . '?v=' . $coreBuild; ?>" rel="preload" as="style" importance="high">
+        <?php foreach ($styles as $style) : ?>
+            <link rel="preload" href="<?= $style . '?v=' . $coreBuild; ?>" as="style">
+            <link rel="stylesheet" href="<?= $style . '?v=' . $coreBuild; ?>"/>
         <?php endforeach; ?>
 
-        <?php foreach ($scripts as $script): ?>
-            <script data-cfasync="false" src="<?= $script . '?v=' . $coreBuild; ?>" importance="high"></script>
```
LukeTowers, line 37: `data-cfasync` is still required AFAIK is it not? It's to prevent CloudFlare's Rocket Loader from breaking the backend.

AIC-BV: Doesn't seem to be doing anything, nothing is broken without it

AIC-BV: but can add if it makes it safer: https://developers.cloudflare.com/speed/optimization/content/rocket-loader/ignore-javascripts/

bennothommo: It will break sites for people who use CloudFlare Rocket Loader, but ideally, I would like for us to not have platform-specific fixes in core code and delegate it to a plugin, if it all possible.

LukeTowers: Normally we do handle it via a plugin (https://github.com/wintercms/wn-cloudflare-plugin/blob/main/Plugin.php#L43-L54) but these are different because they're manually included instead of going through the AssetMaker trait.

bennothommo: Is it worth us changing these to go through AssetMaker then?

LukeTowers: @bennothommo if you feel like doing that, sure 😉

AIC-BV:
👀

LukeTowers:
@bennothommo any thoughts on this?

bennothommo:
I think we can certainly remove the `importance` attribute - it's not being used. I still think we need to keep the `data-cfasync` for now until I pull my finger out and sort out some way of allowing a plugin to add it, in which case we can delegate it to the plugin.

AIC-BV:
Added the data-cfasync back to the line

LukeTowers:
Replaced by #1022, thanks for your help @AIC-BV!

PR #1019, opened 2023-12-08
Remove unnecessary text-align override on btn class

WebVPF:
I insert
```html
<button type="button" class="btn btn-primary btn-lg btn-block">Block level button</button>
<button type="button" class="btn btn-default btn-lg btn-block">Block level button</button>
```

Result:

Removed property overrides
```css
text-align: center;
```

The correct value of the property is written here: https://github.com/wintercms/winter/blob/78af271c2e34372f81bd481f3e21406b1fed8350/modules/system/assets/ui/storm.css#L475C16-L475C16

Less file: https://github.com/wintercms/winter/blob/78af271c2e34372f81bd481f3e21406b1fed8350/modules/system/assets/ui/less/button.base.less#L9

### DIFF
--- a/modules/system/assets/ui/less/button.less
+++ b/modules/system/assets/ui/less/button.less
@@ -17,7 +17,6 @@
 
 .btn {
     font-size: @font-size-base - 1;
-    text-align: left;
     outline: none !important;
     .box-shadow(~"inset 0 -2px 0 rgba(0,0,0,.15)");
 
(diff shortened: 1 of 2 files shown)


LukeTowers:
This seems like it could cause unintentional side effects, what have you done so far to ensure that it is really redundant and unnecessary? Are there any specific issues that the code in its current state is causing?

WebVPF:
@LukeTowers

Did you understand me correctly? I deleted:

```css
.btn {
    text-align: left;
}
```

The center alignment property remains (I indicated them in the links)

```css
.btn {
    text-align: center;
}
```

What problems might there be? The text inside the button should always be centered.

In the screenshot I inserted block level buttons (full width). Here is their code at the very bottom https://wintercms.com/docs/v1.2/ui/controls/button#sizes

Here are the same buttons in the Bootstrap documentation https://getbootstrap.com/docs/5.3/components/buttons/#block-buttons

After the fix, I looked through the entire interface and found no problems. I have installed the Test plugin and UI-Documentation with corrected visualization of all examples.

The media manager has a button to minimize the preview panel. Instead of the `btn` class, it uses the `btn-icon` class, which does not have text alignment.

PR #1020, opened 2023-12-12
Added validation check for colourpicker form widget

jaxwilko:
This PR adds validation for the colorpicker formwidget.

On save, the value will be tested against the configured format. If `all` is selected, then the colorpicker will validate against any supported format.

If somebody is using a custom format (not sure why they would but currently nothing prevents it), then the save value will be returned unmodified.

If the value fails validation, then an `ApplicationException` is thrown.


modules/backend/formwidgets/ColorPicker.php
```diff
@@ -59,6 +59,16 @@ class ColorPicker extends FormWidgetBase
...
     public $formats = 'hex';
 
+    /**
+     * @var array|string[] Patterns to validate colour string on save
+     */
+    protected array $validationPatterns = [
+        'cmyk' => '/^cmyk\((\d{1,2}\.?\d{0,2}%,? ?){4}\)$/',
+        'hex' => '/^#[\w\d]{6}$/',
+        'hsl' => '/^hsla\((\d{1,3}\.?\d{0,2}%?, ?){3}\d\.?\d{0,2}?\)$/',
+        'rgb' => '/^rgba\((\d{1,3}\.?\d{0,2}, ?){3}\d\.?\d{0,2}?\)$/'
```
LukeTowers:
```suggestion
        'rgb' => '/^rgba\((\d{1,3}\.?\d{0,2}, ?){3}\d\.?\d{0,2}?\)$/',
```

modules/backend/lang/en/lang.php
```diff
@@ -11,6 +11,7 @@
         'options_method_not_exists' => "The model class :model must define a method :method() returning options for the ':field' form field.",
         'options_static_method_invalid_value' => "The static method ':method()' on :class did not return a valid options array.",
         'colors_method_not_exists' => "The model class :model must define a method :method() returning html color HEX codes for the ':field' form field.",
+        'colors_invalid_input' => 'The color value supplied failed to meet expectations, please try again.',
```
LukeTowers:
```suggestion
        'colors_invalid_input' => 'The color value supplied is invalid, please try again.',
```

modules/backend/formwidgets/ColorPicker.php
```diff
@@ -244,6 +254,33 @@ protected function loadAssets()
...
+        if (!strlen($value)) {
+            return null;
+        }
+
+        switch (is_array($this->formats) ? 'all' : $this->formats) {
+            case 'cmyk':
+            case 'hex':
+            case 'hsl':
+            case 'rgb':
+                if (!preg_match($this->validationPatterns[$this->formats], $value)) {
```
LukeTowers, line 266: is the value already trimmed / do the regex patterns allow for untrimmed values?

jaxwilko: All the values should be trimmed by the frontend, the colorpicker doesn't allow extra spaces, so in theory yes, even if you type extra spaces they should be removed

PR #1021, opened 2023-12-13
Add breadcrumb translation to controller update.stub

damsfx:


damsfx:
@LukeTowers Is there a sticking point with this RP?  
I'm creating a lot of controllers at the moment, and I'm tired of having to edit their update.htm file every time.

LukeTowers:
Only sticking point was my life is a disaster and I'm only seeing this now 😂

PR #1031, opened 2024-01-08
PHP 8.3: Incrementor Fix

Denoder:
- This fixes the Error that occurs in php 8.3 when running the command `php artisan winter:up`:

```txt
Increment on type bool has no effect, this will change in the next major version of PHP
```


LukeTowers:
@Denoder looks like that's causing the tests to fail, could you double check the logic? Perhaps just explicitly coerce the value from a bool to an integer?

PR #1038, opened 2024-01-17
Increased caching time for combine resources

WebVPF:
Increased caching time to 1 year
https://developer.chrome.com/docs/lighthouse/performance/uses-long-cache-ttl#how_to_cache_static_resources_using_http_caching


WebVPF:
The Lighthouse service does not like the current caching time value

lighthouse

The value `604800` seconds was recommended by Google 7 years ago. Google currently recommends setting the value to `31536000` seconds.

LukeTowers (approved):

PR #1041, opened 2024-01-19
Add support for table-based lists that need sortable behavior.

lex0r:
At the moment only ol > li HTML structure is supported if one needs to implement reordering of sortable items. Having support for tbody > tr structure would came at a minimum cost of changes and is nearly supported by underlying code. I guess I don't need to prove that it's rather typical to sort tabular data and doing so shouldn't require developers to go for a completely new sort implementation on the FE.

### DIFF
--- a/modules/system/console/WinterVersion.php
+++ b/modules/system/console/WinterVersion.php
@@ -101,7 +101,7 @@ class WinterVersion extends \Winter\Storm\Console\Command
                 $this->line('');
                 $this->info('Files removed:');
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
@lex0r can you add a PR to the docs or some usage examples?

mjauvin:
I like this!

LukeTowers:
@mjauvin can you provide any additional context to this (i.e. docs or usage examples) or otherwise confirm that this works fine without breaking anything else?

PR #1044, opened 2024-01-22
Relation controller popup: fix size, add cssClass and allowDismiss

AIC-BV:
Add the possibility to define the popup size and its custom CSS classes.
You can now define these properties at the manage part of the config_relation.yaml:
```
myRelation:
    manage:
        size: giant
        cssClass: test
```
See all popup sizes at https://wintercms.com/docs/v1.2/ui/controls/popup#data-attributes

`size` already existed, but was hardcoded and not dynamically filled in.
`cssClass` is added.

## Release Notes
* **New Features**
  * Popups can now be configured with a custom CSS class and an option to allow/backdrop dismissal.
  * In popups with tracked unsaved changes, backdrop dismissal is blocked and users see an unsaved-changes warning instead.
  * Multi-view relation record popups can now be configured with popup size, custom styling, and backdrop dismissal behavior.
* **Documentation**
  * Added new popup data attribute documentation for CSS class and backdrop dismissal.
* **Localization / Bug Fixes**
  * Added an “unsaved changes” popup message across supported languages and corrected select locale text.

PR #1053, opened 2024-02-09
Close popup when clicking next to it

AIC-BV:
Close popup when clicking next to it

### DIFF
--- a/modules/system/assets/ui/js/popup.js
+++ b/modules/system/assets/ui/js/popup.js
@@ -215,6 +215,7 @@
             if (target.classList.contains('control-popup')) {
                 modal.hide()
                 $('.popup-backdrop').remove()
+                $(document.body).removeClass('modal-open')
             }
         });
         
--- a/modules/system/assets/ui/js/popup.js
+++ b/modules/system/assets/ui/js/popup.js
@@ -210,6 +210,14 @@
         if (this.options.zIndex !== null)
             modal.css('z-index', this.options.zIndex + 20)
 
+        modal.on('click', function(e) {
+            const target = e.target;
+            if (target.classList.contains('control-popup')) {
+                modal.hide()
+                $('.popup-backdrop').remove()
+            }
+        });
+        
         return modal.append(modalDialog.append(modalContent))
     }
 


AIC-BV:
Can probably be optimised by reusing the right function but I couldn't find how to call it without having undefined error

LukeTowers:
@AIC-BV would you be able to provide a screen recording of the new behaviour?

bennothommo:
@AIC-BV does this functionality make it if you click on the translucent overlay, it closes the modal? I'm not a big fan of that at all - too many times, people mis-click and miss something important (or lose a completed form if the modal contains one).

AIC-BV:
> @AIC-BV does this functionality make it if you click on the translucent overlay, it closes the modal? I'm not a big fan of that at all - too many times, people mis-click and miss something important (or lose a completed form if the modal contains one).

That is the case! For me it is good UX

AIC-BV:
@LukeTowers 

https://github.com/wintercms/winter/assets/89913092/cb559089-23e6-437c-8413-310612b714ad

AIC-BV:
It is a read only popup relation and it is very annoying if you can't click next to it 😆

bennothommo:
@AIC-BV 

> That is the case! For me it is good UX

Apologies, but I strongly disagree.

I would begrudgingly allow it for "message" modals, where an action has already occurred and the modal is simply letting the user know it's done - albeit, a flash message would generally be better in this case and those can already be easily dismissed. Definitely not for any modal that requires user interaction or attention, ie. a modal form, a confirmation modal where the user has to Yes or No, or an error message.

I would only accept this PR if such modal behaviour were optional (ie. could be defined by configuration) and by default, it was disabled.

AIC-BV:
@bennothommo 
If acceptable I can adjust https://github.com/wintercms/winter/pull/1044 to add variable "closeModalWhenClickingNextToIt". Could do with a better name.

And then update this PR to
```
if (this.options.closeModalWhenClickingNextToIt) {
    modal.on('click', function(e) {
        const target = e.target;
        if (target.classList.contains('control-popup')) {
            modal.hide()
            $('.popup-backdrop').remove()
            $(document.body).removeClass('modal-open')
        }
    });
}
 ```
 Let me know

bennothommo:
That'd be all good. Maybe `allowDismiss` as the option? Or `closeOutside`?

AIC-BV:
Added in this PR
https://github.com/wintercms/winter/pull/1044

PR #1064, opened 2024-02-26
Add disabled state styling for switch fields

AIC-BV:
Currently has 1 opacity. Needs 0.5 for default WinterCMS disabled state

Before
image

After
image


bennothommo:
@AIC-BV @LukeTowers does this also switch the cursor to a "disabled" cursor? If not, I think it should for usability. :)

AIC-BV:
No it doesn't @bennothommo, good call!
See https://github.com/wintercms/winter/pull/1078

PR #1066, opened 2024-02-26
Make `twig.environment.cms` singleton

parallels999:
https://github.com/wintercms/winter/blob/d9ff4e0fca3a92b0bae995637130c33abd200370/modules/system/ServiceProvider.php#L322

https://github.com/wintercms/winter/blob/d9ff4e0fca3a92b0bae995637130c33abd200370/modules/system/ServiceProvider.php#L327

Is there any reason why it is not a singleton like the other environments?

### DIFF
--- a/modules/cms/ServiceProvider.php
+++ b/modules/cms/ServiceProvider.php
@@ -90,7 +90,7 @@ class ServiceProvider extends ModuleServiceProvider
     protected function registerTwigParser()
     {
         // Register CMS Twig environment
-        $this->app->bind('twig.environment.cms', function ($app) {
+        $this->app->singleton('twig.environment.cms', function ($app) {
             // Load Twig options
             $useCache = !Config::get('cms.twigNoCache');
             $isDebugMode = Config::get('app.debug', false);


LukeTowers:
@parallels999 why close this? I don't remember off the top of my head but I think there was a reason.

parallels999:
>If you need to use the CMS Controller's Twig instance then you should call getTwig() directly on the controller instance instead of relying upon twig.environment.cms being a singleton.
It being a singleton causes issues if for some reason you want to render another route in the same request as the second call to Controller->run() will pollute the variables of the first.

Because it seems there was a reason
https://github.com/wintercms/winter/commit/7dd0eb1b7fb4d9b0c0ea3d9816630a7f1d7f05f6

parallels999:
Maybe `CmsTwigLoader` could be a singleton
https://github.com/wintercms/winter/blob/d9ff4e0fca3a92b0bae995637130c33abd200370/modules/cms/ServiceProvider.php#L124

parallels999:
Also, shouldn't `Config::get` facade be replaced with `$app['config']->get`?
https://github.com/wintercms/winter/blob/d9ff4e0fca3a92b0bae995637130c33abd200370/modules/cms/ServiceProvider.php#L93-L99

LukeTowers:
Sure, feel free to

PR #1069, opened 2024-02-27
Call model mutators before setting attributes

### LINKED ISSUE #1047: Model mutators not working using SettingsModel behavior
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

Hi,

### Steps to replicate

1. Create a Settings model.
3. Read the persisted value.

### Workaround

Example:
```php
{
}
```

mjauvin:
Replaced by #1071

### DIFF
--- a/modules/system/behaviors/SettingsModel.php
+++ b/modules/system/behaviors/SettingsModel.php
@@ -3,9 +3,10 @@
 use App;
 use Artisan;
 use Cache;
-use Log;
 use Exception;
+use Log;
 use Illuminate\Database\QueryException;
+use Str;
 use System\Classes\ModelBehavior;
 
 /**
@@ -178,6 +179,11 @@ class SettingsModel extends ModelBehavior
     {
         if ($this->isKeyAllowed($key)) {
             return;
+        }
+
+        if ($this->model->hasSetMutator($key)) {
+            $mutator = 'set'.Str::studly($key).'Attribute';
+            $value = $this->model->{$mutator}($value);
         }
 
         $this->fieldValues[$key] = $value;


mjauvin:
@LukeTowers this fixes the issue for that particular case, but I found these which would cause similar issues for all models (including Halcyon models):

https://github.com/wintercms/storm/blob/develop/src/Database/Model.php#L1379-L1395
https://github.com/wintercms/storm/blob/develop/src/Halcyon/Model.php#L782-L798

Note: I think we should replace the value returned by the mutator (or parent::setAttribute() which does call mutators if they exist) so that the `model.setAttribute` event handler receive the mutated value.

mjauvin:
@LukeTowers this is what I am suggesting:
```diff
diff --git a/src/Database/Model.php b/src/Database/Model.php
index 40390596..743e522f 100644
--- a/src/Database/Model.php
+++ b/src/Database/Model.php
@@ -1376,7 +1376,9 @@ public function setAttribute($key, $value)
             $value = trim($value);
         }
 
-        $result = parent::setAttribute($key, $value);
+        if ($_value = parent::setAttribute($key, $value)) {
+            $value = $_value;
+        }
 
         /**
          * @event model.setAttribute
diff --git a/src/Halcyon/Model.php b/src/Halcyon/Model.php
index f51c3ac4..58eeda1e 100644
--- a/src/Halcyon/Model.php
+++ b/src/Halcyon/Model.php
@@ -785,7 +785,9 @@ public function setAttribute($key, $value)
             // 'model.setAttribute' event, and then no third party plugins will be able to implement any kind of
             // post processing logic when an attribute is set with explicit mutators. Returning from the mutator
             // call will also break method chaining as intended by returning `$this` at the end of this method.
-            $this->{$method}($value);
+            if ($_value = $this->{$method}($value)) {
+                $value = $_value;
+            }
         }
         else {
             $this->attributes[$key] = $value;
```

PR #1073, opened 2024-03-01
Restore original md filter behavior for null content

mjauvin:
This used to work:
```twig
{% set content = null %}

{{ content | md }}
```

Since Markdown::parse() method signature was changed we need to sanitize the content passed to `md`:
```twig
{% set content = null %}
{{ content | default('') | md }}
```
This potentially broke a lot of legacy templates.

This PR restore the previous behavior.

### DIFF
--- a/modules/system/ServiceProvider.php
+++ b/modules/system/ServiceProvider.php
@@ -218,9 +218,15 @@ class ServiceProvider extends ModuleServiceProvider
                 'studly'         => ['Str', 'studly'],
                 'trans'          => ['Lang', 'get'],
                 'transchoice'    => ['Lang', 'choice'],
-                'md'             => ['Markdown', 'parse'],
-                'md_safe'        => ['Markdown', 'parseSafe'],
-                'md_line'        => ['Markdown', 'parseLine'],
+                'md'             => function($value) {
+                    return $value ? Markdown::parse($value) : null;
+                },
+                'md_safe'        => function($value) {
+                    return $value ? Markdown::parseSafe($value) : null;
+                },
+                'md_line'        => function($value) {
+                    return $value ? Markdown::parseLine($value) : null;
+                },
                 'time_since'     => ['System\Helpers\DateTime', 'timeSince'],
                 'time_tense'     => ['System\Helpers\DateTime', 'timeTense'],
             ]);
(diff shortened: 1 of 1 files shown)


mjauvin:
Not sure if the filters should return null or empty string though.

modules/system/ServiceProvider.php
```diff
@@ -218,9 +218,15 @@ protected function registerMarkupTags()
                 'studly'         => ['Str', 'studly'],
                 'trans'          => ['Lang', 'get'],
                 'transchoice'    => ['Lang', 'choice'],
-                'md'             => ['Markdown', 'parse'],
-                'md_safe'        => ['Markdown', 'parseSafe'],
-                'md_line'        => ['Markdown', 'parseLine'],
+                'md'             => function($value) {
+                    return $value ? Markdown::parse($value) : null;
```
bennothommo: @mjauvin Maybe something like this? `null` would fail `is_string`, and it would ensure that we're only passing strings to `parse()`.
```suggestion
                    return (is_string($value) && $value !== '') ? Markdown::parse($value) : '';
```

modules/system/ServiceProvider.php
```diff
@@ -218,9 +218,15 @@ protected function registerMarkupTags()
...
-                'md_line'        => ['Markdown', 'parseLine'],
+                'md'             => function($value) {
+                    return (is_string($value) && $value !== '') ? Markdown::parse($value) : '';
+                },
+                'md_safe'        => function($value) {
+                    return (is_string($value) && $value !== '') ? Markdown::parseSafe($value) : '';
+                },
+                'md_line'        => function($value) {
+                    return (is_string($value) && $value !== '') ? Markdown::parseLine($value) : '';
+                },
```
bennothommo:
```suggestion
                'md'             => function ($value) {
                    return (is_string($value) && $value !== '') ? Markdown::parse($value) : '';
                },
                'md_safe'        => function ($value) {
                    return (is_string($value) && $value !== '') ? Markdown::parseSafe($value) : '';
                },
                'md_line'        => function ($value) {
                    return (is_string($value) && $value !== '') ? Markdown::parseLine($value) : '';
                },
```

bennothommo (approved):

PR #1077, opened 2024-03-02
Do not activate .env.example on project creation

### LINKED ISSUE #1060: Error with the database when installing version 1.2.4
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

SQLite

### Plugins installed

_No response_

### Issue description

After installing v1.2.4, when accessing the Backend, the error is "Database missing".

In the console after executing ```php artisan winter:up``` error:

```
In Connection.php line 760:                                                                                                                                                     
could not find driver (SQL: select * from information_schema.tables where table_schema = database and table_name = migrations and table_type = 'BASE TABLE')  
In Exception.php line 18:
could not find driver  
In PDOConnection.php line 40:
could not find driver
```

I carry out the installation as always, there have never been any problems before. Tried it on several independent hosting providers.

In the new version, I noticed that it is now necessary to specify the path to the database. I tried in the root (by default) and the storage folder (as was the case in older versions).

To work around the bug, I install v1.2.2 first and then update. Everything works without errors.

### Steps to replicate

Just install a new project with a new version :)

### Workaround

To work around the bug, I install v1.2.2 first and then update. Everything works without errors.

mjauvin:
Fix #1060

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,9 +46,6 @@
         "dms/phpunit-arraysubset-asserts": "^0.1.0|^0.2.1"
     },
     "scripts": {
-        "post-root-package-install": [
-            "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
-        ],
         "post-create-project-cmd": [
             "@php artisan key:generate"
         ],


bennothommo:
Waiting for [confirmation](https://github.com/wintercms/winter/issues/1060#issuecomment-1973469259) that this doesn't break the workflow for using Winter in Laravel Sail

PR #1080, opened 2024-03-05
Allow mutators for Settings models and improve API

### LINKED ISSUE #1047: Model mutators not working using SettingsModel behavior

bennothommo:
Fixes #1047. Replaces #1071.

- Use `getAttribute` method in the `model.setAttribute` listener within the `SettingsModel` behavior to retrieve the (possibly) mutated value before saving into settings array.
- Ensure that default values populated by `initSettingsData` method in a Settings model are populated into settings array if not yet persisted in the DB, so that they are accessible via model properties.
- Allow values to be retrieved as model properties at any time.

### DIFF
--- a/modules/system/behaviors/SettingsModel.php
+++ b/modules/system/behaviors/SettingsModel.php
@@ -7,6 +7,7 @@ use Log;
 use Exception;
 use Illuminate\Database\QueryException;
 use System\Classes\ModelBehavior;
+use Winter\Storm\Database\Model;
 
 /**
  * Settings model extension
@@ -81,7 +82,7 @@ class SettingsModel extends ModelBehavior
     /**
      * Create an instance of the settings model, intended as a static method
      */
-    public function instance()
+    public function instance(): Model
     {
         if (isset(self::$instances[$this->recordCode])) {
             return self::$instances[$this->recordCode];
@@ -89,6 +90,7 @@ class SettingsModel extends ModelBehavior
 
         if (!$item = $this->getSettingsRecord()) {
             $this->model->initSettingsData();
+            $this->afterModelFetch();
             $item = $this->model;
         }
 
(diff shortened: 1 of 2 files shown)


bennothommo:
@goldmont would you be willing to test this PR as an alternate fix for #1047? I've added a test case for your scenario, but happy to hear if it works out in the wild.

modules/system/behaviors/SettingsModel.php
```diff
@@ -174,12 +176,14 @@ public function getSettingsValue($key, $default = null)
     /**
      * Set a single setting value, if allowed.
      */
-    public function setSettingsValue($key, $value)
+    public function setSettingsValue($key)
```
LukeTowers: This change confused for me for a bit before I realized that is method is somewhat performing double duty. The first use-case is the public api of setSettingsValue($key, $value), but the second use-case (the one addressed by this PR and the other PRs) is to sync the model data to the internal fieldValues array whenever setAttributes() is called. 

@bennothommo what do you think about changing it so that the model.setAttribute event listener is instead a protected method `syncModelAttribute($key)` that then does `$this->setSettingsValue($key, $this->model->getAttribute($key))` and leave this method unmodified?

bennothommo: @LukeTowers I think we should come to a conclusion on what we're going to be doing about the actual event. If we either fix up the event or add the third parameter, then I'll have to change this anyway - fixing up the event would mean we don't need this PR at all (beyond maybe the additional `afterModelFetch` calls to ensure everything works as properties).

LukeTowers: @bennothommo perhaps we can get on a call about this, I'm still not really understanding the problem with the event itself, would probably be easier to talk it through when you have a minute 😄

goldmont: Hi,

Let me know what and when to test. Thanks for your efforts.

LukeTowers: @goldmont is this still an issue for you?

goldmont: Hi,

I "solved" by using the Encryption trait.

PR #1083, opened 2024-03-11
Add "Duplicate" feature to mediamanager

damsfx:
Add a button for duplicating files or folders in the Media Manager menu bar.

### Single item selection: 
If a file or folder is selected, the user is prompted for the new name of the duplicated file/folder.

### Multiple item selection:
If multiple files and/or folders are selected, a warning is displayed to the user that the files/folders will be duplicated with an automatically generated name.

### Automatically generated names : 
"-x" is added to the name of the item to be duplicated, where x is a numeric counter.
The folder where the item is located is scanned for occurrences with identical names.

For example, `myfile.pdf` will give `myfile-1.pdf`.
If `myfile-1.pdf` is already present, then the new name will be `myfile-2.pdf`.

Required: https://github.com/wintercms/storm/pull/186


modules/system/classes/MediaLibrary.php
```diff
@@ -834,4 +896,45 @@ protected function generateRandomTmpFolderName($location)
...
     }
+
+    /**
+     * Generates a incremental file name based
+     * on the existing files in the same folder.
+     *
+     * @param string $path Specifies a file path to check.
+     * @return string The generated file name.
+     */
+    public function generateIncrementedFileName($path): string
```
LukeTowers, line 909: There's a Str::unique() helper, and the original PR for it proposed a File::unique() helper as well. Any thoughts on either using the str helper or re-adding the proposed `File::unique()` helper: https://github.com/wintercms/storm/pull/114?

damsfx: @LukeTowers I've tested with `Str::unique` for directory names ... it seems correct.
However, for files it doesn't work, I think the `File::unique()` method is required.

Is there any particular reason why it was removed from the PR?

LukeTowers: It wasn't being used at the time, the Str::unique() was in an internal project, but at the time the File::unique() wasn't being used at all.

damsfx: I can update the code for this PR if the feature makes a comeback in Winter's core.

LukeTowers: Do you mind submitting a PR to Storm to add it back in?

damsfx: Will try to find the time this end's week.

LukeTowers: @damsfx ping 😉

damsfx: @LukeTowers pong 😉

LukeTowers:
@damsfx can you add some screenshots?

damsfx:
@LukeTowers 
[WinterCMS_Mediamanager_Duplicate.webm](https://github.com/user-attachments/assets/35b14224-28fb-414d-a631-cdb281e23803)
<video src="https://github.com/user-attachments/assets/35b14224-28fb-414d-a631-cdb281e23803" width="100%" />

**New button :** 
image

**Popup to give new name (with suggested one) :** 
image

Popup for multiple selected files or folders : 
image


**Automatically generated names :**

"_x" is added to the name of the item to be duplicated, where x is a numeric counter.
The folder where the item is located is scanned for occurrences with identical names.

For example, `myfile.pdf` will give `myfile_1.pdf`.
If `myfile_1.pdf` is already present, then the new name will be `myfile_2.pdf`.
If `myfile_1.pdf` and `myfile_4.pdf` is already present, then the new name will be `myfile_5.pdf`.

LukeTowers:
Fantastic @damsfx! What happens if they rename the proposed name in the duplicate file popup to the original name (ignoring the suggested name)?

damsfx:
@LukeTowers the golden question! 
I don't remember to have make that test.
It use storage facade copy method witch is just a return of the PHP native one.

The php copy manual says:
> Warning
If the destination file already exists, it will be overwritten.

So I think it may be the same here.

Test needed and if it's the case, maybe ask the user to confirm the overwrite?

damsfx:
@LukeTowers I can confirm that a file with the same name is overwritten.

What solution do you plan to use to solve this problem?  
An alert informing the user/requesting his confirmation, a return with no action other than a flash message?

LukeTowers:
@damsfx I would say probably start with just throwing an ApplicationException which will popup the error. I'm not sure how we're currently handling it in the rename popup and / or on file upload, maybe take a quick look there and then we can standardize across all three?

damsfx:
@LukeTowers At this time, both upload and rename actions simply replace an existing file, no errors, no warnings.

LukeTowers:
@damsfx thats what I thought, I think there was an issue at some point to fix that by auto renaming to the suffixed form, any thoughts on the approach that we should take for all of the scenarios so that we can standardize on that? I think the options are:

- Do nothing and silently allow the replacement 
- throw an applicationexception refusing to overwrite an existing file
- allow the file put but force a rename to an auto suffixed filename
- present a dialog offering the choice to the user

The last option is slightly more difficult to implement but I think is the best option. Do you have any other thoughts or perhaps any options I missed?

Also would like to get input from any of the other maintainers if they're available (@bennothommo, @mjauvin, @jaxwilko).

mjauvin:
I would definitely prefer the following option:

allow the file put but force a rename to an auto suffixed filename

bennothommo:
I agree with @mjauvin. It's the least intrusive and destructive. If someone wants to be clever with their naming, they will most certainly be skilled enough to find the file with an additional suffix.

damsfx:
It's hard to say ... 

Allowing an existing file to be overwritten (without warning) can be an advantage, but it can also be a risk.

Having a warning before any file overwrite action would be a bonus, but perhaps this behaviour should be configurable in the backend settings?

LukeTowers:
The way it's typically handled in OS level file managers is to present a dialog that asks the user if they want to Replace, Keep Both (auto rename), or Cancel. That would be my preferred way to handle it here, but I'm also fine with defaulting to Keep Both if that's easier to implement to start with.

damsfx:
@LukeTowers , @mjauvin , @bennothommo 
Would you prefer this PR to include the changes so that the existing files are no longer overwritten by the uploaded/renamed/duplicated files, or would you prefer this to be done at a later date with a new PR?

mjauvin:
I prefer to include it in this PR

AIC-BV:
looking GREAT! This will be used a lot by my team creating the content

damsfx:
@AIC-BV  As a user likely to use this function a lot, what are your feelings about overwriting files already present with the same name (see previous messages)?

AIC-BV:
> @AIC-BV As a user likely to use this function a lot, what are your feelings about overwriting files already present with the same name (see previous messages)?

It can work both ways but I think that the correct way, or the expected UX default behaviour, is to not automatically overwrite it.
Preventing accidental data loss is the most important.
Overwriting it is a risk, like you said above, and I think you should avoid risk if you can choose to.

When you do this in Windows File Explorer it automatically changes the name to 'filename - Copy' which makes sense to me. Just like you did with the _1, _2, _3, ...
It is up to the content creater to name the files correct, or remove the duplicate ones.

For me it doesn't need a popup, just the 'version' behind the string.
A popup can work too but that is not the default file explorer behaviour, but perhaps its different on Mac

modules/system/classes/MediaLibrary.php
```diff
@@ -834,4 +898,51 @@ protected function generateRandomTmpFolderName($location)
...
+        $sameFilesInFolder = Arr::map(array_filter(
+            $this->getStorageDisk()->files($dirName),
+            function ($file) use ($pathInfos) {
+                return preg_match('/'. $pathInfos['filename'] .'(_(\d*))?\.'. $pathInfos['extension'] .'$/U', $file);
+            }
+        ), function ($value) use ($dirName) {
+            return str_replace($dirName .'/', '', '/'. $value);
+        });
+
+        return File::unique($pathInfos['basename'], $sameFilesInFolder);
```
LukeTowers: Is the filtering necessary here or can we just send the list of filenames in the folder unfiltered?

damsfx: You're right, filtering on file name and extension was not required.

modules/system/classes/MediaLibrary.php
```diff
@@ -834,4 +898,49 @@ protected function generateRandomTmpFolderName($location)
...
+     */
+    public function generateIncrementedFileName($path): string
+    {
+        $pathInfos = pathinfo($path);
+        $dirName = dirname($this->getMediaPath($path));
+
+        $sameFilesInFolder = Arr::map(
+            $this->getStorageDisk()->files($dirName),
+            function ($value) use ($dirName) {
+                return str_replace($dirName .'/', '', '/'. $value);
```
LukeTowers: We should probably make use of DIRECTORY_SEPARATOR for these operations. Also, are you able to add unit tests for the new methods?

damsfx: @LukeTowers DIRECTORY_SEPARATOR work nice on Linux but give me a backslash on Windows.
`$this->getStorageDisk()->files($dirName),` give me an array with slashes : 
```php
array:3 [▼
  0 => "media/test-files/file_example_LESS.less"
  1 => "media/test-files/file_example_PDF.pdf"
  2 => "media/test-files/file_example_PDF_1.pdf"
]
```
`$dirName . '/'` = `/media/test-files/` while `$dirName . DIRECTORY_SEPARATOR` = `/media/test-files\`.

That's why I use hardcoded `'/'` because I need to get only files names.

damsfx: @LukeTowers Another solution can be to use `basename` !?

```php
$sameFilesInFolder = Arr::map(
    $this->getStorageDisk()->files($dirName),
    function ($path) {
        return basename($path);
    }
);
```
In this case it's work for both `generateIncrementedFileName` and `generateIncrementedFolderName` methods.

PR #1084, opened 2024-03-21
Remove encryption setting on backend mail settings page & config

mjauvin:
Remove smtp encryption setting.


bennothommo:
@mjauvin  I'm not sure I agree with the port changes. Port 25 is still the default port for unencrypted SMTP - even if most people are using other ports these days. While we obviously want to "encourage" people to use encryption, it's still up to the developer and their environment.

The other changes look fine to me though.

mjauvin:
> @mjauvin  I'm not sure I agree with the port changes. Port 25 is still the default port for unencrypted SMTP - even if most people are using other ports these days. While we obviously want to "encourage" people to use encryption, it's still up to the developer and their environment.
> 
> The other changes look fine to me though.

The submission port is 587 (look in /etc/services). Port 25 is used by mail servers, not by mail clients.

mjauvin:
Read this for a very good explanation:
https://www.mailgun.com/blog/email/which-smtp-port-understanding-ports-25-465-587/#chapter-3

josephcrowell:
Outlook, Thunderbird and most other clients that I've seen default to port 25 for unencrypted connections although most servers have 587 open for both these days.

mjauvin:
Also, it looks like we're using 'tls' & 'ssl' for the encryption but Laravel only ever use 'tls':

ref. https://github.com/laravel/framework/blob/11.x/src/Illuminate/Mail/MailManager.php#L177-L199

I think I'll change the encryption field to a checkbox, and leave the port number untouched.

For BC, I'll convert the encryption option to 'tls' if it was 'ssl'/'tls' and leave it unset otherwise.

Thoughts?

mjauvin:
Ok, so I removed `ssl` encryption as this is irrelevant (it gets converted to `tls` if it was previously set).

Port if left untouched, allowing full configurability.

Note: if port is 587, whether encryption is set or not, the Mail client (symfony here) will use STARTTLS automatically if it's listed in the mail server's capabilities.

below is what symfony does in `Transport/Smtp/EsmtpTransport.php`:
```php
if (!$stream->isTLS() && \defined('OPENSSL_VERSION_NUMBER') && \array_key_exists('STARTTLS', $this->capabilities)) {
   $this->executeCommand("STARTTLS\r\n", [220]);
   ...
}
```

Arguably, we COULD force Encryption when port 465 is used as this is usually mandatory.

josephcrowell:
465 is usually SSL only, not starttls. So if you wanted a reason to assume/turn on SSL, that would be a good one.

bennothommo:
@josephcrowell is correct - port 465 is explicitly an SSL port for SMTP servers, and requires an SSL connection, although some servers do allow TLS to be used (hence the check for the capabilities).

Nonetheless, I think it's reasonable to assume that if 465 is used, it will be encrypted, so I'm fine with it being forced through the form.

mjauvin:
> Nonetheless, I think it's reasonable to assume that if 465 is used, it will be encrypted, so I'm fine with it being forced through the form.

What are you saying, this PR removes any forced rules, giving more flexibility for any possible setup.

LukeTowers:
What's the status of this?

mjauvin:
If you guys are OK with dropping `filterFields()` (i.e. not auto-select port based on encryption), this is complete.

mjauvin:
Notes: Laravel Mailer/Transport/EsmtpTransport does the following:
```php
if (null === $tls) {
    if (465 === $port) {
        $tls = true;
    } else {
        $tls = \defined('OPENSSL_VERSION_NUMBER') && 0 === $port && 'localhost' !== $host;
    }
}
```
Also, no matter what encryption we set, Laravel does this for smtp transport:
```php
        if (! $scheme) {
            $scheme = ! empty($config['encryption']) && $config['encryption'] === 'tls'
                ? (($config['port'] == 465) ? 'smtps' : 'smtp')
                : '';
        }   
```
And symfony does this in the EsmtpTransportFactory:
```php
final class EsmtpTransportFactory extends AbstractTransportFactory
{
    public function create(Dsn $dsn): TransportInterface
    {   
        $tls = 'smtps' === $dsn->getScheme() ? true : null;
        $port = $dsn->getPort(0);
        $host = $dsn->getHost();
        
        $transport = new EsmtpTransport($host, $port, $tls, $this->dispatcher, $this->logger);
    ...
}
```
So encryption is only forced for port 465. Encryption will be automatically enabled using STARTTLS for other ports (e.g. 25 and 587)  if available from the mail server CAPABILITIES.

Here
```php
if (!$stream->isTLS() && \defined('OPENSSL_VERSION_NUMBER') && \array_key_exists('STARTTLS', $this->capabilities)) {
    $this->executeCommand("STARTTLS\r\n", [220]);
    ...
}
```

mjauvin:
@bennothommo @LukeTowers All the above mean we should drop the encryption config for SMTP transport.

modules/system/models/MailSetting.php
```diff
@@ -117,46 +111,19 @@ public static function applyConfigValues()
...
 
-    /**
-     * @return array smtp_encryption options values
-     */
-    public function getSmtpEncryptionOptions()
-    {
-        return [
-            '' => 'system::lang.mail.smtp_encryption_none',
-            'tls' => 'system::lang.mail.smtp_encryption_tls',
-            'ssl' => 'system::lang.mail.smtp_encryption_ssl',
```
LukeTowers, line 128: We should drop these keys as well if they're not being used anymore.

mjauvin: Done

PR #1087, opened 2024-03-24
Add --no-progress and --json options to mix:compile and mix:list

Nimdoc:
- Add --no-progress option to mix:compile

This will suppress the mix progress output while allowing the webpack progress to be output. 

- Add --json to mix:list

This outputs the mix package information in a json format. Also instead of "yes" and "no" for the "active" property, with the --json option the values are true and false.

These additions will be useful for scripts or automation tools that need to parse information about mix packages and information for when they're compiled.

### DIFF
--- a/modules/system/console/MixCompile.php
+++ b/modules/system/console/MixCompile.php
@@ -227,8 +227,8 @@ class MixCompile extends Command
         $fixture = File::get(__DIR__ . '/fixtures/mix.webpack.js.fixture');
 
         $config = str_replace(
-            ['%base%', '%notificationInject%', '%mixConfigPath%', '%pluginsPath%', '%appPath%', '%silent%'],
-            [addslashes($basePath), 'mix._api.disableNotifications();', addslashes($mixJsPath), addslashes(plugins_path()), addslashes(base_path()), (int) ($this->option('silent') || $this->option('no-progress'))],
+            ['%base%', '%notificationInject%', '%mixConfigPath%', '%pluginsPath%', '%appPath%', '%silent%', '%noProgress%'],
+            [addslashes($basePath), 'mix._api.disableNotifications();', addslashes($mixJsPath), addslashes(plugins_path()), addslashes(base_path()), (int) $this->option('silent'), (int) $this->option('no-progress')],
             $fixture
         );
 
--- a/modules/system/console/MixCompile.php
+++ b/modules/system/console/MixCompile.php
@@ -102,7 +102,7 @@ class MixCompile extends Command
                 continue;
             }
 
-            if (!$this->option('no-progress')) {
+            if (!$this->option('silent')) {
                 $this->info(sprintf('Mixing package "%s"', $name));
             }
 
--- a/modules/system/console/MixCompile.php
+++ b/modules/system/console/MixCompile.php
@@ -22,7 +22,8 @@ class MixCompile extends Command
         {--s|silent : Silent mode}
         {--e|stop-on-error : Exit once an error is encountered}
         {--m|manifest= : Defines package.json to use for compile}
-        {--p|package=* : Defines one or more packages to compile}';
+        {--p|package=* : Defines one or more packages to compile}
+        {--no-progress : Do not show mix progress}';
 
     /**
      * @var string The console command description.
@@ -101,7 +102,9 @@ class MixCompile extends Command
                 continue;
             }
 
-            $this->info(sprintf('Mixing package "%s"', $name));
+            if (!$this->option('no-progress')) {
+                $this->info(sprintf('Mixing package "%s"', $name));
+            }
 
             $exitCode = $this->mixPackage(base_path($relativeMixJsPath));
 
@@ -225,7 +228,7 @@ class MixCompile extends Command
 
         $config = str_replace(
             ['%base%', '%notificationInject%', '%mixConfigPath%', '%pluginsPath%', '%appPath%', '%silent%'],
-            [addslashes($basePath), 'mix._api.disableNotifications();', addslashes($mixJsPath), addslashes(plugins_path()), addslashes(base_path()), (int) $this->option('silent')],
+            [addslashes($basePath), 'mix._api.disableNotifications();', addslashes($mixJsPath), addslashes(plugins_path()), addslashes(base_path()), (int) ($this->option('silent') || $this->option('no-progress'))],
             $fixture
         );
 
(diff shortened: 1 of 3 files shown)


modules/system/console/MixCompile.php
```diff
@@ -101,7 +102,9 @@ public function handle(): int
                 continue;
             }
 
-            $this->info(sprintf('Mixing package "%s"', $name));
+            if (!$this->option('no-progress')) {
+                $this->info(sprintf('Mixing package "%s"', $name));
+            }
```
bennothommo: Perhaps this should be hidden only when `silent` mode is defined, but should still show when `no-progress` is defined. Silent implies no (or very limited) output, but no progress just means we hide any progress bars from showing.

```suggestion
            if (!$this->option('silent')) {
                $this->info(sprintf('Mixing package "%s"', $name));
            }
```

modules/system/console/MixList.php
```diff
@@ -35,16 +34,13 @@ public function handle(): int
...
-        $this->line('');
-
         $errors = [];
 
         $rows = [];
         foreach ($packages as $name => $package) {
             $rows[] = [
                 'name' => $name,
-                'active' => $package['ignored'] ?? false ? '<fg=red>No</>' : '<info>Yes</info>',
+                'active' => $this->option('json') ? !$package['ignored'] : ($package['ignored'] ? '<fg=red>No</>' : '<info>Yes</info>'),
```
bennothommo: I'm not a fan of double-ternary usage - it makes it hard to visually parse what is happening. Perhaps we just change this to a boolean, and we'll control the output below.

```suggestion
                'active' => !$package['ignored'],
```

modules/system/console/MixList.php
```diff
@@ -54,9 +50,15 @@ public function handle(): int
             }
         }
 
-        $this->table(['Name', 'Active', 'Path', 'Configuration'], $rows);
-
-        $this->line('');
+        if($this->option('json')) {
+            echo json_encode($rows, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES);
```
bennothommo: This should still go through the Laravel output methods for compatibility.

```suggestion
        if ($this->option('json')) {
            $this->line(json_encode($rows, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES));
```

modules/system/console/MixList.php
```diff
@@ -54,9 +50,15 @@ public function handle(): int
...
+        } else {
+            $this->line('');
+            $this->info('Packages registered:');
+            $this->line('');
+            $this->table(['Name', 'Active', 'Path', 'Configuration'], $rows);
```
bennothommo: Following from above, we can apply the Yes / No styling directly here:

```suggestion
            $this->table(['Name', 'Active', 'Path', 'Configuration'], array_map(function ($row) {
                $row['active'] = ($row['active']) ? '<info>Yes</info>' : '<fg=red>No</>';
                return $row;
            }, $rows));
```

bennothommo (changes requested):

Nimdoc:
@bennothommo  Those are all good suggestions and I've made a commit with them all together. 

I was a little worried about the silent option interfering with the output of webpack, because adding the silent option to mix:compile also adds `--stats=none` to the webpack command, and there are situations where the webpack output is wanted and not the mix output. But with `php artisan mix:compile --silent -- stats=normal`, the latter option overwrites the preceding one. So we end up with an overall webpack command like

```bash
/home/tom/projects/winter/node_modules/.bin/webpack build --stats=none --config=/home/tom/projects/winter/modules/backend/mix.webpack.js --stats=normal --json
```

And it all works out.

bennothommo:
@Nimdoc thanks for making those changes.

Conventionally, a silent flag means suppress *all* output. In general, this allows scripts and automated systems to use the command and simply check the exit code for success, so yes, we would want that flag to hide both the Mix and Webpack output.

The `--no-progress` flag, by its definition, should simply suppress the progress bar that shows when compiling the assets. It should still show the final stats.

As long as your PR does both of the above, I'll be happy to merge :)

Nimdoc:
@bennothommo I can confirm, that running just `php artisan mix:compile --silent` produces no output with these changes. The `mix:compile` command automatically puts `--stats=none` on the webpack command, and if someone desired they could override that (this was the behavior before, and hasn't changed). So yes, the `--silent` flag, without anything else to override it, is now completely silent with these changes. Would you prefer that the `--silent` flag override everything including what users may supply after the `--` though? If so then we'd have to change the `createCommand` method from

```php
    /**
     * Create the command array to create a Process object with
     */
    protected function createCommand(string $mixJsPath): array
    {
        $basePath = base_path();
        $command = $this->argument('webpackArgs') ?? [];
        array_unshift(
            $command,
            $basePath . sprintf('%1$snode_modules%1$s.bin%1$swebpack', DIRECTORY_SEPARATOR),
            'build',
            $this->option('silent') ? '--stats=none' : '--progress',
            '--config=' . $this->getWebpackJsPath($mixJsPath)
        );
        return $command;
    }
```

To something like

```php
    /**
     * Create the command array to create a Process object with
     */
    protected function createCommand(string $mixJsPath): array
    {
        $basePath = base_path();
        
        $command = [];

        $command[] = $basePath . sprintf('%1$snode_modules%1$s.bin%1$swebpack', DIRECTORY_SEPARATOR);
        $command[] = 'build';
        $command = array_merge($command, $this->argument('webpackArgs') ?? []);
        $command[] = $this->option('silent') ? '--stats=none' : '--progress';
        $command[] = '--config=' . $this->getWebpackJsPath($mixJsPath);

        return $command;
    }
```

Move the user supply arguments to the middle so that the `--stats=none` argument overrides the user arguments.

What do you think? I prefer the current behavior with the current changes, letting users silence everything but then let some stuff through by adding webpack options on the end. 

___

Also, I just made a couple changes to `MixCompile.php` and `mix.webpack.js.fixture` to suppress just the progress bar, but leave the final stats table output.

bennothommo (approved):

bennothommo:
@Nimdoc looks good to me! I doubt many people will use the silent flag alongside others that display things, but I don't mind that it would allow people to tailor their output as they see fit, so I have no problems with that at all.

bennothommo:
@LukeTowers @jaxwilko do you envisage any problems with the above?

jaxwilko:
@bennothommo I don't think there's any issues with it, happy to do some testing in the morning tho if needed :)

bennothommo:
That'd be awesome @jaxwilko :)

LukeTowers:
Any updates needed for the docs with this PR? Otherwise I'm happy to merge

LukeTowers:
nvm, found the PR :)

PR #1089, opened 2024-03-28
Note for debug MixList.php package not active

Sorg3:
After registering my theme i couldn't compile because not active. It was automatically added to the ignored package in package.json


LukeTowers:
What caused it to be added to the list of ignored packages?

bennothommo:
AFAICT, the only way it would be added to ignored package would be if they mistakingly said no to adding it to package.json in the first place, so I'll close this for now, unless a bug can be demonstrated.

PR #1111, opened 2024-04-26
Correcting backend timestamps to use timezone preference

interworks-morr:
This corrects the timestamps on the backend to apply the timezone preference if set.  The logic has been there, but it was the function wasn't being called.

This also adds a use statement for consistency with the other places the backend preference class is used.

### DIFF
--- a/modules/backend/helpers/Backend.php
+++ b/modules/backend/helpers/Backend.php
@@ -11,6 +11,7 @@ use Winter\Storm\Router\Helper as RouterHelper;
 use System\Helpers\DateTime as DateTimeHelper;
 use Backend\Classes\Skin;
 use Backend\Helpers\Exception\DecompileException;
+use Backend\Models\Preference as BackendPreference;
 
 /**
  * Backend Helper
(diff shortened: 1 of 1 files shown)


LukeTowers:
I'm going to close this for now, timestamp conversion happens on the client side; refer to https://github.com/wintercms/winter/blob/develop/modules/backend/assets/js/winter.datetime.js.

Let me know if there's still an issue that needs to be resolved.

mjauvin:
@LukeTowers this looks valid to me, why are you closing this?

LukeTowers:
Because it would break everything 😂 

The 	`Backend::dateTime()` method returns a `<time>` element that the `winter.datetime.js` interacts with to display the correct timezone adjusted time for the user by making use of the [`<meta name="backend-timezone">`](https://github.com/wintercms/winter/blob/develop/modules/backend/layouts/_head.php#L7) tag in the document's head. 

The issue that the reporter had is that their timezone wasn't being applied to a particular time value being displayed, they were trying to fix it by forcing the `Backend::dateTime()` method to output a `<time>` element with the timezone already applied which would completely break every other use of that helper method as every timezone converted date would now be double converted (as the winter.datetime.js logic would assume that the value of the <time> element was in the application timezone) and every field / column that had `ignoreTimezone: true` would no longer respect that.

The correct solution for the user's reported issue is to figure out why the specific time value they were having issues with wasn't being converted; not to break the conversion for every other part of the system.

PR #1113, opened 2024-04-29
Add boot/initialize traits methods to backend controller

mjauvin:
Allow traits's initialize/boot methods to be called on Traits used by the Backend controller (similar to Eloquent Model).


LukeTowers:
What's the difference between the bootTrait() and initializeTrait() methods?

mjauvin:
I am now using a behavior instead as suggested.

Thanks.

PR #1117, opened 2024-05-03
Use overriden storm process in console commands

mjauvin:
Depends on https://github.com/wintercms/storm/pull/175

Use our overriden Process class that will warn users when open_basedir restrictions are in effect, and let the process be created without TTY mode enabled.


mjauvin:
Fixed upstream in Symfony/Process

PR #1128, opened 2024-05-16
Add descriptionFrom support for nested properties

AIC-BV:
Adds support for JSON columns
descriptionFrom: name
descriptionFrom: json.name
descriptionFrom: json[name]

### DIFF
--- a/modules/backend/formwidgets/RecordFinder.php
+++ b/modules/backend/formwidgets/RecordFinder.php
@@ -290,7 +290,12 @@ class RecordFinder extends FormWidgetBase
             return null;
         }
 
-        return $this->relationModel->{$this->descriptionFrom};
+        // Convert array-like notation to dot notation
+        if (preg_match('/\[(.*?)\]/', $this->relationModel, $matches)) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
@AIC-BV could you add a unit test for this?

AIC-BV:
I don't know how to do this 😬

modules/backend/formwidgets/RecordFinder.php
```diff
@@ -290,7 +290,12 @@ public function getDescriptionValue()
             return null;
         }
 
-        return $this->relationModel->{$this->descriptionFrom};
+        // Convert array-like notation to dot notation
+        if (preg_match('/\[(.*?)\]/', $this->relationModel, $matches)) {
+            $this->descriptionFrom = str_replace(['[', ']'], ['.', ''], $this->descriptionFrom);
+        }
+
+        return data_get($this->relationModel, $this->descriptionFrom);
```
LukeTowers, line 298: @jaxwilko @bennothommo any thoughts on assigning the modified value back to the local property instead of using a scoped variable at this point? I'm slightly concerned that this current approach causes getDescriptionValue() to have a secondary effect of modifying the value of $this->descriptionFrom in some cases.

PR #1129, opened 2024-05-17
WIP: Add extension events for the main relation widgets

mjauvin:
Add the following events for the primary RelationController widgets:
```
- relation.extendViewWidget
- relation.extendManageWidget
- relation.extendPivotWidget
```

Questions: 
   - do we need the global events?
   - do we need the following events:
```
   - relation.extendViewFilterWidget
   - relation.extendManageFilterWidget
   - relation.extendRefreshResults
   - relation.extendConfig
```

_Note:_

I didn't add parenthesis around the arguments in the handler method signatures because I find this is a pain when copy/pasting an example to get you started as they need to be removed.

I think we should do the same for the other events examples elsewhere in the codebase.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -394,6 +394,19 @@ class RelationController extends ControllerBehavior
 
         /*
          * View widget
+         *
+         * @event relation.extendViewWidget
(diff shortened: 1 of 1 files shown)


mjauvin:
Any feedback on this welcome.

PR #1131, opened 2024-05-20
Sign in page should redirect authenticated users to the backend

mjauvin:
logged in users should not be allowed to stay on this page without logging out first.

### DIFF
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -1,18 +1,18 @@
 <?php namespace Backend\Controllers;
 
-use Mail;
-use Flash;
-use Backend;
-use Request;
-use Validator;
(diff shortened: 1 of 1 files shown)


modules/backend/controllers/Auth.php
```diff
@@ -52,6 +52,10 @@ public function index()
      */
     public function signin()
     {
+        if (BackendAuth::user()) {
+            return Backend::redirect('backend');
```
LukeTowers, line 56: The user doesn't necessarily have access to the dashboard, so it might be nice at some point to provide a helper method that would return the first menu item they have access to and use that instead, but this is probably good enough for now given that the current behaviour is just showing them the signin page when they're already logged in.

PR #1133, opened 2024-05-21
Add signin/register event methods.

mjauvin:
Will add method docbloc once this is confirmed viable.


bennothommo:
@mjauvin this will just need the `@event` docblocks detailing the event and its usage so that the Winter docs will pick it up as an available event.

mjauvin:
@luketowers @bennothommo @jaxwilko Can you guys give your feedback on these ?

LukeTowers:
@mjauvin can you take a look at https://github.com/wintercms/winter/issues/296 and let me know your thoughts?

PR #1141, opened 2024-06-05
Integrate Vite & Rework Mix

jaxwilko:
This PR implements support for Vite side by side with Mix. All exsiting Mix functionality has been retained, but all the commands have been abstracted to allow for re-use by Vite. This allows for full support of both compilers, at the same time.

## Modified
- Commands
    - `mix:compile`, `mix:watch`, `mix:install`, `mix:list` all now rely on abstract `Asset*` commands and inherit 90% of the functionality to allow for `Vite*` commands to reuse the same code.
    - `mix:run` & `mix:update` - these were used for interacting with `npm` via the Winter CLI, these have been renamed to refelect their function to `npm:run` & `npm:update`. Alias have been set to refer back to their previous names incase anybody is using them in scripts.

## New
- Commands
    - `vite:compile` - shares args with `mix:compile` called `vite build` under the hood
    - `vite:watch` - shares args with `mix:watch` called `vite` under the hood (defaults to "watching")
    - `vite:install` - shares args with `mix:install` allows for configuring `package.json` workspaces and installing required global packages
    - `vite:list` - shares args with `mix:list`
    - `vite:config` - new allows for setting up stub files
    - `mix:config` - new allows for setting up stub files
- `System\Classes\PackageJson` - Now provides a consistant interface for modifying `package.json` files rather than using json_encode/json_decode everywhere
- Twig function `vite()` has been added, this works similarly to Laravel's `@vite()` in blade, however it required the base path to also be passed. 
- `vite:watch`, `vite:compile`, `mix:watch` & `mix:compile` now support `--disable-tty` which passes output back to the Symfony `Process` allowing for capture by callers (useful in testing).

## Docs

Add a `vite.config.js` file to whatever package (plugin, theme, module) you want, then expect the same behaviour as you would with the `mix:*` commands, but with `vite:*` instead. This can be done for you via the `vite:config` command.

When using Vite, to load your assets in your markup you need to use the new `vite()` function. It takes 2 arguments, an array of entry points, and the package name for where the assets are loaded from.

This should end up looking like:
```twig
{{ vite(['assets/css/theme.css', 'assets/javascript/theme.js'], 'theme-example') }}
```
> Note: The default vite port is `5173`, you should ensure that port is open. If you are running multiple `vite:watch` commands at the same time, the port will increment with each watch, so open those ports too if required.

Vite can also be used in the backend:
```php
<?= \System\Classes\Vite::tags(['assets/css/example-plugin.css'], 'example.plugin') ?>
```

## TODO

- [x] Testing
    - [x] Fresh install
    - [x] Existing install with mix
    - [x] Fresh plugin
    - [x] Fresh theme
    - [x] Linux
    - [x] Mac
    - [x] Windows
- [x] Vite TestCases
- [x] `*:config` TestCases


modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,307 @@
...
+use Winter\Storm\Filesystem\PathResolver;
+use Winter\Storm\Support\Facades\Config;
+use Winter\Storm\Support\Facades\File;
+use Winter\Storm\Support\Str;
+
+/**
+ * Compilable assets using for Node.js compilation and processing.
+ *
+ * This works similar to the `System\Classes\CombineAssets` class in that it allows modules, plugins and themes to
+ * register configurations that will be passed on to Laravel Mix and Node.js for compilation and processing.
```
jaxwilko:
```suggestion
 * register configurations that will be passed on to Node.js based compilers for processing.
```

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,307 @@
...
+     */
+    public function getPackageCount(): int
+    {
+        return array_sum(array_map(fn ($packages) => count($packages), ...$this->packages));
+    }
+
+    /**
+     * Returns all packages registered.
+     */
+    public function getPackages(bool $includeIgnored = false, string $type = 'mix'): array
```
jaxwilko: Kind of want to swap the arg order here, not sure if anything relies on it external to the core

modules/system/ServiceProvider.php
```diff
@@ -142,6 +143,8 @@ protected function registerSingletons()
         $this->app->singleton('backend.auth', function () {
             return \Backend\Classes\AuthManager::instance();
         });
+
+        $this->app->singleton(Vite::class, Vite::class);
```
jaxwilko:
```suggestion

        // Register the Laravel Vite singleton
        $this->app->singleton(Vite::class, Vite::class);
```

modules/cms/twig/Extension.php
```diff
@@ -166,6 +168,20 @@ public function themeFilter($url): string
         return $this->controller->themeUrl($url);
     }
 
+    /**
+     * Generates Vite tags via Laravel's Vite Object.
+     *
+     * @param array $arguments
```
jaxwilko:
```suggestion
     * @param array $arguments The list of entry points for Vite
```

modules/system/classes/CompilableAssets.php
bennothommo: Shouldn't be anything besides the Mix/Vite commands using this, so should be safe to swap.

bennothommo:
Couple of initial comments:
- Perhaps we could add `*:build` and `*:dev` command aliases to `*:compile` and `*:watch` commands, respectively. Both terms are more commonly used for JS libraries.
- If we're going to be separating the `run` and `update` commands under an `npm` namespace, then `install` should probably go under there too. Perhaps we should abstract that command to be able to define which path the user wants to go down (ie. install Mix or install Vite).
- With the `vite` command in Twig, I would think we can determine the theme's base path automatically. Is there any specific reason why we need to define it explicitly?

jaxwilko:
@bennothommo 
> - Perhaps we could add *:build and *:dev command aliases to *:compile and *:watch commands, respectively. Both terms are more commonly used for JS libraries.

100% agree, makes sense.

> - If we're going to be separating the run and update commands under an npm namespace, then install should probably go under there too. Perhaps we should abstract that command to be able to define which path the user wants to go down (ie. install Mix or install Vite).

Install isn't just installing packages, it's validating the configs are set and adding packages to workspaces, then adding deps if required, I kinda like how it is atm, although maybe it should be named something different.

> - With the vite command in Twig, I would think we can determine the theme's base path automatically. Is there any specific reason why we need to define it explicitly?

Yes for theme, maybe for plugins, and maybe for somebody adding assets from a plugin to a theme (idk why but they could). All these edge cases are solved by just setting the same base that we pass when executing the command. Basically the php side needs to match up with the js web socket server. We could default to a guess, but it may cause more confusion when it fails than just setting a path and being done with it.

bennothommo:
Thanks for the feedback @jaxwilko.

> Install isn't just installing packages, it's validating the configs are set and adding packages to workspaces, then adding deps if required, I kinda like how it is atm, although maybe it should be named something different.

That's cool by me. I just would like a rational and natural way for someone to "update" their dependencies - especially with the frequency of JS libraries and their dependencies being affected by security vulnerabilities. I suppose that `npm:update` would work - perhaps allowing us to reserve `vite:upgrade` for doing more thorough upgrades to Vite if, say, down the track we need to make configuration changes as part of an upgrade process (ie. allowing the command to do *major* upgrades to Vite).

> Yes for theme, maybe for plugins, and maybe for somebody adding assets from a plugin to a theme (idk why but they could). All these edge cases are solved by just setting the same base that we pass when executing the command. Basically the php side needs to match up with the js web socket server. We could default to a guess, but it may cause more confusion when it fails than just setting a path and being done with it.

I've been thinking about this exact scenario for some time - pretty much since we introduced Mix. I think the easiest way we can move forward on this is that we have to consider that each plugin and theme is responsible for its *own* build and assets and that's it. The fact is, not everyone is going to write plugins and components that work nicely with every type of build used by themes, especially when our goal with the CMS side is that "people can use whatever front-end framework they want".

As you said too, people can have the plugins (or at the very least, the components) inject assets into the theme when used. I still would argue that it would be most likely that the "entrypoint" to these assets would still reside in the theme if they built the plugins themselves. For third-party plugins, it's fair game for them to have their own build process and you simply use the compiled assets (or perhaps the build process in the theme could concatenate/minify these assets into the theme build). Either way, I think they're both out of scope for the `vite` filter you would be introducing.

So I wonder if we can just simplify the `vite` filter to assume all paths are relative to the theme, and make it so that the entrypoints always exist within the theme (these entrypoints can however import plugin assets however they wish).

Thoughts on this?

jaxwilko:
I don't think vite will play ball, unless we only enable vite for themes and not plugins (which was my first thought before I was able to get plugins working too).

By all means have a play with it, but I think currently just adding the base covers all possible situations with the most flexibility.

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,307 @@
...
+            'configFile' => 'winter.mix.js'
+        ],
+        'vite' => [
+            'registrationMethod' => 'registerVitePackages',
+            'configFile' => 'vite.config.js'
+        ]
+    ];
+
+    /**
+     * A list of packages registered for mixing.
```
LukeTowers:
```suggestion
     * A list of packages registered for compiling.
```

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,307 @@
...
+        /*
+         * Get packages registered in plugins.
+         *
+         * In the Plugin.php file for your plugin, you can define the `registerMixPackages` or `registerVitePackages`
+         * method and return an array, with the name of the package being the key, and the build config path - relative
+         * to the plugin directory - as the value.
+         *
+         * Example:
+         *
+         *   public function registerMixPackages()
```
LukeTowers:
```suggestion
         *   public function registerMixPackages(): array
```

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,307 @@
...
+         *   {
+         *       return [
+         *           'package-name-1' => 'winter.mix.js',
+         *           'package-name-2' => 'assets/js/build.js',
+         *       ];
+         *   }
+         *
+         *   public function registerVitePackages()
```
LukeTowers:
```suggestion
         *   public function registerVitePackages(): array
```

modules/system/classes/CompilableAssets.php
LukeTowers: I'm fine with swapping the order

jaxwilko:
@LukeTowers @bennothommo the param ordering for `getPackages()` has been swapped. I've also added some extra helpers to `CompilableAssets` & `PackageJson` to simplify logic in some commands.

LukeTowers:
@mjauvin do you feel like having a play around with this to help test it?

mjauvin:
You all know the love I have for npm... why do you even ask? 😂

jaxwilko:
@bennothommo had a thought, what if we swap the "base path thing" for the package name, i.e. `theme-example` or `winter.test`, then resolve the base via `CompilableAssets`. 

That way it makes a lot more sense as you're "loading entry files from this package".

Any thoughts?

bennothommo:
That sounds a lot better. I still think it can default to the theme base path, but yeah having it allow the theme and plugin codes that Mix uses would give that extra flexibility in a simple way. 👍

LukeTowers:
> You all know the love I have for npm... why do you even ask? 😂

That's why I ask 😂

jaxwilko:
I've put together a simple testing guide to try out vite + the new config command.

Set up a testing plugin, configure it and add stubs:
```bash
./artisan create:plugin Example.Test
./artisan create:component Example.Test Testing
./artisan vite:config Example.Test --tailwind --stubs
./artisan vite:install

echo  -e "<p class=\"text-red-900\">hello world</p>\n{{ vite(['assets/css/example-test.css'], 'example.test') }}" > plugins/example/test/components/testing/default.htm
```
Enable the component in the plugin:
```php
return [
    \Example\Test\Components\Testing::class => 'testing',
];
```
Add the component to one of your pages:
```twig
...
[testing]
...
==
...
{% component testing %}
...
```
Run vite:
```bash
./artisan vite:watch example.test
```

modules/cms/twig/Extension.php
```diff
@@ -166,6 +170,24 @@ public function themeFilter($url): string
         return $this->controller->themeUrl($url);
     }
 
+    /**
+     * Generates Vite tags via Laravel's Vite Object.
+     *
+     * @param array $arguments The list of entry points for Vite
+     * @param string $package The package name of the plugin or theme
+     * @return \Illuminate\Support\HtmlString
```
bennothommo: No need to double-up @return doc with return type definition, unless you want to specify some additional content for the return doc.
```suggestion
```

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,340 @@
...
+{
+    use \Winter\Storm\Support\Traits\Singleton;
+
+    /**
+     * The filename that stores the package definition.
+     */
+    protected string $packageJson = 'package.json';
+
+    /**
+     * List of package types and registration methods
```
bennothommo: Sure is painful, but let's start describing array shapes in the event I decide to tackle PHPStan with the modules.
```suggestion
     * @var array<string, array<string, string>> List of package types and registration methods
```

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,340 @@
...
+use Winter\Storm\Support\Str;
+
+/**
+ * Compilable assets using for Node.js compilation and processing.
+ *
+ * This works similar to the `System\Classes\CombineAssets` class in that it allows modules, plugins and themes to
+ * register configurations that will be passed on to Node.js based compilers for processing.
+ *
+ * @package winter\wn-system-module
+ * @author Ben Thomson <git@alfreido.com>, Jack Wilkinson <me@jackwilky.com>
```
bennothommo: This is all you, bro - I ain't had anything to do with it :P
```suggestion
 * @author Jack Wilkinson <me@jackwilky.com>
```

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,340 @@
...
+ * @author Winter CMS
```
bennothommo: Conventionally, I've stuck with defining us for copyright. Feel free to discard if you wish.
```suggestion
 * @copyright Winter CMS Maintainers
```

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,340 @@
...
+
+    /**
+     * Registers an entity as a package for compilation.
+     *
+     * Entities can include plugins, components, themes, modules and much more.
+     *
+     * The name of the package is an alias that can be used to reference this package in other methods within this
+     * class.
+     *
+     * By default, the CompilableAssets class will look for a `package.json` file for Node dependencies, and a config
```
bennothommo: Always use code blocks around names, as that way people using translation tools won't inadvertently translate a named entity, unless it's within a docblock tag.
```suggestion
     * By default, the `CompilableAssets` class will look for a `package.json` file for Node dependencies, and a config
```

modules/cms/twig/Extension.php
jaxwilko: I'll go through and remove all of them where relevant :)

jaxwilko:
I've added a helper so you can run vite in the backend too.
```php
<?= \System\Classes\Vite::tags(['assets/css/example-plugin.css'], 'example.plugin') ?>
```

LukeTowers:
@jaxwilko does it make sense to add support to the AssetMaker trait as well?

jaxwilko:
Relies on: https://github.com/wintercms/storm/pull/183

jaxwilko:
I've tested this with an existing theme in a completed project, it functioned as expected, however as that project uses https by default it caused some mixed content issues. This could be resolved by switching back to http only or via a fix from here: https://stackoverflow.com/questions/69417788/vite-https-on-localhost or sticking vite behind nginx. Either way it falls outside the scope of this PR but will be necessary to document.

bennothommo:
@jaxwilko I've given this a test - it's really cool, and worked nicely as far as I could tell.

However, if I am to be brutally honest, I still prefer the configuration API of Mix. It's an API that someone who detests JavaScript (ie. @LukeTowers  :stuck_out_tongue:) would still feel comfortable using. Vite is obviously leaps and bounds more easy to wrangle than Webpack, but it's still very declarative for all but the most simple of cases (someone using the base config and simply using the `vite()` method in their templates).

As soon as someone needs to do something complex, you're back into defining huge config and plugin objects in the config file. There's no way (I think) that I can do my Monaco Editor build with Vite currently.

I'm not saying we should do this now given that your intention is to still allow Mix to work, but since the underlying `laravel-mix` library is all but abandoned at this point, I'm thinking that it would be nice down the track to create a wrapper to configure Vite in the same way Mix did for Webpack, if that's at all feasible.

jaxwilko:
@bennothommo 

> I still prefer the configuration API of Mix. It's an API that someone who detests JavaScript (ie. @LukeTowers 😛)

The beauty of Winter is we can support both (and with all the abstraction I added, we can now easily support FrameworkX+n when it becomes the default next month). If I'm honest I used to write Webpack configs before writing that mix plugin years back so either mix or vite configs don't bother me much :joy:.

> I'm thinking that it would be nice down the track to create a wrapper to configure Vite in the same way Mix did for Webpack

Rollup vs. Webpack are very different, the way they opperate is different and it comes with some advantages and some disadvantages.

One issue I found earlier is that if you're running a site in https, then vite will cause issues as it's webserver is http by default, you need to go out of your way to either wrap it under nginx or use one of the many plugins that exist to fix the issue for you (or manually configure vite with your own certs).

I'm sure there are many more similar issues due to the differing paradigms between Mix & Vite, which is why I want to offer support for allowing plugins, themes and registerable things (i.e. direct calls to `CompilableAssets`) to manage a vite/mix "package", but kind of be hands off at that point, allow for whatever configuration the user wants to do with vite/mix but not get in their way.

The way I see it, if they have an issue with vite, they will google "Vite not work with this thing", which will probably get a result that tells them what to put in their config to fix it. If we change that to "winter-vite-wrapper thing not work" then they will likely get far less results and the issue will fall on us to support our custom extension.

>  but since the underlying laravel-mix library is all but abandoned at this point

Till it starts causing issues I'd like to keep supporting it, I think Vite will work very well for frontend themes but I'm not sure how well it'll work in the backend and I think Mix may generally work better due to it's simplicity (that said, @LukeTowers made be add a `AssetMaker::addVite()` method which does work nicely). We're still supporting `Assetic` after all.

> As soon as someone needs to do something complex, you're back into defining huge config and plugin objects in the config file. There's no way (I think) that I can do my Monaco Editor build with Vite currently.

Ping me when you're next free on discord and we can try and get it working :). I've tried out 2 production sites converting existing mix configs over but they weren't doing much clever stuff so it would be a good test to find any edge cases.

LukeTowers:
@jaxwilko is there a docs PR?

jaxwilko:
@LukeTowers working on it atm :)

jaxwilko:
Docs PR: https://github.com/wintercms/docs/pull/200

modules/cms/twig/Extension.php
```diff
@@ -166,6 +168,17 @@ public function themeFilter($url): string
         return $this->controller->themeUrl($url);
     }
 
+    /**
+     * Generates Vite tags via Laravel's Vite Object.
+     *
+     * @param array $entrypoints The list of entry points for Vite
+     * @param string $package The package name of the plugin or theme
```
LukeTowers:
```suggestion
```

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,340 @@
+<?php namespace System\Classes;
```
LukeTowers:
```suggestion
<?php

namespace System\Classes;
```

modules/system/classes/PackageJson.php
```diff
@@ -0,0 +1,263 @@
+<?php
+
+namespace System\Classes;
+
+use InvalidArgumentException;
+use RuntimeException;
+use Winter\Storm\Support\Facades\File;
+
+/**
+ * Php based interface for interacting with package.json files. This allows for modification of deps, devDeps, package
```
LukeTowers:
```suggestion
 * PHP based interface for interacting with package.json files. This allows for modification of deps, devDeps, package
```

modules/system/classes/PackageJson.php
```diff
@@ -0,0 +1,263 @@
...
+ * name and workspaces.
+ *
+ * @package winter\wn-system-module
+ * @author Jack Wilkinson <me@jackwilky.com>
+ * @author Winter CMS
+ */
+class PackageJson
```
LukeTowers, line 17: @jaxwilko @bennothommo any thoughts on moving this into the winter/laravel-config-writer package?

modules/system/classes/Vite.php
```diff
@@ -0,0 +1,50 @@
...
+     * @param string|array $entrypoints The list of entry points for Vite
+     * @param string|null $package The package name of the plugin or theme
+     * @return HtmlString
+     *
+     * @throws SystemException
+     */
+    public function __invoke($entrypoints, $package = null)
+    {
+        if (!$package) {
+            throw new SystemException('A package must be passed');
```
LukeTowers: Should this be an InvalidArgumentException instead? Why do we initialize it to null if we then perform a falsey check on it and reject if it fails?

modules/system/console/asset/AssetCompile.php
```diff
@@ -0,0 +1,253 @@
...
+    /**
+     * File path being watched, used for cleanup by mix:watch
+     */
+    protected string $watchingFilePath;
+
+    public function compileHandle(string $type): int
+    {
+        // Exit early if node_modules isn't available yet
+        if (!File::exists(base_path('node_modules'))) {
+            $this->error('The Node dependencies are not available, try running mix:install first.');
```
LukeTowers: Should this still be `mix:install` as the recommended command?

modules/system/console/asset/AssetConfig.php
```diff
@@ -0,0 +1,201 @@
...
+     */
+    protected $description = 'Create configuration.';
+
+    /**
+     * @var array|array[] List of meta packages
+     */
+    protected array $packages = [
+        'tailwind' => [
+            'tailwindcss' => '^3.4.0',
+            '@tailwindcss/forms' => '^0.5.2',
```
LukeTowers, line 28:
```suggestion
            '@tailwindcss/forms' => '^0.5.2',
            '@tailwindcss/typography' => '^0.5.2',
```

modules/system/console/asset/AssetInstall.php
```diff
@@ -180,62 +210,39 @@ public function handle(): int
...
+            if (!File::exists($package['config'])) {
                 $this->info(sprintf(
-                    'No Mix file found for %s, creating one at %s...',
+                    'No config file found for %s, creating one at %s...',
                     $name,
-                    $package['mix']
+                    $package['config']
                 ));
-                File::put($package['mix'], File::get(__DIR__ . '/fixtures/winter.mix.js.fixture'));
+                // @TODO: consider this when using vite
```
LukeTowers: Is this TODO done?

modules/system/console/asset/vite/ViteWatch.php
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    /**
+     * Handle the cleanup of this command if a termination signal is received
+     */
+    public function handleCleanup(): void
+    {
+        $this->newLine();
+        $this->info('Running compile to ensure files exist after exit');
```
LukeTowers, line 77: ❤️

modules/system/traits/AssetMaker.php
```diff
@@ -189,6 +200,32 @@ public function addRss($name, $attributes = [])
...
+    {
+        if (!is_array($entrypoints)) {
+            $entrypoints = [$entrypoints];
+        }
+
+        // If package was not set, attempt to guess
+        if (is_null($package)) {
+            $caller = get_called_class();
+            if (!($plugin = PluginManager::instance()->findByNamespace($caller))) {
+                throw new ApplicationException('Unable to determine vite package from namespace: ' . $caller);
```
LukeTowers: Shouldn't this be a SystemException?

modules/system/classes/CompilableAssets.php
```diff
@@ -0,0 +1,340 @@
...
+    /**
+     * @var array<string, array<string, string>> List of package types and registration methods
+     */
+    protected array $compilableConfigs = [
+        'mix' => [
+            'registrationMethod' => 'registerMixPackages',
+            'configFile' => 'winter.mix.js'
+        ],
+        'vite' => [
+            'registrationMethod' => 'registerVitePackages',
```
LukeTowers: Does this need to be a config item? Can't it be inferred from the name of the array? Does it support some level of extensibility this way that I'm not seeing at the moment?

LukeTowers (changes requested):

modules/system/classes/PackageJson.php
jaxwilko: I'm not sure people would need a single package to do both, but it's quite small so might not hurt.

modules/system/classes/Vite.php
jaxwilko: Because it's overloading the parent's behaviour which is to execute on `__invoke`, however in the parent [Vite](https://github.com/laravel/framework/blob/11.x/src/Illuminate/Foundation/Vite.php) the second param is "buildPath" which:

A. We don't need.
B. We replace with `package` to allow us to manage which package is being used.

We can't modify the type hint to not allow null because php no like so have to validate in the method's logic.

I've updated the exception tho :)

modules/system/console/asset/AssetCompile.php
jaxwilko: No it should not, see: https://github.com/wintercms/winter/pull/1141/commits/6409346e5528ff539d5e3137a1b2057bc9f1f47e

.gitattributes
```diff
@@ -8,3 +8,4 @@
 .gitattributes export-ignore
 .gitpod.yml export-ignore
 CHANGELOG.md export-ignore
+/package.json export-ignore
```
jaxwilko, line 11: @bennothommo thoughts?

The rational being with the new `*:config` command, the mix/vite deps can be added dynamically and therefore we don't need to ship a default package.json

LukeTowers (approved):

LukeTowers:
Just waiting for @bennothommo to weigh in on the remaining two items (package.json being excluded from the source export & moving the PackageJson class into either the Storm Parse package or the Laravel Config Writer package).

modules/system/classes/PackageJson.php
bennothommo: This is kinda tied to our commands and usage though, is it not? All of that exists in the System module. I don't know how much use it would be in either the Config Writer package or Storm on its own.

.gitattributes
bennothommo: @jaxwilko yeah makes sense to do this. We'll just need to perhaps document this for people who wish to develop Winter CMS, that they will need to run the `*:config` commands if they wish to develop anything built by Vite/Mix.

modules/system/classes/PackageJson.php
jaxwilko: At the moment it's only being used by Asset commands, it could be used later but until we need it somewhere else maybe we leave it be?

bennothommo: Yeah, I feel it probably should remain here unless there's a compelling use case outside of the CMS.

LukeTowers:
@jaxwilko @bennothommo are there any remaining items before we can merge this?

bennothommo:
Not that I'm aware of.

PR #1154, opened 2024-07-02
Improves drag and drop sorting of tree views

der-On:
Tested on the static pages tree.

This PR improves the drag n drop sorting of tree views, especially for large static page trees.

The main issues with the previous situation have been the pages getting very height if they have long URLs, pages resizing once dragging started and pages staying expanded once dragged. To drop pages precicesly they need to be collapsed to take as little space as possible, so one can concentrate on actually finding the right spot to drop on.

This PR does the following:

- makes text ellipsis on page URLs in the page tree work, so URL will never exceed a single line
- collapses expanded pages, once they get dragged, so they take up less screen space for better aiming at the drop target
- removes right padding of nested pages during a drag as this made pages resize vertically

~~This PR does not include the build CSS. This must be build using `php artisan winter:util compile less` to see the result.~~


mjauvin:
PR should ALWAYS include the compiled css/js files (if they are setup)

mjauvin:
Can we see a before/after screenshot for this?

der-On:
Here is a screenshot showing the changes:

Before:

grafik

After:

grafik

der-On:
Added the compiled css

der-On:
@mjauvin Please have a look.

mjauvin:
Looks cleaner for sure.

PR #1157, opened 2024-07-03
Fix issues with components with latest Foundation\Application changes

mjauvin:
Key the args by their property name

related to: https://github.com/wintercms/storm/pull/183


jaxwilko (approved):
LGTM

PR #1164, opened 2024-07-17
Add asset BundleManager & re-organize the asset helper classes

jaxwilko:
This PR adds a new `BundleManager` class which is responsible for managing the different "asset bundles" available in Winter. Bundles will automatically be offered by the `vite:create` & `mix:create` scaffolding commands as option flags.

This allows us to:
- Have a single entity responsible for managing versions
- Allow the user to configure their own bundles / overwrite the existing bundles if desired
- Manage the update process better by reducing magic numbers in the system

I've also cleaned up some of the install logic and added some user feedback similar to `create:plugin` when running `{$compiler}:create`.

In addition to the above, this PR removes the majority of the fixtures for Mix/Vite configs and now lets subscribers modify the base config on the fly.

```php
\System\Classes\Asset\BundleManager::instance()->registerCallback(function ($manager) {
    // Register a bundle (using a compiler as a key and having an array as a value means that the array 
    // will be squashed into the bundle when that compiler is being called, else the array will be discarded):
    $manager->registerBundle('jaxjs', [
        'jaxjs' => 'v1.0.0',
        'mix' => [
            'mix-jaxjs-adapter' => 'v1.2.0'
        ],
        'vite' => [
            'vite-jaxjs-adapter' => 'v1.6.0'
        ],
    ]);
    
    // Register a setup handler (is executed on install to allow for bundles to generate their own config
    // files / prep the plugin)
    $manager->registerSetupHandler('jaxjs', function (string $packagePath, string $packageType) {
        $this->writeFile(
            $packagePath . '/jax.js',
            File::get(__DIR__ . '/../fixtures/jax.js')
        );
    });

    // Register a scaffold handler (is executed during stub/config generation), `contentType`
    // can be: vite, mix, js, css.
    $manager->registerScaffoldHandler('jaxjs', function (string $contents, string $contentType) {
        return match ($contentType) {
            'mix' => $contents . PHP_EOL . <<<JAVASCRIPT
            console.log('jaxjs installed');
            JAVASCRIPT,
             'vite' => $contents . PHP_EOL . '// example',
            'css' => $content . PHP_EOL . $this->getFixture('css/jaxjs.css.fixture'),
            'js' => $this->getFixture('js/jaxjs.fixture')
            default => $contents
        };
    });
]);
```
Once the configs have been set, either via a config file, or from config calls in a plugin, the following can be ran in the CLI:
```bash
./artisan vite:create Acme.Plugin --jaxjs
# Or
./artisan mix:create Acme.Plugin --jaxjs
```

Plugins can call these methods to register handlers to modify either stub or config generation, or do additional tasks when the `{$compiler}:create` command is ran.

> **NOTE**: `setupHandlers` & `scaffoldHandlers` are executed within the context of `AssetCreate` this means they have access to the internal properties of the class such as `assetType` (`mix` or `vite`) and have access to methods such as `writeFile()` & `getFixture()`.


modules/system/classes/NodePackageVersions.php
```diff
@@ -0,0 +1,47 @@
...
+
+        // Plugins
+        'tailwindcss' => '^3.4.0',
+        '@tailwindcss/forms' => '^0.5.3',
+        '@tailwindcss/typography' => '^0.5.2',
+        'vue' => '^3.4.0',
+        '@vitejs/plugin-vue' => '^5.0.5',
+
+        // Test package - this allows us to statically test with a version that will not change
+        'winter-test-js-package' => '^1.0.3',
```
LukeTowers: Is this really required? What exactly is the test case testing for?

bennothommo:
@jaxwilko you might consider putting the "groups" of packages in that same class and call it `NodePackages`. Then the commands could simply call the group in. You could even make it extensible - ie. allow plugins to define packages if the plugins themselves provide front-end functionality?

modules/system/classes/NodePackageVersions.php
jaxwilko: See: https://github.com/wintercms/winter/blob/4f1ac3e9b5116a22580e2725d069de6c4a4e771c/modules/system/tests/classes/NodePackageVersionsTest.php

Basically just there so when we update the real package versions in the future we don't also have to update the tests.

jaxwilko:
> @jaxwilko you might consider putting the "groups" of packages in that same class and call it `NodePackages`. Then the commands could simply call the group in. You could even make it extensible - ie. allow plugins to define packages if the plugins themselves provide front-end functionality?

Thy will be done

LukeTowers (approved):

modules/system/classes/asset/BundleManager.php
```diff
@@ -0,0 +1,253 @@
...
+        $this->registeredBundles = array_replace_recursive($this->registeredBundles, $definitions);
+    }
+
+    /**
+     * Registers a single asset bundle.
+     */
+    public function registerBundle(string $name, array $definition): void
+    {
+        $this->registerBundles([$name => $definition]);
+    }
```
bennothommo: I'd personally prefer that `registerBundles` method does a loop call to the `registerBundle` method (not the other way round) and some documentation on the `registerBundle` method be added so that someone visually parsing the code (or using an IDE) will have some idea of what the methods are expecting (even if it's simple and provides a link to the docs site)

modules/system/classes/asset/BundleManager.php
```diff
@@ -0,0 +1,253 @@
...
+        $this->registeredBundles[$name][static::HANDLER_SETUP] = $closure;
+    }
+
+    /**
+     * Registers a single bundle scaffold handler.
+     */
+    public function registerScaffoldHandler(string $name, Closure $closure): void
+    {
+        $this->registeredBundles[$name][static::HANDLER_SCAFFOLD] = $closure;
+    }
```
bennothommo: Same as above, maybe some documentation of the API of handlers would be useful here.

modules/system/console/asset/AssetCreate.php
```diff
@@ -0,0 +1,233 @@
...
+        $compilableAssets = PackageManager::instance();
+        $compilableAssets->fireCallbacks();
+
+        $packages = $compilableAssets->getPackages($this->assetType, true);
+
+        if (
+            isset($packages[$package])
+            && !$this->confirm('Package `' . $package . '` has already been configured, are you sure you wish to continue?')
+        ) {
+            return 1;
```
bennothommo: A user cancelling the action should not be an exit 1. It's an intended action, so it should be exit 0.
```suggestion
            return 0;
```

bennothommo (dismissed):

modules/system/classes/asset/BundleManager.php
LukeTowers: I don't really want to document this too much, I'd prefer people don't use it unless they've got a good reason.

bennothommo: Then perhaps put that as a recommendation in the documentation, but I find our code is littered with functionality that is hardly documented and it annoys me to no end, so I don't want to add to that.

jaxwilko: > I'd personally prefer that `registerBundles` method does a loop call to the `registerBundle` method (not the other way round) and some documentation on the `registerBundle` method be added so that someone visually parsing the code (or using an IDE) will have some idea of what the methods are expecting (even if it's simple and provides a link to the docs site)

Added here: https://github.com/wintercms/winter/pull/1164/commits/acc59b9b5ca7aa8514114d25c231f9929f92f821

jaxwilko:
Docs for new BundleManager features have been added here: https://github.com/wintercms/docs/pull/203

modules/system/classes/asset/BundleManager.php
bennothommo: Thanks for sorting out the looping, @jaxwilko. However, the documentation is still a bit sparse.

Would you be able to at least summarise what is a "setup" handler and a "scaffold" handler, when they are run, and the contexts that they support?

PR #1166, opened 2024-07-19
Allow for File model IDs to be UUIDs

Denoder:
I've come to notice that when i'm making projects for people they want to utilize uuids for file uploads (everything really) regardless of performance impact, so I end up having to manually edit this every time... So can I just put this here so i don't have to keep doing this?


LukeTowers:
Works for me, will hold off on merging until I fix the develop branch though.

PR #1167, opened 2024-07-19
Added support for enabling mix manifest

jaxwilko:
This PR adds support for enabling the Mix manifest, see [Laravel Mix Asset Versioning](https://github.com/orgs/wintercms/discussions/476#discussioncomment-10088281).

The change here uses the existing options method provided by Mix to override our default behaviour.
```javascript
// Enable default manifest
mix.options({
    manifest: true,
});

// Disable manifiest (default behaviour)
mix.options({
    manifest: false,
});

// Set a custom path for the manifest
mix.options({
    manifest: 'assets/manifest.json',
});
```

### DIFF
--- a/modules/system/console/asset/fixtures/mix.webpack.js.fixture
+++ b/modules/system/console/asset/fixtures/mix.webpack.js.fixture
@@ -6,19 +6,16 @@ module.exports = async () => {
 
     const mix = require(basePath + '/node_modules/laravel-mix/src/Mix').primary;
 
-    // disable manifest
-    mix.manifest.refresh = _ => void 0
-
-    mix.listen('init', function (mix) {
+    mix.listen('init', function (_mix) {
         // Setup Winter path aliases
-        mix._api.alias({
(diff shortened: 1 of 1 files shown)


modules/system/console/asset/fixtures/mix.webpack.js.fixture
```diff
@@ -6,31 +6,37 @@ module.exports = async () => {
...
+            _mix._api.webpackConfig({
                 devtool: 'inline-source-map'
             });
-            mix._api.sourceMaps();
+            _mix._api.sourceMaps();
+        }
+
+        // Disable default manifest, allow for custom manifest
+        if (_mix.config.manifest === 'mix-manifest.json') {
+            mix.manifest.refresh = _ => void 0;
```
jaxwilko, line 35: This could be changed to :
```suggestion
            _mix.config.manifest = false;
```
But I'm not sure if that will work with every install of mix, whereas the other method 100% will.

PR #1168, opened 2024-07-21
Update create_table.stub

WebVPF:
Fix syntax


WebVPF:
Wrong fix. Correct syntax.

PR #1170, opened 2024-07-23
Populate the form fields values.

mjauvin:
Replaces https://github.com/wintercms/winter/pull/1126

### DIFF
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -246,7 +246,9 @@ class FormController extends ControllerBehavior
         $this->controller->formBeforeSave($model);
         $this->controller->formBeforeCreate($model);
 
-        $modelsToSave = $this->prepareModelsToSave($model, $this->formWidget->getSaveData(applyFilters:false));
+        $this->formWidget->setFormValues();
+        $modelsToSave = $this->prepareModelsToSave($model, $this->formWidget->getSaveData());
+
         Db::transaction(function () use ($modelsToSave) {
             foreach ($modelsToSave as $modelToSave) {
                 $modelToSave->save(null, $this->formWidget->getSessionKey());
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -246,9 +246,11 @@ class FormController extends ControllerBehavior
         $this->controller->formBeforeSave($model);
         $this->controller->formBeforeCreate($model);
 
-        $this->formWidget->setFormValues();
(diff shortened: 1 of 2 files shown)


LukeTowers:
@mjauvin is there any chance you could add a unit test that could cover all of the issues we've encountered with this change? (just interacting with the Form widget & some models directly)

mjauvin:
@LukeTowers I'd like to, but I'm not sure I know how to do this.

LukeTowers:
@mjauvin are you able to start just by listing out all the problems we've had (starting with the first issue that caused us to want to apply it in multiple locations) describing the original issue, what should have happened, what did happen, and what we did to fix the issue? I should be able to turn those into some unit tests.

mjauvin:
@LukeTowers

Initial PR #927 fixed #950:
When you define a hidden field you want to show in the filterFields() model method, the later is not correctly saved.
This behavior occurs because the model form fields filtering is not applied in the getSaveData() widget method.

PR #1099 Fixed #1036:
Try to access a model attribute via `$this` (e.g. `$this->someAttribute`) in the `filterFields()` method. Before v1.2.4 the model attributes are populated but after v1.2.4 they are all null.

mjauvin:
@LukeTowers just realized this PR probably breaks #950 ...

mjauvin:
> @LukeTowers just realized this PR probably breaks #950 ...

Just tested it and I was right... I really don't know how to solve all cases here...

mjauvin:
I'd really like to know why the form fields AND the model fields are set when `filterFields` gets called in the Form widget's `onRefresh()` but not in the FormController's `create_onSave()` ... there lies the solution to this issue.

mjauvin:
@LukeTowers I tried many different scenarios and the safest is to revert the changes to `Widgets\Form::getSaveData()`

mjauvin:
@LukeTowers @bennothommo @jaxwilko  this is the most promissing fix yet.

If there is no objections from you guys in calling `$this->formWidget->setFormValues()` from within the FormController's `create_onSave / update_onSave` methods, this should resolve all issues.

bennothommo:
@mjauvin I have no objections if it works, just need to test it in a variety of contexts to make sure it doesn't break something :)

LukeTowers:
I'll review this and see if I can build out some tests, I don't want to have to come back to this again.

mjauvin:
@LukeTowers I am still getting a case where having the `applyFiltersFromModel()` method called in `getSaveData()` creates an issue when saving a new form.

I surrender, strongly suggest we avoid calling this as done in https://github.com/wintercms/winter/pull/1126

mjauvin:
Closing this as it does not resolve the issues.

PR #1175, opened 2024-07-30
Fix Mix webpack config not being cleaned

jaxwilko:
Adds a fix for issue raised by @damsfx where the `mix.webpack.js` file was not being automatically cleaned up after running `mix:watch` / `mix:compile`.


damsfx:
Side note for Windows users : https://github.com/wintercms/winter/issues/1176

PR #1182, opened 2024-08-13
Fix nested form data in Snowboard requests

bennothommo:
This fix allows nested objects within the request data to be correctly converted to nested form data when sending the request.

Previously, if the `data` option in a Request contained nested objects or arrays within the data, these were being incorrectly converted to `[Array array]` or `[Object object]`, as the `FormData` object would simply convert them to strings. With this fix, we traverse the data structure and ensure that they are given the right key in order to keep the shape of the data in the POST data.


mjauvin:
@bennothommo that resolves the issue.

thanks.

bennothommo:
@mjauvin when you get a chance, could you give it one more test with the latest changes - I've just simplified the code a little.

mjauvin:
> @mjauvin when you get a chance, could you give it one more test with the latest changes - I've just simplified the code a little.

Looks good!

mjauvin:
@bennothommo I tested with the data structure below:
```php
$this['data'] = [
    'key' => 'value',
    'nested' => [
      'otherKey' => 'other value',
      'secondLevel' => [
        'moreLevels' => [       
          'third' => 'third level',
          'thirdObj' => (object)[
            'fourth' => 'fourth level',
          ],
        ],
        'second' => 'second level',
      ],
    ],
];
```

got the following result from the ajax handler:
```log
[2024-08-13 14:31:12] dev.INFO: Array
(           
    [key] => value
    [nested] => Array
        (           
            [otherKey] => other value
            [secondLevel] => Array
                (           
                    [moreLevels] => Array
                        (           
                            [third] => third level
                            [thirdObj] => Array
                                (
                                    [fourth] => fourth level
                                )
                
                        )
        
                    [second] => second level
                )

        )

)
```
So all good as far as I can tell!

LukeTowers (approved):

PR #1188, opened 2024-08-23
Enforce PSR-12 and organise imports on Winter CMS core files

bennothommo:
This commit changes the PHP code sniffer to use PSR-12 rules, as that ruleset will pick up a lot of the code style issues being flagged in PRs recently.

In order to set the baseline, all current core files have been reformatted to fit the ruleset.

I have also implemented rules around imports in files - mainly, taking out the aliases and using qualified facades and class names, and stripping out unused imports. This will improve IDE support for PHP intelligence and code completion.

Arrays are also now enforced to have a trailing comma to improve diffs.

Sorry in advance for the code review... :grimacing:

- [x] Docs updates (https://github.com/wintercms/docs/pull/205)
- [x] Add `php artisan winter:phpcs` command
- [x] Ensure stubs follow correct style


.github/workflows/manifest.yml
```diff
@@ -74,13 +43,8 @@ jobs:
     runs-on: ubuntu-latest
     needs: updateManifest
     steps:
-      - name: Cancel previous incomplete runs
-        uses: styfle/cancel-workflow-action@0.8.0
-        with:
-          access_token: ${{ github.token }}
```
LukeTowers: Are we not using caching / cancelling previous runs anymore?

bennothommo: We are in some actions, using the new `concurrency` configuration. However, the manifest should probably complete even if it has been superseded, otherwise it could corrupt the manifest file.

modules/backend/formwidgets/relation/partials/_relation.php
```diff
@@ -1,3 +1,3 @@
 <div class="relation-widget" id="<?= $this->getId() ?>">
-    <?= $this->makePartial('~/modules/backend/widgets/form/partials/_field_'.$field->type.'.php', ['field' => $field]) ?>
+    <?= $this->makePartial('~/modules/backend/widgets/form/partials/_field_' . $field->type . '.php', ['field' => $field]) ?>
 </div>
```
LukeTowers: @bennothommo do we have any rules for trailing newline on all files?

modules/backend/layouts/_custom_styles.php
```diff
@@ -1,6 +1,8 @@
 <?php
+
     use Backend\Models\BrandSetting;
     use Backend\Models\EditorSetting;
+
```
LukeTowers: Do we have to have this whitespace? It doesn't add much IMO

modules/backend/widgets/MediaManager.php
```diff
@@ -926,13 +927,15 @@ protected function getCurrentFolder(): string
...
-        if (!in_array($filter, [
+        if (
+            !in_array($filter, [
             self::FILTER_ALL,
             MediaLibraryItem::FILE_TYPE_IMAGE,
             MediaLibraryItem::FILE_TYPE_AUDIO,
             MediaLibraryItem::FILE_TYPE_DOCUMENT,
-            MediaLibraryItem::FILE_TYPE_VIDEO
-        ])) {
+            MediaLibraryItem::FILE_TYPE_VIDEO,
```
LukeTowers: Should this be indented one more level?

modules/backend/widgets/lists/partials/_list_head_row.php
```diff
@@ -20,7 +20,7 @@
                 <?php if ($column->width): ?>
                     style="width: <?= $column->width ?>"
                 <?php endif ?>
-                class="<?= $this->sortColumn==$column->columnName?'sort-'.$this->sortDirection.' active':'sort-desc' ?> list-cell-name-<?= $column->getName() ?> list-cell-type-<?= $column->type ?> <?= $column->getAlignClass() ?> <?= $column->headCssClass ?>"
+                class="<?= $this->sortColumn == $column->columnName ? 'sort-' . $this->sortDirection . ' active' : 'sort-desc' ?> list-cell-name-<?= $column->getName() ?> list-cell-type-<?= $column->type ?> <?= $column->getAlignClass() ?> <?= $column->headCssClass ?>"
```
LukeTowers: Should this be `===`?

modules/cms/twig/SnowboardNode.php
```diff
@@ -1,11 +1,13 @@
-<?php namespace Cms\Twig;
+<?php
 
-use Config;
+namespace Cms\Twig;
+
+use Illuminate\Support\Facades\URL;
```
LukeTowers: Any issues with the "URL" vs "Url"?

modules/backend/formwidgets/relation/partials/_relation.php
bennothommo: Yeah I believe that's included in PSR12, but might not apply for the templates / partials.

modules/backend/layouts/_custom_styles.php
bennothommo: No we don't, probably just a tweak to the config for one of the rules. I'll have a look.

modules/backend/widgets/MediaManager.php
bennothommo: I believe that's another rule, but not one I'm using right now. Will take a look.

modules/backend/formwidgets/relation/partials/_relation.php
LukeTowers: Can we add a rule to make it apply?

modules/system/models/MailLayout.php
```diff
@@ -1,11 +1,13 @@
-<?php namespace System\Models;
+<?php
 
-use View;
-use Model;
+namespace System\Models;
+
+use File as FileHelper;
```
LukeTowers: Which File is this?

modules/system/partials/_settings_menu.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
LukeTowers: This one doesn't make any sense

modules/backend/formwidgets/relation/partials/_relation.php
bennothommo: It appears not - the code sniffer can only look at PHP blocks, it cannot read HTML as far as I can tell

modules/backend/widgets/lists/partials/_list_head_row.php
bennothommo: The original functionality was just the two `==`. It wouldn't change that, it would just ensure there's correct spacing around. I'd say leave it as is unless we're sure changing it to strict won't break anything.

modules/cms/twig/SnowboardNode.php
bennothommo: Not that I have seen - I don't think PHP is case-sensitive in that regard.

modules/backend/layouts/_custom_styles.php
bennothommo: I've added an exception for that rule for partials, since changing the config changes it across the board and would allow the following in a PHP-only file (which looks wonky IMO):

```php
<?php
namespace Winter\Storm\Classes;
use Illuminate\Support\Facades\DB;
class Something extends SomethingElse
{
...
```

modules/backend/widgets/MediaManager.php
bennothommo: Added rule.

LukeTowers:
@bennothommo can you submit a PR to our developer guidelines docs to reflect these changes? Also, is it worth adding a `winter:phpcs` command to run the coding styles against plugins and then providing that as a command that plugin devs can run against their codebase to make sure that it conforms with the community standards?

Another thing to check is our stub files, I'm not sure if they've all been updated to our latest preferences for generating PHP files.

modules/system/console/WinterSniff.php
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * Create a new command instance.
+     */
+    public function __construct()
+    {
+        parent::__construct();
+
+        // Register aliases for backwards compatibility with October
+        $this->setAliases(['winter:phpcs']);
```
LukeTowers: We have a property for that already: https://github.com/wintercms/winter/blob/develop/modules/system/console/asset/mix/MixCompile.php#L34-L39

phpcs.base.xml
```diff
@@ -0,0 +1,184 @@
...
+    <rule ref="SlevomatCodingStandard.Namespaces.UseFromSameNamespace" />
+    <!-- Don't allow starting backslash for imports as there's no need for them -->
+    <rule ref="SlevomatCodingStandard.Namespaces.UseDoesNotStartWithBackslash" />
+    <!-- Don't allow unused imports (using them in docblocks is OK)-->
+    <rule ref="SlevomatCodingStandard.Namespaces.UnusedUses">
+        <properties>
+            <property name="searchAnnotations" value="true" />
+        </properties>
+    </rule>
+</ruleset>
```
LukeTowers: This file should probably live in the system module and be referenced from the stubs as such to avoid having the core ruleset changed on a per project basis and also to support existing projects making use of this feature when they don't have this file present since it's not distributed through a composer module update.

composer.json
```diff
@@ -42,6 +42,7 @@
         "mockery/mockery": "^1.4.4",
         "fakerphp/faker": "^1.9.2",
         "squizlabs/php_codesniffer": "^3.2",
+        "slevomat/coding-standard": "^8.15.0",
```
LukeTowers, line 45: Is this required for the phpcs command to work? If so we should document it in the commit message / release notes.

LukeTowers: And or maybe include it in the system module's composer.json?

LukeTowers:
Initial thoughts from testing on a client project:

## Problems:
- [x] Running on existing project with old phpcs.xml probably won't work because of the reference to winter.demo
- [x] Running on no file doesn't work because of the dependency on phpcs.base.xml being in the root (solve by moving to system module)
- [x] Old require-dev is insufficient to run as we require SlevomatCodingStandard rules
- [x] The scaffolded root phpcs.xml references demo plugin, doesn't exist on most existing projects, also wouldn't cover their plugins, so what is the purpose of a project root level file?
- [x] bootstrap/app.php, autoload.php fail on existing projects
- [x] Ton of warnings on core files from default configuration
- [x] No ability to add --fix flag to switch to phpcbf

## Coding style suggestions:
- [ ] Update rules to account for changes to controller action method names in v1.2.14
- [ ] Disallow or warn on loose type comparison (== vs ===)
- [x] Use of overridden Illuminate facades when a winter one exists
    - Config, DB, Event, File, Http (Do not block this one, Laravel one should be what we use going forwards), Mail?, Url?, Validator?
- [x] Block "use Model;" class alias instead of Winter\Storm\Database\Model (block or warn against all global use imports if possible, or make it easier for IDEs to play nice with them)
- [ ] Not using anonymous classes in migrations
- [ ] Empty methods in Plugin.php (or other clases, perhaps just issue a warning?)
- [ ] Use of routes in Plugin.php instead of in the routes.php file
- [ ] Semicolon on line by itself (usually unintentional, can result from a fix that applies ;;)
- [ ] Warning for .htm files in the backend view directories (example: controllers/*/.htm)
- [ ] Use of dot syntax for `implement` values (when not using soft implementation `@`). I.e. `Backend.Behaviors.FormController` should always be `\Backend\Behaviors\FormController::class`.
- [ ] Alphabetical order for all second+ level keys in translation files? (i.e. `*.*`, but root level keys can remain unalphabetized)
- [ ] Warning for `$formConfig = 'config_form.yaml'` etc (overridden property matches default value).

## Enhancement Ideas
- [x] Switch WinterTest to extend Winter's console command and alias `test` to it
- [x] Alias `sniff` to `winter:sniff`
- [x] Add support for `--fix` option in sniff
- [ ] Evaluate Laravel coding standards for alignment with Winter: https://docs.styleci.io/presets#laravel
- [ ] Add support for `--baseline` option in sniff to generate a baseline to ignore errors from
- [ ] Add support for module options to match `winter:test`
- [ ] Make winter VSCode extension auto load or recommend the PHPCS extension
- [ ] Add github action support for auto-fixing files on PRs
- [ ] Provide coding standard as an external repository so we can import it on our packages & plugins
- [x] Add composer scripts `test`, `sniff`, etc

## To test
- [x] Test running with the same package identifiers as test
- [x] Test running on a new plugin
- [x] Test phpcsfixer
- [x] Test output of all scaffold commands for adherance to standards
- [ ] Run sniff & fix on all first party Winter plugins and review changes / errors / warnings from those

composer.json
bennothommo: Yes, it is required for the extra rules we use in the coding standard. We can't include it in the system module however because you cannot include development requirements from a Composer dependency AFAIK. It's why we have a lot of duplicate development requirements in the main repo and Storm, for example.

PR #1191, opened 2024-08-26
Fix displaying repeater item titles in preview contexts

AIC-BV:
Cleaned up version of the preview HTML:
```
<div class="text-field" data-field-name="name">
    <label>Name</label>
    <span class="form-control">Test</span>
</div>
```
As you can see, there is no element to select with class `text-field` **if titleFrom is set**, because the target itself has the class:
```
var $disabledTextInput = $('.text-field:first > .form-control', $target)
```
so I removed the `.text-field` part of the query selector

Now it succesfuly shows the text value from `name` in both preview and update context.
If titleFrom is not set, it will take the first `.form-control` value, which is the same field as in update context
image
Before this change, it showed nothing, because the selector was invalid
image

fields.yaml:
```
items:
    context: ['preview', 'update', 'relation']
    tab: Content
    label: aic.aftersales::lang.form.items.title
    titleFrom: name
    type: repeater
    style: collapsed
    dependsOn: order
    form:
        fields:
            name:
                label: aic.aftersales::lang.form.items.name
            requires_payment:
                label: aic.aftersales::lang.form.items.requires_payment
                type: balloon-selector
                options:
                    0: Nee
                    1: Ja
                default: 0
```

### DIFF
--- a/modules/backend/formwidgets/repeater/assets/js/repeater.js
+++ b/modules/backend/formwidgets/repeater/assets/js/repeater.js
@@ -259,7 +259,7 @@
                     return $textInput.val()
             }
         } else {
-            var $disabledTextInput = $('.form-control', $target)
+            var $disabledTextInput = $('.form-control:first', $target)
             if ($disabledTextInput.length) {
                 return $disabledTextInput.text()
             }
--- a/modules/backend/formwidgets/repeater/assets/js/repeater.js
+++ b/modules/backend/formwidgets/repeater/assets/js/repeater.js
@@ -259,7 +259,7 @@
                     return $textInput.val()
             }
         } else {
-            var $disabledTextInput = $('.text-field:first > .form-control', $target)
+            var $disabledTextInput = $('.form-control', $target)
             if ($disabledTextInput.length) {
                 return $disabledTextInput.text()
             }


modules/backend/formwidgets/repeater/assets/js/repeater.js
```diff
@@ -259,7 +259,7 @@
                     return $textInput.val()
             }
         } else {
-            var $disabledTextInput = $('.text-field:first > .form-control', $target)
+            var $disabledTextInput = $('.form-control:first', $target)
```
LukeTowers, line 262: if the `text-field` class is on `$target`, could we do `&.text-field` instead? Also is this currently broken in both update and preview contexts? If so, what is the difference between them that is causing the structure of the fields to be different?

AIC-BV: In update context the HTML structure is different.
It has an [input field](https://github.com/wintercms/winter/blob/732ca4190efc2269b64fc8f119518e5b70e3ff73/modules/backend/formwidgets/repeater/partials/_repeater.php#L29), which is [selected in JavaScript](https://github.com/wintercms/winter/blob/732ca4190efc2269b64fc8f119518e5b70e3ff73/modules/backend/formwidgets/repeater/assets/js/repeater.js#L251).
titleFrom in this context is already working as described in the docs.

In preview mode however, the HTML structure is different.
It does not have an input field, but a `<span class='form-control'>` instead.
The JavaScript will go in the else part and select the 'disabled text input'.

**Here are actually 2 options:**

1. You defined titleFrom: target is the parent of `<span class='form-control'>`
2. You didn't define titleFrom: target is the row itself

**This is where it goes wrong in preview context.**
In case 1, it will try to select `.text-field:first` but that does not exist because its on the element itself resulting in an empty string.
In case 2, it will find the first `ul` and display that text. In my case this is not my text field, but a balloon selector. Remember, there are no form fields in preview context. input & select are not present.

**So I fixed**
Case 1 by removing the `.text-field:first` because it is an invalid selector. But yes, it could have the &. notation but don't see the advantage of that in this case?
Case 2 will have the same behaviour as before: if there is an `<ul>` it will get that item first, if not it will get the `.form-control:first` like in case 1.

--
It is probably not entirely foolproof (working in every form situation) but it must be much better than it currently was because it has the same behaviour and a defined titleFrom now works in preview mode

AIC-BV: Do I still need to do something to finish this PR? 🙂

LukeTowers: @AIC-BV it seems mostly fine, would you be willing to take a stab at writing a test case for this?

PR #1192, opened 2024-08-26
Add support for translation strings in FormField->options()

mjauvin:
- add valueLabel() method
- add support for translation strings in options() method

### DIFF
--- a/modules/backend/classes/FormField.php
+++ b/modules/backend/classes/FormField.php
@@ -4,6 +4,7 @@ namespace Backend\Classes;
 
 use BackedEnum;
 use Html;
+use Lang;
 use Winter\Storm\Database\Model;
 use Winter\Storm\Html\Helper as HtmlHelper;
 use Winter\Storm\Support\Str;
@@ -234,6 +235,9 @@ class FormField
                 $callable = $this->options;
                 return $callable();
             }
+            elseif (is_array($options = Lang::get($this->options))) {
+                return $options;
+            }
 
             return [];
         }
@@ -241,6 +245,24 @@ class FormField
         $this->options = $value;
 
         return $this;
+    }
+
+    /**
+     * Returns the label(s) associated with a field value(s) for the field options
+     */
+    public function valueLabel(): string|array
+    {
+        $defaultLabel = '';
+
+        if (!($options = $this->options()) || is_null($this->value)) {
+            return $defaultLabel;
+        }
+
+        if (is_array($this->value)) {
+            return array_map(fn($value) => array_get($options, $value, $defaultLabel), $this->value);
+        } else {
+            return array_get($options, $this->value, $defaultLabel);
+        }
     }
 
     /**
--- a/modules/backend/classes/FormField.php
+++ b/modules/backend/classes/FormField.php
@@ -259,7 +259,7 @@ class FormField
         }
 
         if (is_array($this->value)) {
-            return array_map(fn($value) => array_get($options, $value, $defaultLabel), $this->value);
+            return array_map(fn ($value) => array_get($options, $value, $defaultLabel), $this->value);
         } else {
             return array_get($options, $this->value, $defaultLabel);
         }
--- a/modules/backend/classes/FormField.php
+++ b/modules/backend/classes/FormField.php
@@ -247,6 +247,7 @@ class FormField
     }
 
     /**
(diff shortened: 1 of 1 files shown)


modules/backend/classes/FormField.php
```diff
@@ -234,6 +235,9 @@ public function options($value = null)
                 $callable = $this->options;
                 return $callable();
             }
+            elseif (is_array($options = Lang::get($this->options))) {
```
LukeTowers:
```suggestion
            elseif (is_string($this->options) && is_array($options = Lang::get($this->options))) {
```

modules/backend/classes/FormField.php
```diff
@@ -243,6 +247,24 @@ public function options($value = null)
         return $this;
     }
 
+    /**
+     * Returns the label(s) associated with a field value(s) for the field options
+     */
+    public function valueLabel(): string|array
```
LukeTowers: What is this used for?

mjauvin: Good question! :rofl:

It's usefull if managing a Form widget that has fields with options (dropdown, etc) to extract the label associated with its value in the field's options.

modules/backend/classes/FormField.php
```diff
@@ -234,6 +235,9 @@ public function options($value = null)
                 $callable = $this->options;
                 return $callable();
             }
+            elseif (is_string($this->options) && is_array($options = Lang::get($this->options))) {
```
bennothommo:
```suggestion
            } elseif (is_string($this->options) && is_array($options = Lang::get($this->options))) {
```

modules/backend/classes/FormField.php
```diff
@@ -243,6 +247,24 @@ public function options($value = null)
...
     }
 
+    /**
+     * Returns the label(s) associated with a field value(s) for the field options
+     */
+    public function valueLabel(): string|array
+    {
+        $defaultLabel = '';
+
+        if (!($options = $this->options()) || is_null($this->value)) {
```
bennothommo: @LukeTowers side note - there's a PHPCS rule that prevents side effects within a condition definition (such as defining a variable in an `if` statement).

I'd be keen to use it because I really am not a fan of this :stuck_out_tongue: 

```suggestion
        if (!($options = $this->options()) || is_null($this->value)) {
```

modules/backend/classes/FormField.php
```diff
@@ -243,6 +247,24 @@ public function options($value = null)
...
+            return $defaultLabel;
+        }
+
+        if (is_array($this->value)) {
+            return array_map(fn($value) => array_get($options, $value, $defaultLabel), $this->value);
```
bennothommo:
```suggestion
            return array_map(fn ($value) => array_get($options, $value, $defaultLabel), $this->value);
```

bennothommo (dismissed):

modules/backend/classes/FormField.php
mjauvin: I'm not a fan of this PHPCS rule... 😛

modules/backend/classes/FormField.php
```diff
@@ -4,6 +4,7 @@
 
 use BackedEnum;
 use Html;
+use Lang;
```
LukeTowers: Can we use the full version of this import instead of the global alias? Would like to gradually migrate that way.

LukeTowers:
@mjauvin are any changes to the docs required?

modules/backend/classes/FormField.php
mjauvin: done

mjauvin:
> @mjauvin are any changes to the docs required?

Not really, it was documented when the counterpart to this was added to `Backend\Widgets\Form::getOptionsFromModel()`.

modules/backend/classes/FormField.php
```diff
@@ -3,7 +3,8 @@
 namespace Backend\Classes;
 
 use BackedEnum;
-use Html;
+use Winter\Storm\Support\Facades\Html;
+use Illuminate\Support\Facades\Lang;
```
LukeTowers, line 6: Can you order the imports please? Still waiting for @bennothommo's coding standards PR to be able to be applied so that we can do them automatically

mjauvin: done

PR #1201, opened 2024-09-11
Make stripe loader configurable with data attribute

webbati:
Add option to define stripe visibility using data attribute (data-request-stripe=['true'|'false']

If the option is omitted stripe option will be true.

Now only using Snowboard with JavaScript API stripe option is available, with this commit will be availabe also using data attribute

Created PR with documentation in docs https://github.com/wintercms/docs/pull/208


LukeTowers:
@bennothommo do you have any feedback for this?

bennothommo:
Looks great. Happy for it to be merged.

PR #1202, opened 2024-09-12
Completed npm commands (npm:install, npm:update, npm:run)

jaxwilko:
This PR removes `[vite|mix]:install`s ability to install an npm package into the root `package.json` (a hold over from the previous mix system). In it's place it adds the commands `npm:install <compileablePackage?> <npmPackage>` which allows the user to install a packge into a workspace.

I.e. if I want to add `leftpad` to my plugin `JaxWilko.Example` I can run `./artisan npm:install jaxwilko.example leftpad` which will then add the dependency to `plugins/jaxwilko/example/package.json` then run the install command as the root project owner (root package). This is powered by npm and we just need to run the right command in the workspace package.

This PR also reworks `npm:update` and `npm:run` so they share a parent with the npm execution logic and simplifies their internal logic.

## NPM Run

`npm:run` still works as before. `./artisan npm:run jaxwilko.example test` will execute the `test` script within the `jaxwilko.example` package.

## NPM Update

`npm:update` now supports a package to update, as well as now taking a assetPackage to specify which workspace to run the command within.

- `./artisan npm:update jaxwilko.example tailwindcss` will update `tailwindcss` within `jaxwilko.example`
- `./artisan npm:update tailwindcss` will update `tailwindcss` globally (within the root package, workspaces can still install their own independent version).

## NPM Install

- `./artisan npm:install jaxwilko.example tailwindcss` will install `tailwindcss` within `jaxwilko.example`
- `./artisan npm:install tailwindcss` will install `tailwindcss` in the root package.json
- `./artisan npm:install tailwindcss --dev` will install `tailwindcss` within `devDependencies` rather than `dependencies`

## TODO
- [x] New tests
    - [x] npm:install 
    - [x] npm:update 
    - [x] npm:run
 - [x] Fix detection of ignored packages in `asset:install`
 - [x] Fix detection of missing configuration and offer to run config creation command in `asset:install`
 - [x] Restict `asset:install` to only install requested package and not all unless no package is specified


modules/system/console/asset/AssetInstall.php
```diff
@@ -194,14 +195,36 @@ protected function validateRequirePackagesPresent(PackageJson $packageJson): Pac
...
+
+        if (!$foundRequestedPackage) {
+            foreach ($registeredPackages as $name => $package) {
+                if ($requestedPackage === $name) {
+                    $foundRequestedPackage = true;
+                }
+            }
+
+            // We did not find the package, exit
+            if (!$foundRequestedPackage) {
```
LukeTowers:
```suggestion
            if (!isset($registeredPackages[$requestedPackage]])) {
```

modules/system/console/asset/NpmCommand.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+namespace System\Console\Asset;
```
LukeTowers: This should probably be in the `System\Console\Asset\Npm` namespace shouldn't it?

josephcrowell:
Yes that's the expected behavior.

josephcrowell:
In the mean time, until this is merged, what would be the correct npm command to do the equivalent npm update on a plugin or theme?

I was working on a puppeteer based html to pdf plugin specific to winter and capable of using Firefox that I would like to continue development on and a Daisy  UI based theme.

bennothommo:
@josephcrowell yes, you can use `npm` directly if you wish. I would be using it on the project root as opposed to within the plugin or theme itself, in order to completely replicate what the `artisan` command does.

modules/system/classes/asset/PackageManager.php
```diff
@@ -24,10 +24,14 @@ class PackageManager
 {
     use \Winter\Storm\Support\Traits\Singleton;
 
+    public const TYPE_THEME = 'theme';
+    public const TYPE_MODULE = 'module';
+    public const TYPE_PLUGIN = 'plugin';
```
LukeTowers, line 29: At some point it might be worth making enums for the extension types in Winter to centralize this sort of thing (and others like paths etc)

modules/system/console/asset/NpmCommand.php
LukeTowers: @jaxwilko any thoughts on the above?

modules/system/console/asset/npm/NpmInstall.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace System\console\asset\npm;
```
LukeTowers:
```suggestion
namespace System\Console\Asset\Npm;
```

modules/system/console/asset/AssetInstall.php
```diff
@@ -8,6 +8,9 @@
 use System\Classes\Asset\PackageJson;
 use System\Classes\Asset\PackageManager;
 use System\Classes\PluginManager;
+use System\Console\Asset\Exceptions\PackageIgnoredException;
+use System\console\asset\exceptions\PackageNotConfiguredException;
```
LukeTowers:
```suggestion
use System\Console\Asset\Exceptions\PackageNotConfiguredException;
```

LukeTowers:
@jaxwilko are there any changes that could be made to the docs?

jaxwilko:
@LukeTowers probably yes, at least some work on clarification of the expected / actual behaviour, i'll open a PR tomorrow and also add some more comments to the code in this PR to make it more clear what's going on :)

jaxwilko:
Docs update PR: https://github.com/wintercms/docs/pull/213

modules/system/classes/asset/PackageManager.php
```diff
@@ -24,10 +24,14 @@ class PackageManager
...
+
     /**
      * The filename that stores the package definition.
```
jaxwilko, line 32: This needs updating

modules/system/classes/asset/PackageManager.php
```diff
@@ -325,11 +351,40 @@ public function registerPackage(string $name, string $path, string $type = 'mix'
         $this->packages[$type][$name] = [
             'path' => $path,
             'package' => $package,
-            'config' => $config,
-            'ignored' => $this->isPackageIgnored($path),
+            'config' => $config
         ];
     }
```
jaxwilko, line 357: Needs phpdoc

modules/system/classes/asset/PackageManager.php
```diff
@@ -325,11 +351,40 @@ public function registerPackage(string $name, string $path, string $type = 'mix'
...
+        }
+
+        // Check if a package could be a plugin
+        if (PluginManager::instance()->exists($package)) {
+            return static::TYPE_PLUGIN;
+        }
+
+        return null;
+    }
+
```
jaxwilko, line 384: Needs phpdoc

modules/system/console/asset/AssetInstall.php
```diff
@@ -44,127 +47,140 @@ abstract class AssetInstall extends Command
...
 
-                // Check if a package could be a plugin
-                if (PluginManager::instance()->exists($package)) {
-                    $compilableAssets->registerPackage(
-                        $package,
-                        PluginManager::instance()->getPluginPath($package) . '/' . $this->configFile,
-                        $this->assetType
-                    );
-                    continue;
+                switch ($packageManager->getPackageTypeFromName($package)) {
```
jaxwilko, line 150: Need to review, not clean

modules/system/console/asset/AssetInstall.php
```diff
@@ -194,64 +210,147 @@ protected function validateRequirePackagesPresent(PackageJson $packageJson): Pac
...
+
+        // Process each found package
+        foreach ($registeredPackages as $name => $package) {
+            $this->processPackage($packageJson, $name, $package);
+        }
+
+        return $packageJson;
+    }
+
+    protected function processPackage(PackageJson $packageJson, string $name, array $package, bool $force = false): bool
```
jaxwilko, line 284: Needs phpdoc

modules/system/console/asset/AssetInstall.php
```diff
@@ -265,18 +364,26 @@ protected function installPackageDeps(): int
 
             return 1;
         }
+    }
 
-        $this->info('');
+    protected function getPackageManager(): PackageManager
```
jaxwilko, line 386: Needs phpdoc

modules/system/console/asset/NpmCommand.php
jaxwilko: I kinda prefer it to be above the commands as it's an abstract but no real preference other than that

modules/system/console/asset/NpmCommand.php
```diff
@@ -0,0 +1,76 @@
...
+
+use Symfony\Component\Process\Exception\ProcessSignaledException;
+use Symfony\Component\Process\Process;
+use System\Classes\Asset\PackageJson;
+use System\Classes\Asset\PackageManager;
+use System\Console\Asset\Exceptions\PackageNotRegisteredException;
+use Winter\Storm\Console\Command;
+
+abstract class NpmCommand extends Command
```
jaxwilko: Needs phpdoc - whole file

modules/system/console/asset/exceptions/PackageIgnoredException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
```
jaxwilko: @LukeTowers what do you think about all these exceptions?

modules/system/console/asset/npm/NpmInstall.php
```diff
@@ -0,0 +1,49 @@
...
+     */
+    protected $signature = 'npm:install
+        {package? : The package name to add configuration for}
+        {npmArgs?* : Arguments to pass through to the "npm" binary}
+        {--npm= : Defines a custom path to the "npm" binary}
+        {--d|dev : Install packages in devDependencies}
+        {--s|silent : Silent mode.}
+        {--disable-tty : Disable tty mode}';
+
+    public function handle(): int
```
jaxwilko, line 34: Phpdoc here

modules/system/console/asset/NpmCommand.php
LukeTowers: Sure it's an abstract, but it's an abstract that will only ever be used by the classes in the npm folder, so I think it makes sense for it to live there instead.

jaxwilko: Fair, it shall be moved :)

modules/system/console/asset/exceptions/PackageIgnoredException.php
LukeTowers: @jaxwilko eh, does it serve a purpose having custom exceptions or would it be better to just have custom messages on a SystemException? What's actually using them? Less code = more betterer

LukeTowers:
@jaxwilko what's left?

jaxwilko:
@LukeTowers just reworking the exceptions

jaxwilko:
@LukeTowers custom exceptions have been removed and replaced with `SystemException`

PR #1206, opened 2024-09-15
Initialize the Filter widget even when no scopes are initially defined

mjauvin:
This makes it possible to dynamically extend filter scopes even if no scopes exist.

### DIFF
--- a/modules/backend/behaviors/ListController.php
+++ b/modules/backend/behaviors/ListController.php
@@ -233,7 +233,7 @@ class ListController extends ControllerBehavior
          * Prepare the filter widget (optional)
          */
         if (isset($listConfig->filter)) {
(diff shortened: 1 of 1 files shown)


mjauvin:
This is related to our issue:
ref. https://github.com/twigphp/Twig/issues/4090

I was hoping I'd see a reference to a PR that fixed this BC break in the issue, no such luck...

mjauvin:
Also related:
ref. https://github.com/twigphp/Twig/issues/4090

mjauvin:
I think this is the one that restored the previous behavior:
ref. https://github.com/twigphp/Twig/pull/4216

bennothommo (approved):

PR #1207, opened 2024-09-18
Allow command names to contain number

damsfx:
Allow command name to contains numbers, eg:
```bash
php artisan create:command Acme.MyPlugin4You MyCommand
```


LukeTowers:
@damsfx can you add some unit tests for this? I still want it to create valid class names, which can't start with a number IIRC

damsfx:
> @damsfx can you add some unit tests for this? I still want it to create valid class names, which can't start with a number IIRC

Done : 
 - Plugin and command names can contain a number, but they can't start with it
 - Add tests

Issue #1007, opened 2023-11-11
artisan command create Factory 


alesha2911:
### Package targeted

Winter CMS

### Description

Hello, please it add  artisan command create factory.
https://laravel.com/docs/9.x/eloquent-factories

LukeTowers:
@alesha2911 are you interested in submitting a PR to add this scaffolding command? There's lots of existing scaffolding commands in modules/system/console for you to reference if you're interested.

sephyld:
Hey everyone, I'd like to try to tackle this one if it's ok.
I'll probably submit a draft PR as soon as I get the chance.

LukeTowers:
Implemented in #1209

PR #1209, opened 2024-09-22
Add create:factory command

sephyld:
also added `--F|factory` option to make:model for factory creation

---

Implements feature request https://github.com/wintercms/winter/issues/1007

---

Hello everyone, I'd love a review of this PR. My main concern is that I'm not using a Winter-specific Factory class to inherit from in the factory.stub, but the created Factory inherits directly from the Laravel base factory class.

```php
// Created Factory Class PostFactory.php
//[...]
use Illuminate\Database\Eloquent\Factories\Factory;

class PostFactory extends Factory
//[...]
```

Unlike the related model, which inherits from the Winter Model class

```php
// Created Model Class Post.php
//[...]
use Winter\Storm\Database\Model;

/**
 * Post Model
 */
class Post extends Model
{
```

I don't think this is inherently an issue, but in the case it is, I think I'll have to add a Factory base class in the Storm library

### DIFF
--- a/modules/system/console/CreateFactory.php
+++ b/modules/system/console/CreateFactory.php
@@ -1,0 +1,62 @@
+<?php namespace System\Console;
+
+use System\Console\BaseScaffoldCommand;
+
+class CreateFactory extends BaseScaffoldCommand
+{
+    /**
+     * @var string|null The default command name for lazy loading.
+     */
+    protected static $defaultName = 'create:factory';
+
+    /**
+     * @var string The name and signature of this command.
+     */
+    protected $signature = 'create:factory
+        {plugin : The name of the plugin. <info>(eg: Winter.Blog)</info>}
+        {factory : The name of the factory to generate. <info>(eg: PostFactory)</info>}
+        {--m|model= : The name of the model. <info>(eg: Post)</info>}
+        {--f|force : Overwrite existing files with generated files.}
+
+        {--uninspiring : Disable inspirational quotes}
+    ';
+
+    /**
+     * @var string The console command description.
+     */
+    protected $description = 'Creates a new factory.';
+
+    /**
+     * @var array List of commands that this command replaces (aliases)
+     */
+    protected $replaces = [
+        'make:factory',
+    ];
+
+    /**
+     * @var string The type of class being generated.
+     */
+    protected $type = 'Factory';
+
+    /**
+     * @var string The argument that the generated class name comes from
+     */
+    protected $nameFrom = 'factory';
+
+    /**
+     * @var array A mapping of stubs to generated files.
+     */
+    protected $stubs = [
+        'scaffold/factory/factory.stub' => 'database/factories/{{studly_name}}.php',
+    ];
+
+
+    protected function processVars($vars): array
+    {
+        $vars = parent::processVars($vars);
+
+        $vars['model'] = $this->option('model');
+
+        return $vars;
+    }
+}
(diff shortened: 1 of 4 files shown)


modules/system/console/CreateFactory.php
```diff
@@ -0,0 +1,62 @@
...
+    protected $nameFrom = 'factory';
+
+    /**
+     * @var array A mapping of stubs to generated files.
+     */
+    protected $stubs = [
+        'scaffold/factory/factory.stub' => 'database/factories/{{studly_name}}.php',
+    ];
+
+
```
LukeTowers:
```suggestion
```

modules/system/console/CreateFactory.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * @var string The argument that the generated class name comes from
+     */
+    protected $nameFrom = 'factory';
+
+    /**
+     * @var array A mapping of stubs to generated files.
+     */
+    protected $stubs = [
+        'scaffold/factory/factory.stub' => 'database/factories/{{studly_name}}.php',
```
LukeTowers, line 50: So this creates a factory class inside of the project level `database/factories` folder?

sephyld: Hey there!

Nope, it creates the factory class in the _plugin's_ `database/factories` folder. I've tried it out with the demo plugin and the Winter.Notes plugin.

For example, using
```sh
php artisan make:factory Winter.Notes NoteFactory -m Note
```
The result is

```
File generated: /plugins/winter/notes/database/factories/NoteFactory.php
Factory created successfully.
  
  “ People find pleasure in different ways. I find it in keeping my mind clear. ”
  — Marcus Aurelius
```
And the created file's content is this
```php
<?php

namespace Winter\Notes\Database\Factories;

use Illuminate\Database\Eloquent\Factories\Factory;

/**
 * NoteFactory Factory
 * @extends \Illuminate\Database\Eloquent\Factories\Factory<\Winter\Notes\Models\Note>
 */
class NoteFactory extends Factory
{
    /**
     * Define the model's default state.
     *
     * @return array<string, mixed>
     */
    public function definition()
    {
        return [
            //
        ];
    }
}

```

LukeTowers: @gseph do you have any thoughts on putting it in the plugin's `factories` folder instead of `database/factories`? Although I suppose if we kept the `database` folder we could eventually maybe use it for squashing plugin migrations...

sephyld: I don't have a hard-set opinion on the placement of factories files, I just wanted to mimic the laravel folder structure, so I've no problem changing it.

However, now that you mention the chance to start using the `database` folder for plugin migrations in the future, I think it would also be a good idea to use it for seeders too considering that developing big plugins leads to the presence of a lot of different directories nested in the plugin's directory. Not a big deal to be fair, but something worth considering imho.

Would that be a breaking change though? Maybe, I'm not sure about that.

Probably the best course of actions right now is to simply generate the factories in the `factories` directory like you hinted. This way factories files placement stays consistent with models and seeders placement.

@LukeTowers what do you think?

LukeTowers: Plugin migrations will probably remain in the updates folder, but I could see an argument being made for storing factories, seeders, and schema dumps in the database folder.

Right now seeders are run as migrations in the updates folder if you want to use them, so I think keeping this as the database/factories folder for us to expand the usage of the database folder in the future is probably best.

@bennothommo @jaxwilko @mjauvin any thoughts?

jaxwilko: @LukeTowers I like it with the idea that moving forward at least seeders and factories would live inside `database`. +1 from me :)

PR #1210, opened 2024-09-23
Improved French translation

Quendi6:
### Issue:
When current language is French, front-end users may think they are using the wrong username when they are simply using the wrong password to log in. The reason is that `'L\'utilisateur saisi ne correspond à aucun utilisateur enregistré. Merci de vérifier votre saisie et de réessayer.'` means `'The user entered does not match any registered user. Please check your entry and try again.'`, which is not the true meaning of this error.

### How to reproduce:
You have to set French as the front-end language, have `Winter.User` configured well with a login page, and debug mode set to false. Then, try to log in with a correct front-end username, but a wrong password. I know this happens with a plugin and not with core `Winter`, but this plugin is using **`Winter\Storm\Auth\AuthenticationException` which uses the translation reference `backend::lang.auth.invalid_login`, both part of core.**

### Solution:
The French translation `'invalid_login' => 'Les détails saisis ne correspondent à aucun de nos enregistrements. Merci de vérifier et de réessayer.'` is more accurate to the original English sentence `'invalid_login' => 'The details you entered did not match our records. Please double-check and try again.'`.

### Result:
After this fix, French users are encouraged to verify all their entries, like English ones do.

### DIFF
--- a/modules/backend/lang/fr/lang.php
+++ b/modules/backend/lang/fr/lang.php
@@ -3,7 +3,7 @@
 return [
     'auth' => [
         'title' => 'Zone d\'administration',
-        'invalid_login' => 'L\'utilisateur saisi ne correspond à aucun utilisateur enregistré. Merci de vérifier votre saisie et de réessayer.'
+        'invalid_login' => 'Les détails saisis ne correspondent à aucun de nos enregistrements. Merci de vérifier et de réessayer.'
     ],
     'field' => [
         'invalid_type' => 'Type de champ invalide :type.',


Quendi6:
I updated my request with more details.

bennothommo:
@Quendi6 to be honest, I'd be more inclined to change the English message to something like "Unable to sign in. Please check your username and password." and change the French translation to suit. Keep it really basic and not imply anything about user "records". Any thoughts @LukeTowers?

LukeTowers:
@bennothommo I like how it is now, a username and password will not always be the method of authentication (email, SSO, etc). We don't necessarily have to say "records" but the intent is to communicate that the information that was provided was unable to provide the user with the requested access.

Quendi6:
@LukeTowers I agree with that, English message is good to me. French translation need to suit.

PR #1224, opened 2024-10-08
Improve Tag List default value and useKey option support

### LINKED ISSUE #1223: TagList not saving an empty value
It appears that a `TagList` widget in `relation` mode is not saving an empty value, because it no longer sends a POST value when empty.

### Replication steps

In the Test plugin:

- Go to **Users**, edit either user.
- Click **Tag List Field** in the top alternative views box.
- Empty the **Roles** field, then try and save.

It does not fail validation, so the form saves, as the last value in the tag list is kept on refresh even though all tags are removed.

bennothommo:
Fixes #1223.

### DIFF
--- a/modules/backend/formwidgets/taglist/partials/_taglist.php
+++ b/modules/backend/formwidgets/taglist/partials/_taglist.php
@@ -1,6 +1,6 @@
 <?php
 $selectedValues = is_array($selectedValues) ? $selectedValues : [];
-$availableOptions = $useKey ? $fieldOptions : array_unique(array_merge($selectedValues, $fieldOptions));
+$availableOptions = $fieldOptions;
 $displayOnlyOptions = [];
 
 foreach ($availableOptions as $key => $option) {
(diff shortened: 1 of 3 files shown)


modules/backend/formwidgets/taglist/partials/_taglist.php
```diff
@@ -19,20 +19,11 @@
...
-            type="hidden"
-            name="<?= $field->getName() ?>[]"
-            value="<?= $option ?>">
-        <?php endforeach ?>
-    <?php else: ?>
-        <input
-            type="hidden"
-            name="<?= $field->getName() ?>"
-            value="<?= $field->value ?>">
-    <?php endif ?>
```
LukeTowers, line 34: Why was this removed? Isn't it needed for trigger support in preview / readonly contexts?

bennothommo: Ah so that's what it's for. I'll re-add.

PR #1226, opened 2024-10-08
Support --isolatable flag on migration command

### LINKED ISSUE #1225: Winter migrations are not concurrency-friendly
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

Laravel's console `Command` implementation allows commands to run in isolation. Essentially it means an isolated command can't have more than one active process. A typical use case is when there's some parallelism in a system and command is not mean to be run multiple times let alone in parallel. Database migration is another example when isolation is needed.

Normally, in a single instance deployment this is not an issue. However, Winter CMS may be used in a multi-server setup, when some CI/CD tool runs a parallel deployment (e.g. using Kubernetes etc), leading to `winter:up` command being called automatically more than once.

The simplest solution is to rely on the built-in support and `--isolated` option to the migration command which can be achieved by just implementing `Isolatable` interface (which has no methods at all).

### Steps to replicate

1. Create a non-idempotent migration (like "CREATE TABLE ...")
2. Run it in a multi-server deployment, making sure it's executed in parallel.
3. Observe the failure of the 2nd and other instances who attempted to run the migration (because they all started at the same time and "thought" they had to run it)

### Workaround

_No response_

lex0r:
Fixes https://github.com/wintercms/winter/issues/1225

### DIFF
--- a/modules/system/console/WinterUp.php
+++ b/modules/system/console/WinterUp.php
@@ -1,6 +1,7 @@
 <?php namespace System\Console;
 
 use Illuminate\Console\Command;
+use Illuminate\Contracts\Console\Isolatable;
 use System\Classes\UpdateManager;
 
 /**
@@ -11,7 +12,7 @@ use System\Classes\UpdateManager;
  * @package winter\wn-system-module
  * @author Alexey Bobkov, Samuel Georges
  */
-class WinterUp extends Command
+class WinterUp extends Command implements Isolatable
 {
     /**
      * The console command name.


bennothommo:
@lex0r would you be willing to provide a PR to the [docs](https://github.com/wintercms/docs) explaining this usage? Namely, explaining that there needs to be an external cache set up (and cache likely cannot be `database` or `file`) and you need to provide the `--isolated` flag in order to use isolated migrations.

PR #1228, opened 2024-10-10
Allow dynamic methods to be used for custom list column types

mjauvin:
This will allow using the following code to add custom column types:
```php
\Backend\Widgets\Lists::extend(function () {
    $this->addDynamicMethod('evalMyCustomColumnTypeValue', function ($record, $column, $value) {
        return strtoupper($value);
    });
}, true);


bennothommo (approved):

LukeTowers:
@mjauvin can you add something on this to the docs?

PR #1229, opened 2024-10-10
Make entire fileupload field clickable in single file mode

mjauvin:
Before:
image
After:
image
image


LukeTowers (approved):
This is one of those things that's been bugging me for a while, thanks @mjauvin ❤️

PR #1234, opened 2024-10-18
Enable framework extras in backend

### LINKED ISSUE #1233: Framework extras missing in backend
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

Trying to use a button in a backend view with a `data-attach-loading` does not work. 

Note: this works fine on a frontend page

### Steps to replicate

Add the following button to a backend controller:
```html
<button data-attach-loading data-request=onAjax>TEST</button>
```

### Workaround

_No response_

mjauvin:
Fixes #1233

### DIFF
--- a/modules/backend/layouts/_head.php
+++ b/modules/backend/layouts/_head.php
@@ -29,7 +29,8 @@ foreach ($styles as $style) {
 $scripts = [
     Backend::skinAsset('assets/js/vendor/jquery.min.js'),
     Backend::skinAsset('assets/js/vendor/jquery-migrate.min.js'),
-    Url::asset('modules/system/assets/js/framework-combined-min.js'),
+    Url::asset('modules/system/assets/js/framework.js'),
+    Url::asset('modules/system/assets/js/framework.extras.js'),
     Url::asset('modules/system/assets/js/build/manifest.js'),
     Url::asset('modules/system/assets/js/snowboard/build/snowboard.vendor.js'),
     Url::asset(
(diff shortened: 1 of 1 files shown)


LukeTowers:
@mjauvin does this cause any duplication / double handling of existing backend functionality?

mjauvin:
@LukeTowers I checked thoroughly and there is no duplication whatsoever.

mjauvin:
I just realized there may be a need to also include the extras css as well... will check asap and report back here.

mjauvin:
Ok, so the extras styles are already contained in `snowboard.extras.css`, so no need here.

mjauvin:
@bennothommo what was the reason you included `snowboard.extras.css` in the backend ?

bennothommo:
@mjauvin the backend build of Snowboard does use some utilities that need styling (Flash, Stripe Loader, Transition).

Also, I'm thinking it might be better that we try and hook up the original AJAX framework's events to Snowboard, and use the Snowboard Attach Loading utility instead. Should hopefully be doable.

bennothommo:
I've submitted an alternate way of handling this - #1235. I've seen one plugin manually including the `framework.extras` (Test plugin), so I'm leaning towards it was only meant to be included when needed and not globally, but up to you guys if we want to hook it into Snowboard or just use @mjauvin's PR.

LukeTowers:
@jaxwilko any thoughts?

mjauvin:
replaced by #1235

PR #1235, opened 2024-10-21
Attach original AJAX framework events to Snowboard in Backend

### LINKED ISSUE #1233: Framework extras missing in backend

bennothommo:
Replaces #1234, fixes #1233.

Instead of loading `framework.extras.css`, which appears to be excluded from the Backend, this PR hooks the necessary original AJAX framework events into Snowboard's events, allowing the applicable events to execute UI functionality such as `data-attach-loading` and the Stripe loader.

This is technically a stop-gap solution for now, since it doesn't emulate a full Snowboard request, so there's the potential for custom Snowboard event listeners to break if they are relying on Snowboard request information, but I'm not aware of that happening yet in core.


mjauvin:
Nice!

modules/backend/assets/ui/js/ajax/Handler.js
```diff
@@ -46,6 +50,42 @@ export default class Handler extends Snowboard.Singleton {
...
+                options: {},
+            });
+        });
+        document.addEventListener('$ajaxFail', (event) => {
+            this.requests[event.target].reject(event.detail.textStatus);
+            this.snowboard.globalEvent('ajaxDone', event.detail.data, {
+                element: event.target,
+                options: {},
+            });
+        });
```
LukeTowers, line 92: Can you add this functionality to the functions comment at the top of the class?

LukeTowers: @bennothommo ping

bennothommo: Cheers for the ping @LukeTowers. I've added it

PR #1239, opened 2024-10-26
Add routePersistence parameter to Page::resolveMenuItem()

mjauvin:
This permits reusing Cms Page resolveMenuItem with or without route persistence for the call to pageUrl().

The default behavior is unchanged as we default to false for route persistence.

### DIFF
--- a/modules/cms/classes/Page.php
+++ b/modules/cms/classes/Page.php
@@ -185,7 +185,7 @@ class Page extends CmsCompoundObject
      *
      * @param \Winter\Sitemap\Classes\DefinitionItem|\Winter\Pages\Classes\MenuItem $item Specifies the menu item.
      */
-    public static function resolveMenuItem(object $item, string $url, Theme $theme): ?array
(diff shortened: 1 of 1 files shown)


LukeTowers:
Can you provide your usage example here along with why you needed routePersistence enabled for your use case?

mjauvin:
@LukeTowers On the `Project` sub-section of this website, the menubar changes and all pages depend on the project slug, so their `url` is `project/:slug/page` (e.g. `project/:slug/about`, `project/:slug/team`, etc.)

In order to render the menu for those pages, the slug in their url needs to inherit the project slug, hence the route persistence.

mjauvin:
With this PR, I can use the following in my component:
```php
Event::listen('pages.menuitem.resolveItem', function ($type, $item, $url, $theme) {
    if ($type === 'cms-page' && str_starts_with($item->reference, 'project/')) {
        return \Cms\Classes\Page::resolveMenuItem($item, $url, $theme, true);
    }
}, 10000);
```

mjauvin:
Without this PR, I need to duplicate the resolveMenuItem method in my component.

PR #1244, opened 2024-10-31
Automate Mix installation and compilation for Tailwind themes 

ghost:
### Wintercms
 dev-develp

**PHP 2.2.24**
**MySqs/MarianDB 8.0.30**

_This PR adds automatic Mix installation and asset compilation when creating a new theme with Tailwind scaffold._

Changes:
- Added automatic `mix:install` execution with preset "yes" responses
- Added automatic `mix:compile` execution after theme creation
- Improved user experience by removing manual steps
- Added proper error handling and user feedback

This resolves the issue where users had to manually run mix:install and mix:compile after creating a Tailwind theme.

Testing:
1. Create a new theme with Tailwind scaffold
2. Theme is created with all dependencies installed
3. Assets are automatically compiled
4. No manual commands required


modules/cms/console/CreateTheme.php
```diff
@@ -77,6 +78,70 @@ class CreateTheme extends GeneratorCommand
...
+                    $output = stream_get_contents($pipes[1]);
+                    $this->info($output);
+                    
+                    // Fehler lesen
+                    $errors = stream_get_contents($pipes[2]);
+                    fclose($pipes[1]);
+                    fclose($pipes[2]);
+                    
+                    proc_close($process);
+                }
```
LukeTowers, line 124: Is this approach used instead of Artisan::call() because the mix:install command is interactive? Could we instead add support for a `--force` option on `mix:install`? @jaxwilko any thoughts?

josephcrowell: It would be better to have it as an option rather than assuming people want to install Mix/Vite right at this point.

LukeTowers: @josephcrowell what situation can you think of when someone scaffolds a tailwind based theme and doesn't want to have the asset compilation working on it?

ghost: > Is this approach used instead of Artisan::call() because the mix:install command is interactive? Could we instead add support for a `--force` option on `mix:install`? @jaxwilko any thoughts?

on my local machine ite mix:install is automaticlly not interactive   

@josephcrowell i didn't know a situation to use scaffolds tailwind without this compilation but you can "create" a exta button for tailwind without mix - this is easy   

Today I will finish scaffolding tailwind with & vue

josephcrowell: Ah right this is when the dev is compiling/creating the theme... Never mind.

jaxwilko: You could run the command the same way it's done in the tests. Laravel creates a mock output which simulates questions being asked and allows you to answer them:
```php
(new \Illuminate\Testing\PendingCommand(new \System\Tests\Bootstrap\TestCase(), $this->getLaravel(), 'mix:install', []))
    ->expectsQuestion('Detected winter.blocks (plugins/winter/blocks), should it be added to your package.json?', true)
    ->expectsQuestion('Detected winter.sentry (plugins/winter/sentry), should it be added to your package.json?', true);
```
However this is really difficult to work with as depending on the packages installed by the user you may get different results...

The other option is to call `mix:install --no-interaction` which would just auto accept all the options which I think is what you want:
```php
Artisan::call('mix:install', [
    '--no-interaction' => true
]);
```

jaxwilko:
It might be nice to default to `vite` instead of `mix` as `vite` is now the default in Laravel, having both as optional flags and defaulting to one would be ideal :)

PR #1248, opened 2024-11-04
Media Manager upload size and file name validation

Denoder:
This is more of a getting started guideline as I'm not sure how to proceed with the changes or if they're good enough quality.
The first issue I encountered is the fact that by default `mediamanager.js` will have a max filesize of 256MB (dropzone default) and cant be increased any further.

2nd issue was actually trying to find why an error was popping up which lead to the first issue. When the file size is too big it just gives a console error. As it turns out the xhr property doesn't exist within the context so i had to make it conditional so it'd go to the actual popup.

3rd issue (or rather more of a preference) is that in some cases I'd prefer to keep the file names intact without any cleanup. So I made it so that you can optionally disable file name validation cleanup.

I'm not that familiar with dropzone, so im unsure how to integrate filesizes with winter so it works like how the fileupload widget does for now it's set to `0`. There also came an issue where because it's using `post_max_size` to upload a file to media manager, a memory exhaustion error popped up, and rather than having to globally set the memory limit higher, it would be better to use `ini_set` temporarily.

### DIFF
--- a/modules/backend/traits/UploadableWidget.php
+++ b/modules/backend/traits/UploadableWidget.php
@@ -27,6 +27,11 @@ trait UploadableWidget
     // public $uploadPath;
 
     /**
+     * @var bool $validateFileName Determines whether the file name should be validated.
+     */
+    public bool $validateMediaFileName = true;
+
+    /**
      * Returns the disk that will be used to store the uploaded file
      */
     public function uploadableGetDisk(): FilesystemAdapter
@@ -171,7 +176,7 @@ trait UploadableWidget
         /*
          * File name contains non-latin characters, attempt to slug the value
          */
-        if (!$this->validateFileName($fileName)) {
+        if (!$this->validateFileName($fileName) && $this->validateMediaFileName) {
             $fileName = $this->cleanFileName(File::name($fileName)) . '.' . $extension;
         }
 
--- a/modules/backend/widgets/MediaManager.php
+++ b/modules/backend/widgets/MediaManager.php
@@ -62,6 +62,7 @@ class MediaManager extends WidgetBase
     {
         $this->alias = $alias;
         $this->readOnly = $readOnly;
+        $this->validateMediaFileName = Config::get('cms.storage.media.validateMediaFileName', true);
 
         parent::__construct($controller, []);
     }
@@ -108,6 +109,32 @@ class MediaManager extends WidgetBase
     //
     // AJAX handlers
     //
+
+    /**
+     * Process file uploads submitted via AJAX
+     *
+     * @throws ApplicationException If the file "file_data" wasn't detected in the request or if the file failed to pass validation / security checks
+     */
+    public function onUpload(): ?\Illuminate\Http\Response
+    {
+        ini_set('memory_limit', Config::get('cms.storage.media.memoryLimit', ini_get('memory_limit')));
+
+        if ($this->readOnly) {
+            return null;
+        }
+
+        /**
+         * @event backend.widgets.uploadable.onUpload
+         * Provides an opportunity to process the file upload using custom logic.
+         *
+         * Example usage ()
+         */
+        if ($result = Event::fire('backend.widgets.uploadable.onUpload', [$this], true)) {
+            return $result;
+        }
+
+        return $this->onUploadDirect();
+    }
 
     /**
      * Perform a search with the query specified in the request ("search")
--- a/modules/backend/widgets/mediamanager/assets/js/mediamanager.js
+++ b/modules/backend/widgets/mediamanager/assets/js/mediamanager.js
@@ -838,7 +838,7 @@
     MediaManager.prototype.uploadError = function(file, message) {
         this.updateUploadBar('error', 'progress-bar progress-bar-danger');
 
-        if (file.xhr.status === 413) {
+        if (file?.xhr?.status === 413) {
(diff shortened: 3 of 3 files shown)


modules/backend/widgets/mediamanager/assets/js/mediamanager.js
```diff
@@ -732,7 +732,8 @@
             paramName: 'file_data',
             timeout: 0,
             headers: {},
-            createImageThumbnails: false
+            createImageThumbnails: false,
+            maxFilesize: this.options.maxFilesize,
```
LukeTowers, line 736: We should probably just populate this from an ini_get(post_max_size) instead

modules/backend/widgets/mediamanager/assets/js/mediamanager.js
```diff
@@ -837,7 +838,7 @@
     MediaManager.prototype.uploadError = function(file, message) {
         this.updateUploadBar('error', 'progress-bar progress-bar-danger');
 
-        if (file.xhr.status === 413) {
+        if (file?.xhr?.status === 413) {
```
LukeTowers, line 841: What does this change accomplish?

modules/backend/widgets/MediaManager.php
```diff
@@ -109,6 +110,32 @@ public function render(): string
...
     //
 
+    /**
+     * Process file uploads submitted via AJAX
+     *
+     * @throws ApplicationException If the file "file_data" wasn't detected in the request or if the file failed to pass validation / security checks
+     */
+    public function onUpload(): ?\Illuminate\Http\Response
+    {
+        ini_set('memory_limit', Config::get('cms.storage.media.memoryLimit', ini_get('memory_limit')));
```
LukeTowers, line 120: Why do you have this here? Why not just put this code in the event listener instead if you really need it for your application? Why do you even need to increase the memory limit here in the first place?

modules/backend/traits/UploadableWidget.php
```diff
@@ -171,7 +176,7 @@ public function validateMediaFileName(string $fileName, string $extension): stri
         /*
          * File name contains non-latin characters, attempt to slug the value
          */
-        if (!$this->validateFileName($fileName)) {
+        if (!$this->validateFileName($fileName) && $this->validateMediaFileName) {
```
LukeTowers, line 179: Why would we want to allow a filename to skip validation? The validation exists for a reason.

modules/backend/widgets/mediamanager/assets/js/mediamanager.js
Denoder: When an error relating to the filesize by dropzone was coming up it was trying to open an alert in `$.wn.alert(message)` but this was preventing it because `file.xhr` didn't exist

modules/backend/widgets/MediaManager.php
Denoder: I had explained earlier when i was uploading large files I got a memory exhausted error (this was after fixing the error above). I can move it out if it works in the event listener (at the same time i was unsure if the event listener even work because everything else uses `$this->fireSystemEvent` that I thought that Event was unused.

LukeTowers: A memory exhausted error feels more like the image is being processed in some way, I don't think I've seen it happen just from a file being uploaded before. I'd need more evidence that this is actually a problem that affects most Winter users before I'd want it in the core, you should be able to implement for your own project just with the existing event listener for now though.

modules/backend/traits/UploadableWidget.php
Denoder: When I'm adding multiple files and their have name formatting that i use elsewhere in a plugin so they can be associated with a relation for upload after batch uploading (An argument can be made to use fileupload, but the batch uploading is for individual relations that use file upload not just one using multiupload). I made it opt-out, but I do need it to keep the name formatting.

modules/backend/widgets/mediamanager/assets/js/mediamanager.js
Denoder: I am unsure how i'd proceed with this part.

modules/backend/traits/UploadableWidget.php
LukeTowers: I wouldn't recommend abusing the media library system for that. I'd need to know more about your exact use case, but I could see you using an attachMany to accept multiple files and then re-assigning them to the relevant attachOne relationships post upload. Also why can't you have the files renamed when they get moved out of the medialibrary?

modules/backend/widgets/mediamanager/assets/js/mediamanager.js
LukeTowers: Make sure you're using the [latest version of winter/storm](https://github.com/wintercms/storm/commit/8675a200e88fa03ba66696e54afacc0ee7caa5fd) and add `data-max-filesize="<?= e(\File::getMaxUploadSize()); ?>"` to https://github.com/wintercms/winter/blob/develop/modules/backend/widgets/mediamanager/partials/_body.php#L13

modules/backend/traits/UploadableWidget.php
Denoder: I decided to create a clone of the mediamanager as a plugin/widget rather than using that media manager.

PR #1252, opened 2024-11-15
Add event logs deletion with filters

damsfx:
While the event log list allows you to delete events in a bulk fashion by filtering the list with a keyword, then selecting all the events and clicking on the "Delete selected" button, if the log includes more events responding to the query than the list can display, you need to repeat the action several times to delete all the events.

This PR provides a new button to display a pop-up window to the user, allowing him to enter a search term and date range to obtain the set of events matching the criteria, which will be deleted after confirmation.

image

image


LukeTowers:
@damsfx what do you think about instead adding date filters to the list itself and then making a PR to add support for a special "select all" mode that allows you to select all records that match the current query (like gmail has)?

Untitled 3
Untitled 2

LukeTowers:
Could even make the select all text even more useful by adding a count to it ("Select all records that match this search (402)")

damsfx:
@LukeTowers This would be a good improvement for the event manager and certainly the best UI/UX.

My tests in this direction are meeting several challenges: 
- The list search field is too generic and searches the different model properties.
- The search field doesn't allow searching for special characters like `%`.

In addition, for the selection of items not displayed in the list, what do you think is the best solution for adding these capabilities to the controller? 
- Add them to `\Backend\Behaviors\ListController`.
- Create a new type of list controller
- Add them only to `\System\Controllers\EventLogs`.

LukeTowers:
The search behaviour can be configured by setting `searchable: false` on columns that you don't want to be searched on. 

I'd prefer the expanded selection logic to be added to the ListController behavior, it would be useful in pretty much every list.

damsfx:
Special "select all" introduce in PR https://github.com/wintercms/winter/pull/1542

PR #1255, opened 2024-11-25
Tailwind theme create improvements

jaxwilko:
This PR removes the old css/js scaffold files and adds better error reporting for creating a vite theme.

This includes missing / outdated NPM detection:
image

And feedback to the user if any command should fail:
image

I also added better support for extension, in the future we can look into adding mix support too via passing the processor here:
image

The new `npm:version` command will by default just report the version to the user, if you want to use it to test if the version is compatible with Winter you need to pass the `--compatible` flag and check the exit code. See:
image


jaxwilko:
Ideally, we should validate the NPM version prior to generating the initial theme structure, as if it fails we have no way to reset the state which means that the user will not be able to "try again" as the dir already exists.

LukeTowers (approved):

PR #1257, opened 2024-11-27
Move after login logic to event listener instead of controller

mjauvin:
User model now handles running the migrations in its afterLogin() method.

### DIFF
--- a/modules/backend/ServiceProvider.php
+++ b/modules/backend/ServiceProvider.php
@@ -33,6 +33,7 @@ class ServiceProvider extends ModuleServiceProvider
         $this->registerMailer();
         $this->registerAssetBundles();
         $this->registerBackendPermissions();
+        $this->registerBackendUserEvents();
 
         /*
          * Backend specific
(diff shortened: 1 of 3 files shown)


bennothommo:
This is burying that sort of functionality deep, in a place I wouldn't think to look if I was looking for it.

Could we not just add a listener in the Backend's service provider for the `backend.user.login` event and just fire the functionality there?

Or even better, begone with it completely. Migrations should be explicitly run via an action (installing plugins, running the console command) - not arbitrarily when logging in.

mjauvin:
> This is burying that sort of functionality deep, in a place I wouldn't think to look if I was looking for it.

That's why they invented "grep"! :stuck_out_tongue_closed_eyes: 

> Could we not just add a listener in the Backend's service provider for the `backend.user.login` event and just fire the functionality there?

I don't mind, as long as we all agree.

> Or even better, begone with it completely. Migrations should be explicitly run via an action (installing plugins, running the console command) - not arbitrarily when logging in.

The main reason to keep the functionality is for users without shell access: they can run the migrations just by login into the backend.

Personally, I don't care about this feature.

LukeTowers:
Let's move it to the service provider in an event listener for now with the eventual plan to remove it once we have the updates / installation system working more reliably and consistently.

mjauvin:
How's that?

LukeTowers (approved):

PR #1258, opened 2024-11-27
Remove redundant backend route

mjauvin:
I tested this on nginx and apache on both Laravel 9 and Laravel 11 by making sure the route cache was cleared.

No issues so far.

### DIFF
--- a/modules/backend/routes.php
+++ b/modules/backend/routes.php
@@ -20,7 +20,7 @@ Event::listen('system.route', function () {
     Event::fire('backend.beforeRoute');
 
     /*
-     * Other pages
+     * Entry point
      */
     Route::group([
         'middleware' => ['web'],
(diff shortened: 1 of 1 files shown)


LukeTowers:
@mjauvin does this still work if backendUri is set to an empty string? `''`

mjauvin:
> @mjauvin does this still work if backendUri is set to an empty string? `''`

Yes, that works too.

modules/backend/routes.php
```diff
@@ -20,7 +20,7 @@
     Event::fire('backend.beforeRoute');
 
     /*
-     * Other pages
+     * Entry point
```
bennothommo: Let's describe what is actually happening so it's clear
```suggestion
     * Route everything to `Backend\Classes\BackendController`, which itself acts as a router for the Backend.
```

mjauvin:
@LukeTowers care to merge this so we can test it in develop at a larger scale before we release 1.2.8 ?

PR #1259, opened 2024-11-29
[WIP] Update Manager Refactor

jaxwilko:
Me and Luke are working on a little refactor of how the UpdateManager works...

It also includes some components from my old composer PR found here: https://github.com/wintercms/winter/pull/967

The core idea of this is that there should be consistent interfaces to interact with different things built on top of Winter (i.e. `modules`, `plugins` & `themes`). Each of these elements will now have their class (i.e. `Theme`, `PluginBase`, `ModuleServiceProvider`) will now extend `WinterExtension` which will allow them to all interact with the update manager and other system functions directly.

In addition `PluginManager`, `ThemeManager` & `ModuleManager` (new) will all implement the `ExtensionManager` interface, which will allow for a consistant way to interact with all Winter extensions.

I.e. You will be able to call `PluginManager::instance()->update($plugin)` or `ThemeManager::instance()->update($theme)`.

This also allows us to do some cool stuff like (only support for plugin installing is currently functional)
```php
(new ExtensionSource(
    ExtensionSource::SOURCE_COMPOSER,
    ExtensionSource::TYPE_PLUGIN,
    composerPackage: 'jaxwilko/datamigrator'
))->install();

(new ExtensionSource(
    ExtensionSource::SOURCE_COMPOSER,
    ExtensionSource::TYPE_THEME,
    composerPackage: 'jaxwilko/custom-theme'
))->install();

(new ExtensionSource(
    ExtensionSource::SOURCE_LOCAL,
    ExtensionSource::TYPE_PLUGIN,
    code: 'winter.demo'
))->install();

(new ExtensionSource(
    ExtensionSource::SOURCE_LOCAL,
    ExtensionSource::TYPE_PLUGIN,
    path: 'plugins/winter/demo'
))->install();
```

By the end of the PR, the UpdateManager may no longer exist as is, instead each `ExtensionManager` will be responsible for its existing functionality.

Requires storm branch: https://github.com/wintercms/storm/tree/wip/support-uploading-packages-update-manager-changes

## Summary by CodeRabbit

* **New Features**
  * Added marketplace UI for browsing and installing plugins and themes directly.
  * Added support for uploading plugins and themes as ZIP files.
  * Integrated marketplace API for enhanced package management and discovery.

* **Improvements**
  * Redesigned plugin and theme management interface with improved navigation and search capabilities.
  * Updated system updates page with streamlined extension installation workflow.


composer.json
```diff
@@ -29,13 +29,14 @@
         "source": "https://github.com/wintercms/winter"
     },
     "require": {
-        "php": "^8.0.2",
+        "php": "^8.1",
```
LukeTowers: What's the PHP 8.1 feature that we're using?

jaxwilko: Interface const overriding (see: https://www.php.net/manual/en/language.oop5.interfaces.php), but we're not using it any more so could rollback that change

modules/backend/ServiceProvider.php
```diff
@@ -322,4 +323,14 @@ protected function registerBackendSettings()
             $manager->registerOwnerAlias('Winter.Backend', 'October.Backend');
         });
     }
+
+    public function getPath(): string
+    {
+        return __DIR__;
```
LukeTowers: Could probably move this into the base ModuleServiceProvider with the following:

```php
public function getPath() {
        $reflector = new \ReflectionClass(get_called_class());
        return dirname($reflector->getFileName());
    }
```

modules/backend/ServiceProvider.php
```diff
@@ -322,4 +323,14 @@ protected function registerBackendSettings()
...
+    }
+
+    public function getIdentifier(): string
+    {
+        return 'Backend';
```
LukeTowers: Could probably move this into the base ModuleServiceProvider with the following:

```php
public function getIdentifier(): string
{
    $reflector = new \ReflectionClass(get_called_class());
    return $reflector->getNamespaceName();
}
```

modules/cms/ServiceProvider.php
```diff
@@ -477,4 +491,14 @@ protected function registerHalcyonModels()
...
+
+    public function getPath(): string
+    {
+        return __DIR__;
+    }
+
+    public function getIdentifier(): string
+    {
+        return 'Cms';
+    }
```
LukeTowers: Same as Backend ServiceProvider comments

modules/system/classes/extensions/ExtensionManager.php
```diff
@@ -0,0 +1,104 @@
...
+    final public static function forgetInstance(?Container $container = null): void
+    {
+        if (!$container) {
+            $container = app();
+        }
+
+        if ($container->bound(static::class)) {
+            $container->forgetInstance(static::class);
+        }
+    }
```
LukeTowers, line 86: Why not use the Singleton trait?

modules/system/classes/extensions/ExtensionManagerInterface.php
```diff
@@ -0,0 +1,131 @@
...
+     * @return mixed
+     */
+    public function uninstall(WinterExtension|string $extension, bool $noRollback = false, bool $preserveFiles = false): mixed;
+
+    /**
+     * Completely uninstall all extensions managed by this manager
+     *
+     * @return static
+     */
+    public function tearDown(): static;
```
LukeTowers, line 134: Seems dangerous 😂 What's it used for?

modules/system/classes/extensions/ModuleManager.php
```diff
@@ -0,0 +1,283 @@
...
+     * @return array<string, WinterExtension>
+     */
+    public function list(): array
+    {
+        return array_merge(...array_map(fn($key) => [$key => $this->get($key)], Config::get('cms.loadModules', [])));
+    }
+
+    public function create(string $extension): WinterExtension
+    {
+        throw new ApplicationException('Support for creating extensions needs implementing');
```
LukeTowers:
```suggestion
        throw new ApplicationException('Support for creating modules is not implemented.');
```

modules/system/classes/UpdateManager.php
```diff
@@ -39,178 +31,44 @@ class UpdateManager
...
-     * These objects are "soft singletons" and may be lost when
-     * the IoC container reboots. This provides a way to rebuild
-     * for the purposes of unit testing.
-     */
-    public function bindContainerObjects()
+    public function isSystemSetup(): bool
     {
-        $this->migrator = App::make('migrator');
-        $this->repository = App::make('migration.repository');
+        return !Schema::hasTable($this->getMigrationTableName());
```
LukeTowers:
```suggestion
        return Schema::hasTable($this->getMigrationTableName());
```

modules/system/classes/extensions/ModuleManager.php
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    /**
+     * @throws ApplicationException
+     */
+    public function update(WinterExtension|string|null $extension = null, bool $migrationsOnly = false): ?bool
+    {
+        $modules = $this->getModuleList($extension);
+
+        $firstUp = UpdateManager::instance()->isSystemSetup();
```
LukeTowers:
```suggestion
        $firstUp = !UpdateManager::instance()->isSystemSetup();
```

modules/system/classes/extensions/ModuleManager.php
```diff
@@ -0,0 +1,283 @@
...
+
+        if ($firstUp) {
+            $this->repository->createRepository();
+            $this->output->info('Migration table created');
+        }
+
+        if (!$migrationsOnly && !Config::get('cms.disableCoreUpdates')) {
```
LukeTowers: We should probably include an info or warning message if `cms.disableCoreUpdates` is set; I used to use it on every site to prevent the backend from updating the core when I was actually managing them with composer so I could see it causing issues if we silently ignore them using that setting.

modules/system/classes/extensions/ModuleManager.php
```diff
@@ -0,0 +1,283 @@
...
+                    && Composer::updateAvailable($composerPackage)
+                ) {
+                    $this->output->info(sprintf(
+                        'Performing composer update for %s (%s) module...',
+                        $module,
+                        $composerPackage
+                    ));
+
+                    Preserver::instance()->store($extension);
+                    $update = Composer::update(dryRun: true, package: $composerPackage);
```
LukeTowers: Is the dryrun here a temporary thing?

modules/system/classes/extensions/ModuleManager.php
```diff
@@ -0,0 +1,283 @@
...
+                    ($composerPackage = Composer::getPackageNameByExtension($extension))
+                    && Composer::updateAvailable($composerPackage)
+                ) {
+                    $this->output->info(sprintf(
+                        'Performing composer update for %s (%s) module...',
+                        $module,
+                        $composerPackage
+                    ));
+
+                    Preserver::instance()->store($extension);
```
LukeTowers, line 113: Is the intention to use this to restore the previous version if something fails, or perhaps to allow users to rollback to a specific version when it's been preserved?

modules/system/classes/extensions/ModuleManager.php
```diff
@@ -0,0 +1,283 @@
...
+        }
+
+        foreach ($modules as $module) {
+            $this->output->info(sprintf('Migrating %s module...', $module));
+            $this->migrator->run(base_path() . '/modules/' . strtolower($module) . '/database/migrations');
+
+            if ($firstUp) {
+                $className = '\\' . $module . '\Database\Seeds\DatabaseSeeder';
+                if (class_exists($className)) {
+                    $this->output->info(sprintf('Seeding %s module...', $module));
```
LukeTowers, line 136: Probably out of scope for this PR, but should we approach seeders (both module and plugin, which aren't quite the same) as needing the same level of tracking as migrations so that any that haven't been run yet will get run? With the current code, if a new module is added after the project is setup then it will never have its seeder run because it'll never be the firstUp again.

modules/system/classes/extensions/ModuleManager.php
```diff
@@ -0,0 +1,283 @@
...
+            ));
+        }
+
+        if (!$noRollback) {
+            $this->rollback($module);
+        }
+
+        if (!$preserveFiles) {
+            // Modules probably should not be removed
+            // File::deleteDirectory($module->getPath());
```
LukeTowers, line 203: When would this method be called and the files not need to be removed?

LukeTowers: Also something to think about with the uninstall() method is removal from the composer.json

modules/system/classes/extensions/PluginManager.php
```diff
@@ -120,8 +120,389 @@ protected function init(): void
...
+        }
+
+        // Clear the registration values cache
+        $this->registrationMethodCache = [];
+
+        return true;
+    }
+
+    /**
+     * @throws SystemException
```
LukeTowers: What throws the SystemException?

modules/system/classes/extensions/PluginManager.php
```diff
@@ -120,8 +120,389 @@ protected function init(): void
...
+        foreach ($this->list() as $code => $plugin) {
+            if (!($replaces = $plugin->getReplaces())) {
+                continue;
+            }
+            // @TODO: add full support for plugins replacing multiple plugins
+            if (count($replaces) > 1) {
+                throw new ApplicationException(Lang::get('system::lang.plugins.replace.multi_install_error'));
+            }
+            foreach ($replaces as $replace) {
+                $this->versionManager()->replacePlugin($plugin, $replace);
```
LukeTowers: Why is there a method for accessing the property?

modules/system/classes/extensions/PluginManager.php
```diff
@@ -1034,41 +1421,83 @@ protected function sortByDependencies(): array
         return $this->plugins = $sortedPlugins;
     }
 
-    //
-    // Management
-    //
+    public function resolveIdentifier(ExtensionSource|WinterExtension|string $extension): ?string
```
LukeTowers, line 1493: Should we be more specific in the manager resolveIdentifier() methods to only accept the specific WinterExtension implementations that we accept?

modules/system/classes/extensions/PluginVersionManager.php
```diff
@@ -17,25 +21,18 @@
  * @package winter\wn-system-module
  * @author Alexey Bobkov, Samuel Georges
  */
-class VersionManager
+class PluginVersionManager
```
LukeTowers, line 24: Is the only reason this is a separate class to avoid having the PluginManager being too long or is it ever used by anything else?

modules/system/classes/extensions/source/ExtensionSource.php
```diff
@@ -0,0 +1,354 @@
...
+            exit(0);
+        }
+
+        // Read the pipe, this is blocking and will wait for the child to close the pipe handle in it's own process
+        $pipe = fopen($pipeName, 'r');
+        // Read the namespace from the pipe
+        $namespace = fread($pipe, 4096);
+        // Close and clean up
+        fclose($pipe);
+        unlink($pipeName);
```
LukeTowers: Crazy wizard 🧙 😂 
Let's just use regex here instead to scrape the namespace from the plugin.php. Also I think the only valid Plugin.php is `Plugin.php` verbatim, so probably no need to normalize the paths?

modules/system/console/BaseScaffoldCommand.php
```diff
@@ -89,18 +90,27 @@ public function makeStubs(): void
             . DIRECTORY_SEPARATOR
             . 'lang.php'
         );
+
+        $relativeFile = str_replace(base_path(), '', $langFilePath);
```
LukeTowers, line 94: Me likey ❤️

modules/system/console/JaxTest.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Execute the console command.
+     * @throws ApplicationException
+     */
+    public function handle(): int
+    {
+
+        return 0;
+    }
+}
```
LukeTowers: Does this file still need to be here?

modules/system/controllers/updates/traits/ManagesMarketplaceProject.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace System\Controllers\Updates\Traits;
+
+use Backend\Facades\Backend;
+use Exception;
+use Illuminate\Http\RedirectResponse;
+use Illuminate\Support\Facades\Lang;
+use October\Rain\Exception\ApplicationException;
```
LukeTowers:
```suggestion
use Winter\Storm\Exception\ApplicationException;
```

modules/system/tests/README.md
```diff
@@ -59,8 +59,7 @@ php artisan plugin:refresh Acme.Blog
 > **Note:** If your plugin uses [configuration files](../plugin/settings#file-configuration), then you will need to run `System\Classes\PluginManager::instance()->registerAll(true);` in the `setUp` method of your tests. Below is an example of a base test case class that should be used if you need to test your plugin working with other plugins instead of in isolation.
 
 ```php
-use System\Classes\PluginManager;
-use System\Tests\Bootstrap\PluginTestCase;
+use System\Classes\Extensions\PluginManager;use System\Tests\Bootstrap\PluginTestCase;
```
LukeTowers:
```suggestion
use System\Classes\Extensions\PluginManager;
use System\Tests\Bootstrap\PluginTestCase;
```

modules/system/tests/bootstrap/PluginManagerTestCase.php
```diff
@@ -52,7 +51,7 @@ public function setUp() : void
         UpdateManager::forgetInstance();
 
         // Forces plugin migrations to be run again on every test
-        VersionManager::forgetInstance();
+//        PluginVersionManager::forgetInstance();
```
LukeTowers, line 54: Is this temporary?

modules/system/traits/InteractsWithZip.php
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @throws ApplicationException if the archive failed to extract
+     */
+    public function extractArchive(string $archive, string $destination): void
+    {
+        if (!Zip::extract($archive, $destination)) {
+            throw new ApplicationException(Lang::get('system::lang.zip.extract_failed', ['file' => $archive]));
+        }
+
+        @unlink($archive);
```
LukeTowers, line 22: Is it expected that the original archive is always removed?

modules/system/classes/UpdateManagerOld.php
```diff
@@ -0,0 +1,1100 @@
+<?php
+?>
```
LukeTowers: To be removed

modules/system/classes/UpdateManager.php
```diff
@@ -5,27 +5,19 @@
...
-use Illuminate\Console\View\Components\Error;
-use Illuminate\Console\View\Components\Info;
 use Illuminate\Database\Migrations\DatabaseMigrationRepository;
 use Illuminate\Database\Migrations\Migrator;
 use Illuminate\Support\Facades\App;
-use Illuminate\Support\Facades\Cache;
-use Illuminate\Support\Facades\Lang;
-use RecursiveDirectoryIterator;
-use RecursiveIteratorIterator;
+use Mix\TestA\Plugin;
```
LukeTowers: Was the wrong class imported here?

modules/system/classes/InstallManager.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace System\Classes;
+
+class InstallManager
```
LukeTowers: What's the plan for this class?

LukeTowers (changes requested):
Overall very happy with the direction of this PR, great work so far @jaxwilko! ❤️

modules/backend/ServiceProvider.php
jaxwilko: See https://github.com/wintercms/winter/pull/1259/commits/729e46e0abcffdc86cc3609314156dad408497aa & https://github.com/wintercms/storm/commit/6b03b16ff901b953db6322930aab4859b1d53235

modules/backend/ServiceProvider.php
jaxwilko: See https://github.com/wintercms/winter/pull/1259/commits/729e46e0abcffdc86cc3609314156dad408497aa & https://github.com/wintercms/storm/commit/6b03b16ff901b953db6322930aab4859b1d53235

modules/system/classes/extensions/source/ExtensionSource.php
```diff
@@ -0,0 +1,354 @@
...
+            return null;
+        }
+
+        return $this->path = $this->guessPathFromCode($this->code);
+    }
+
+    /**
+     * @throws ApplicationException
+     */
+    public function createFiles(): ?static
```
mjauvin, line 109: Why not move this logic to its dedicated ExtensionSource classes (i.e. ComposerSource, LocalSource, MaketSource)

modules/system/classes/extensions/ExtensionManager.php
jaxwilko: Singleton disallows `__constructor()` which we use to define the default output object... could work round it tho by checking if none exists and creating at use
image

modules/cms/ServiceProvider.php
jaxwilko: See https://github.com/wintercms/winter/pull/1259/commits/729e46e0abcffdc86cc3609314156dad408497aa & https://github.com/wintercms/storm/commit/6b03b16ff901b953db6322930aab4859b1d53235

modules/system/classes/extensions/ExtensionManagerInterface.php
jaxwilko: it's uninstall all, `uninstall` only accepts a single extension, i.e. `PluginManager::instance()->uninstall('Jax.Test')`. I wanted to make it a seperate method to make it veryy obvious that what's about to happen.

I also didn't want a situation like:
```php
$toUninstall = ['A.Test', null, 'B.Test'];
foreach ($toUninstall as $plugin) {
    PluginManager::instance()->uninstall($plugin);
}
```
Oh no everything just got deleted, including all installed plugins not listed.

This way, it'll just throw a plugin not found exception and if you want to remove everything you have to explicitly call `tearDown`.

modules/system/classes/extensions/ModuleManager.php
jaxwilko: see: https://github.com/wintercms/winter/pull/1259/commits/4e339308a4fbce24803adee44c9fbb84018c9146

modules/system/classes/extensions/ModuleManager.php
jaxwilko: Yes, I did leave a todo in the plugin manager for it. I may have deleted all my work but not including it so till we're ready I'm keen on leaving it in lol

modules/system/classes/extensions/ModuleManager.php
jaxwilko: Yes, it will keep a local archieve of the extension prior to update, this should be the groundwork for us to revert properly back to a previous version

modules/system/classes/extensions/ModuleManager.php
jaxwilko: Agreed, maybe we could use `Parameter` to record inits for modules? It's currently used like that for themes.

modules/system/classes/extensions/ModuleManager.php
jaxwilko: When you run `./artisan winter:down && ./artisan winter:up` and he won't come up because you just deleted system :'(

modules/system/classes/extensions/PluginManager.php
jaxwilko: I think it was something from a previous revision, see: https://github.com/wintercms/winter/pull/1259

modules/system/classes/extensions/PluginManager.php
jaxwilko: Public method, private property. Property should be used internally tho, this code was moved from `UpdateManager` and this refactor was overlooked. See: https://github.com/wintercms/winter/pull/1259/commits/c8bbe35b90f157ca5f4e832bf29396bf5d8d177e

modules/system/classes/extensions/PluginManager.php
jaxwilko: It's overriding the parent, not sure if we can limit the scope of the type hints

modules/system/classes/extensions/PluginVersionManager.php
jaxwilko: It is used by some other parts of the system but only via the plugin manager, i.e. `PluginManager::instance()->versionManager()`. It could probably be merged in but `PluginManager` is already chonk and confusing...

modules/system/classes/extensions/source/ExtensionSource.php
jaxwilko: Honestly because I only added those classes later after already implementing all of this logic in `ExtensionSource`, but yes I think it should be moved. Was just easier during development to keep them all in the same place.

modules/system/classes/extensions/source/ExtensionSource.php
jaxwilko: Not sure, we already do search for it in 2 different ways :shrug: 
https://github.com/wintercms/winter/blob/c8bbe35b90f157ca5f4e832bf29396bf5d8d177e/modules/system/classes/extensions/PluginManager.php#L883
https://github.com/wintercms/winter/blob/c8bbe35b90f157ca5f4e832bf29396bf5d8d177e/modules/system/classes/extensions/PluginManager.php#L920

jaxwilko: See: https://github.com/wintercms/winter/pull/1259/commits/8da2212fc295e98b7f4aa5e12ecb4a3c29940895

modules/system/console/JaxTest.php
jaxwilko: It's my test, will remove before merge, just commit so it stopped being annoying

modules/system/tests/bootstrap/PluginManagerTestCase.php
jaxwilko: Yes, it's a note to self that I need to look into how not flushing the instance works inside tests

modules/system/traits/InteractsWithZip.php
jaxwilko: I would expect so? Reckon it should be an optional param?

modules/system/classes/UpdateManagerOld.php
jaxwilko: See: https://github.com/wintercms/winter/pull/1259/commits/f950ce3b8e735ed574fe23e97bae5e9b7bcbb2e7

modules/system/classes/UpdateManager.php
jaxwilko: Yes, see: https://github.com/wintercms/winter/pull/1259/commits/16a4841c4c9bccd216c1da9770a34b6f4e9d6fc0

modules/system/classes/InstallManager.php
jaxwilko: See: https://github.com/wintercms/winter/pull/1259/commits/f9773db6ed248380fd1ae135395025214128ff34.

I think the original idea was to do basically what `ExtensionSource` is doing

modules/system/classes/extensions/source/ExtensionSource.php
jaxwilko: Ah, so we might not want to do this as the internal `$source` property is changed during install, i.e. after `composer require ...` is executed and we have the files, we modify the `$source` to local so we don't re-download if `createFiles()` is called again. Just did the work to split it all out and not sure if I like it :thinking:

mjauvin: Ah, makes sense to keep it in here then!

modules/system/classes/extensions/ModuleManager.php
```diff
@@ -0,0 +1,286 @@
...
+                    ));
+
+                    Preserver::instance()->store($extension);
+                    $update = Composer::update(dryRun: true, package: $composerPackage);
+
+                    $versions = $update->getUpgraded()[$composerPackage] ?? null;
+
+                    $versions
+                        ? $this->renderComponent(Info::class, sprintf('Updated module %s (%s) from v%s => v%s', $module, $composerPackage, $versions[0], $versions[1]))
+                        : $this->renderComponent(Error::class, sprintf('Failed to module %s (%s)', $module, $composerPackage));
```
mjauvin:
```suggestion
                        : $this->renderComponent(Error::class, sprintf('Failed to update module %s (%s)', $module, $composerPackage));
```

modules/system/classes/extensions/PluginManager.php
```diff
@@ -1034,41 +1430,83 @@ protected function sortByDependencies(): array
...
+    public function deletePlugin(string $id): void
+    {
+        $this->uninstall($id);
+    }
+
+    /**
+     * Disables the provided plugin using the provided flag (defaults to static::DISABLED_BY_USER)
+     * @deprecated
+     */
+    public function disablePlugin(PluginBase|string $plugin, string|bool $flag = self::DISABLED_BY_USER): bool
```
mjauvin:
```suggestion
    public function disablePlugin(PluginBase|string $plugin, string|bool $flag = self::DISABLED_BY_USER): ?bool
```

modules/system/classes/extensions/PluginManager.php
```diff
@@ -1034,41 +1430,83 @@ protected function sortByDependencies(): array
...
+    {
+        return $this->disable($plugin, $flag);
+    }
+
+    /**
+     * Enables the provided plugin using the provided flag (defaults to static::DISABLED_BY_USER)
+     * @deprecated
+     */
+    public function enablePlugin(PluginBase|string $plugin, $flag = self::DISABLED_BY_USER): bool
```
mjauvin:
```suggestion
    public function enablePlugin(PluginBase|string $plugin, $flag = self::DISABLED_BY_USER): ?bool
```

mjauvin (commented):
called method might return null, we need to use `?bool` return type hint

mjauvin:
@jaxwilko @LukeTowers should we add localization strings afterward or in this PR for info/warning/error messages ?

mjauvin:
@jaxwilko `requestUpdateList()` does not exist anymore in UpdateManager class but is still being called in `check()` method.

mjauvin:
Also getting this error when clicking "install themes" from the `Updates & Plugins` page:
image

jaxwilko:
> @jaxwilko @LukeTowers should we add localization strings afterward or in this PR for info/warning/error messages ?

100%

All the strings in output atm are just placeholders

jaxwilko:
> @jaxwilko `requestUpdateList()` does not exist anymore in UpdateManager class but is still being called in `check()` method.

Yeah, I've not finished fully migrating the backend yet, check for updates should work (but only supports composer packages) and interacting with plugins should work

jaxwilko:
> Also getting this error when clicking "install themes" from the `Updates & Plugins` page:
> image
> 

Luke broke it, it's from his PR this one is based off... I'll look into it when i do more on the backend stuff :)

modules/system/classes/extensions/ExtensionManager.php
LukeTowers: Sounds good, let's give that a try then.

modules/system/classes/extensions/ModuleManager.php
LukeTowers: Works for me

modules/system/classes/extensions/ModuleManager.php
LukeTowers: Right, but shouldn't those commands be passing the preserve files flag?

modules/system/traits/InteractsWithZip.php
LukeTowers: Yes probably. If we're using the preserver then we wouldn't want to delete the copy after restoring it I would think

jaxwilko:
@bennothommo plz need sexy help!
image

modules/system/controllers/updates/assets/src/components/Product.vue
```diff
@@ -0,0 +1,151 @@
...
+}
+.product-footer .packagist .product-badge {
+    background: #f28d1a;
+}
+.product-footer .stars, .product-footer .github {
+    margin-right: 7px;
+}
+</style>
+
+
```
LukeTowers:
```suggestion
```

modules/system/controllers/updates/_install_plugins.php
```diff
@@ -1,119 +1,8 @@
...
-            href="javascript:;">
-            <div class="image"><img src="{{image}}" alt=""></div>
-            <div class="details">
-                <h5 class="text-overflow">{{code}}</h5>
-                <p>{{description}}</p>
-            </div>
-        </a>
-    </div>
-</script>
+
```
LukeTowers:
```suggestion
```

modules/system/controllers/updates/assets/src/components/Product.vue
```diff
@@ -0,0 +1,149 @@
...
+        <div class="product-footer">
+            <div class="product-footer-item">
+                <div title="Stars given" class="stars">
+                    <span class="product-badge"><i class="icon-star"></i></span>
+                    {{product.favers}}
+                </div>
+                <div title="Downloads" class="downloads">
+                    <span class="product-badge"><i class="icon-download"></i></span>
+                    {{product.downloads}}
+                </div>
```
LukeTowers, line 67: We should probably add the license as well

jaxwilko: Sure, it'll have to come from the marketplace tho as the packagist api does not provide it

LukeTowers:
For future reference, @jaxwilko's JS for the event stream handling

```js
const response = await fetch(window.location.href, {
                method: 'POST',
                headers: {
                    'Accept': 'text/event-stream',
                    'X-Requested-With': 'XMLHttpRequest',
                    'X-WINTER-REQUEST-HANDLER': 'onInstallPlugin',
                    'X-WINTER-REQUEST-PARTIALS': '',
                    'X-CSRF-TOKEN': document.querySelector('meta[name="csrf-token"]').getAttribute('content')
                },
                body: data,
                redirect: 'follow',
                mode: 'same-origin',
            });

            if (response.status !== 200) {
                console.log('BAD');
                this.installing = false;
                return;
            }

            $.popup({
                size: 'installer-popup',
                content: `
                    <div class="modal-header">
                        <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
                        <h4 class="modal-title">Installing ${this.product.name}</h4>
                    </div>
                    <div class="modal-body"></div>
                    <div class="modal-footer">
                        <button type="button" class="btn btn-default" data-dismiss="modal">Blue Pill</button>
                        <button type="button" class="btn btn-primary" data-dismiss="modal">Red Pill</button>
                    </div>
                `
            });

            const popup = document.querySelector('.size-installer-popup');

            const chunk = async (response) => {
                const reader = response.body.getReader();
                const decoder = new TextDecoder();

                const parse = (str) => str.split("\n").filter((line) => !!line).map((line) => ({
                    type: line.substring(0, line.indexOf(': ')),
                    value: line.substring(line.indexOf(': ') + 2)
                }));

                const prepareMessage = (str) => {
                    ['INFO', 'ERROR'].forEach((status) => {
                        if (str.indexOf(status) === 0) {
                            str = `<span class="message-${status.toLowerCase()}">${status}</span> <pre>${str.substring(status.length + 1)}</pre>`;
                        }
                    });
                    return `<div class="install-message">${str}</div>`;
                };

                let done, value, message;
                while (!done) {
                    ({value, done} = await reader.read());
                    if (done) {
                        return;
                    }

                    message = parse(decoder.decode(value));

                    if (message[0].value === 'message') {
                        popup.querySelector('.modal-body').innerHTML += prepareMessage(message[1].value);
                    }

                    if (message[0].value === 'error') {
                        alert(message[1].value);
                    }
                }
            };

            await chunk(response);
            this.installing = false;
```

modules/system/controllers/updates/assets/src/components/Product.vue
```diff
@@ -146,4 +215,37 @@ export default {
 .product-footer .stars, .product-footer .github {
     margin-right: 7px;
 }
+.installing:after {
+    content: ' ';
+    display: block;
+    background-size: 50px 50px;
+    background-repeat: no-repeat;
+    background-position: 50% 50%;
+    background-image: url(/modules/system/assets/ui/images/loader-transparent.svg);
```
LukeTowers: Any way for this to be loaded via the ASSET_URL?

jaxwilko: yes there is... but this is all placeholder :innocent:

jaxwilko:
Just pushed a big re-work of the backend UI. This push include [package-list.json](https://github.com/wintercms/winter/blob/ff8a4d7afdf1e42b8cffa230069097218f643129/modules/system/classes/core/package-list.json) which is a sample of demo data that will eventually be provided by the Winter Marketplace API.

As the sample is include and I've commit compiled assets for the backend, people should be able to test if they want.

> This is highly experimental! I've not had a chance to test the install functionality recently and as last time I did I managed to delete a load of work I would recommend testing in a fresh copy of Winter.

## Some Screenshots:
### Desktop
image
image
image
image
image

### Tablet
image

### Mobile
image

jaxwilko:
Pushed up some fixes for the update popup, need to test it actually works but it now loads which is a good first step.

Also added an update icon for products in the marketplace:
image
image

austinkregel:
@jaxwilko This looks really nice 🔥 

but it does appear that we're missing at least one interface 🤔

jaxwilko:
Hey @austinkregel, which interface are you referring to?

jaxwilko:
Plugin Management page after some tweaks:
image

composer.json
```diff
@@ -81,7 +81,8 @@
     "config": {
         "allow-plugins": {
             "composer/installers": true,
-            "wikimedia/composer-merge-plugin": true
+            "wikimedia/composer-merge-plugin": true,
+            "php-http/discovery": true
```
LukeTowers:
```suggestion
            "php-http/discovery": false
```

composer.json
```diff
@@ -30,7 +30,7 @@
     },
     "require": {
         "php": ">=8.1",
-        "winter/storm": "dev-develop as 1.2",
+        "winter/storm": "dev-wip/support-uploading-packages-update-manager-changes as 1.2",
```
LukeTowers, line 33: This will need to be updated before merge.

LukeTowers (changes requested):

LukeTowers:
@coderabbitai review

PR #1260, opened 2024-12-02
Bump vite and laravel vite plugin versions

jaxwilko:
Bumps versions, requires testing.


jaxwilko:
Tested on a few projects, seems to work fine :)

PR #1263, opened 2024-12-04
Add model:prune console command

mjauvin:

### DIFF
--- a/modules/system/console/PruneCommand.php
+++ b/modules/system/console/PruneCommand.php
@@ -63,7 +63,7 @@ class PruneCommand extends BasePruneCommand
          * Example usage:
          *
          *     Event::listen('system.console.model.prune.findModels', function () {
-         *         return collect(['example model' => '\System\Models\File']);
+         *         return ['example model' => '\System\Models\File'];
          *     });
          *
          */
--- a/modules/system/console/PruneCommand.php
+++ b/modules/system/console/PruneCommand.php
@@ -31,9 +31,7 @@ class PruneCommand extends BasePruneCommand
                     return in_array($model, $except);
                 });
             })->filter(function ($model) {
-                return $this->isPrunable($model);
-            })->filter(function ($model) {
-                return class_exists($model);
+                return class_exists($model) && $this->isPrunable($model);
             })->values();
     }
 
--- a/modules/system/console/PruneCommand.php
+++ b/modules/system/console/PruneCommand.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace System\Console;
+
+use Exception;
+use Illuminate\Database\Console\PruneCommand as BasePruneCommand;
+use Illuminate\Database\Eloquent\MassPrunable;
+use Illuminate\Database\Eloquent\Prunable;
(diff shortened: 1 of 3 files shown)


modules/system/console/PruneCommand.php
```diff
@@ -0,0 +1,77 @@
...
+
+        return collect($this->findModels())
+            ->when(! empty($except), function ($models) use ($except) {
+                return $models->reject(function ($model) use ($except) {
+                    return in_array($model, $except);
+                });
+            })->filter(function ($model) {
+                return $this->isPrunable($model);
+            })->filter(function ($model) {
+                return class_exists($model);
```
LukeTowers: Should we combine these?

modules/system/helpers/ModelFinder.php
```diff
@@ -0,0 +1,57 @@
...
+    public static function findModels(): array
+    {
+        $models = [];
+        $models[] = static::findCoreModels();
+        $models[] = static::findActivePluginsModels();
+
+        return collect($models)->flatten()->all();
+    }
+
+    public static function findCoreModels(): array
```
LukeTowers:
```suggestion
    public static function findModuleModels(): array
```

modules/system/helpers/ModelFinder.php
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @return Collection
+     */
+    public static function findModels(): array
+    {
+        $models = [];
+        $models[] = static::findCoreModels();
+        $models[] = static::findActivePluginsModels();
+
+        return collect($models)->flatten()->all();
```
LukeTowers, line 22: Are we doing any sort of checking to make sure the classes are actual model classes?

modules/system/console/PruneCommand.php
mjauvin: Good call!
Done.

modules/system/helpers/ModelFinder.php
mjauvin: Not really, no.

mjauvin: @LukeTowers But I do search only under `/models/` folders

mjauvin: > Are we doing any sort of checking to make sure the classes are actual model classes?

Any suggestions ?

bennothommo: @mjauvin if the intention is to find out if it's a model *without* loading it, you'll have to use something like [PHP Parser](https://github.com/nikic/php-parser) to statically analyze the file.

mjauvin: @bennothommo I think searching for all class files under `models` is good enough for the purpose, more logic can be added by the code using the ModelFinder helper once it gets the list of class files as this is more efficient.

LukeTowers:
@jaxwilko random thought: 

This could be related to our work on the extension management refactoring where we could make it so that the extension managers could have an API for searching for specific types of classes / files under them (i.e. model classes, components, CLI commands, etc) to make these sorts of use cases easier to implement.

PR #1266, opened 2024-12-04
Added fix to ensure escaped backslash paths are preserved in DB_DATABASE env values

jaxwilko:
This fix  adds support for preserving double escapped backslashes when running `winter:env` on Windows when copying the `DB_DATABASE` env value, this fixes an issue caused by `winter:install` creating the sqlite db database path with a Windows directory seperator.

### DIFF
--- a/modules/system/console/WinterEnv.php
+++ b/modules/system/console/WinterEnv.php
@@ -122,7 +122,11 @@ class WinterEnv extends Command
                     $dbConfig = $this->dbConfig()[$default] ?? [];
 
                     foreach ($dbConfig as $dbEnvKey => $dbConfigKey) {
-                        $env->set($dbEnvKey, config(join('.', [$config, 'connections', $default, $dbConfigKey])));
+                        $value = config(join('.', [$config, 'connections', $default, $dbConfigKey]));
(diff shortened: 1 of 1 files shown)


modules/system/console/WinterEnv.php
```diff
@@ -122,7 +122,11 @@ protected function updateEnvFile(): void
                     $dbConfig = $this->dbConfig()[$default] ?? [];
 
                     foreach ($dbConfig as $dbEnvKey => $dbConfigKey) {
-                        $env->set($dbEnvKey, config(join('.', [$config, 'connections', $default, $dbConfigKey])));
+                        $value = config(join('.', [$config, 'connections', $default, $dbConfigKey]));
+                        if ($dbEnvKey === 'DB_DATABASE' && PHP_OS_FAMILY === 'Windows' && str_contains($value, '\\')) {
```
LukeTowers, line 127:
```suggestion
                        // Fix for https://github.com/wintercms/winter/issues/1242#issuecomment-2515344385
                        if ($dbEnvKey === 'DB_DATABASE' && PHP_OS_FAMILY === 'Windows' && str_contains($value, '\\')) {
```

PR #1267, opened 2024-12-05
Fix TagList formwidget

### LINKED ISSUE #1223: TagList not saving an empty value

mjauvin:
The TagList formwidget was not working properly in relation mode if the relation was not using a pivot table (e.g. morphMany/hasMany)

All modes did not update the model field or relation when all tags had been cleared.

Fixes #1223

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1227,6 +1227,10 @@ class Form extends WidgetBase
             // Exclude fields that didn't provide any value
             $fieldValue = $this->dataArrayGet($result, $parts, FormField::NO_SAVE_DATA);
             if ($fieldValue === FormField::NO_SAVE_DATA) {
+                if ($widget->getConfig('type') === 'taglist' && $widget->getConfig('mode') === 'relation') {
+                    // let the TagList remove its relation items
+                    $widget->getSaveValue(array());
+                }
                 continue;
             }
 
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1227,8 +1227,8 @@ class Form extends WidgetBase
             // Exclude fields that didn't provide any value
             $fieldValue = $this->dataArrayGet($result, $parts, FormField::NO_SAVE_DATA);
             if ($fieldValue === FormField::NO_SAVE_DATA) {
-                if ($widget->getConfig('type') === 'taglist' && $widget->getConfig('mode') === 'relation') {
-                    $fieldValue = [];
+                if ($widget->getConfig('type') === 'taglist') {
+                    $fieldValue = in_array($widget->getConfig('mode'), ['array','relation']) ? [] : '';
                 } else {
                     continue;
                 }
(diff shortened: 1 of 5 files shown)


mjauvin:
@bennothommo I just saw you started a PR on TagList as well, thought you might want to review this one.

mjauvin:
This PR fixes the TagList in relation mode for MorphMany relations, but looks like it breaks it for BelongsToMany relations.

bennothommo:
@mjauvin it's probably worth adding an example in the Test plugin if there isn't one already, if adding the unit tests becomes too difficult.

mjauvin:
yeah, that's what I'm going to do for now.

mjauvin:
tested working with the following modes using Winter.Test plugin in Post & Record models:

- relation (belongsToMany, hasMany, morphMany)
- array (useKey = true/false)
- string (useKey = true/false)

All the above tests also include clearing all values and making sure the model field or relation was updated accordingly.

modules/backend/widgets/Form.php
```diff
@@ -1227,7 +1227,11 @@ public function getSaveData(): array
             // Exclude fields that didn't provide any value
             $fieldValue = $this->dataArrayGet($result, $parts, FormField::NO_SAVE_DATA);
             if ($fieldValue === FormField::NO_SAVE_DATA) {
-                continue;
+                if ($widget->getConfig('type') === 'taglist') {
+                    $fieldValue = in_array($widget->getConfig('mode'), ['array','relation']) ? [] : '';
+                } else {
+                    continue;
+                }
```
mjauvin: The above code snippet should not be required if we can fix the TagList widget to include empty array or empty string in the POST data.

@bennothommo any help would be appreciated.

mjauvin: fixed.

mjauvin:
@LukeTowers Anything more before merging this?

LukeTowers:
@mjauvin @bennothommo Has it been fully tested?

mjauvin:
> @mjauvin @bennothommo Has it been fully tested?

Yes, see previous comment above.

AIC-BV:
Can confirm this fixes the problem
Thanks!

AIC-BV:
image
Oh it seems to generate 2 tags instead of one when creating a tag for the first time

When removing all tags, they also save as an empty string
image

When adding tags
image

mjauvin:
@AIC-BV can you show your tag field definition and relation?

AIC-BV:
@mjauvin 
There is no relation

```
tags:
    label: Tags
    type: taglist
    customTags: false
    options:
        nieuw: Nieuw
        best: Beste deal
    span: auto
```

LukeTowers:
@AIC-BV can you open an issue for that so that we don't release 1.2.8 until it's fixed?

AIC-BV:
https://github.com/wintercms/winter/issues/1275

PR #1276, opened 2024-12-20
Remove empty item for all taglist modes

### LINKED ISSUE #1275: Taglist is saving empty strings
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

image
It seems to generate 2 tags instead of one when creating a tag for the first time

When removing all tags, they also save as an empty string
image

When adding tags there is an empty string before the actual first tag
image

### Steps to replicate

```
tags:
    label: Tags
    type: taglist
    customTags: false
    options:
        nieuw: Nieuw
        best: Beste deal
    span: auto
```

### Workaround

_No response_

mjauvin:
The empty item that was added in previous PR was not removed in array and string modes.
Make sure we always deal with arrays from the POST data.

Fixes #1275

### DIFF
--- a/modules/backend/formwidgets/TagList.php
+++ b/modules/backend/formwidgets/TagList.php
@@ -110,15 +110,17 @@ class TagList extends FormWidgetBase
      */
     public function getSaveValue($value)
     {
-        if (is_array($value)) {
-            $value = array_filter($value);
+        if (!is_array($value)) {
+            $value = [$value];
         }
+
+        $value = array_filter($value);
 
         if ($this->mode === static::MODE_RELATION) {
             return $this->hydrateRelationSaveValue($value);
         }
 
-        if (is_array($value) && $this->mode === static::MODE_STRING) {
+        if ($this->mode === static::MODE_STRING) {
             return implode($this->getSeparatorCharacter(), $value);
         }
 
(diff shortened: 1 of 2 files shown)


mjauvin:
@AIC-BV can you check if this resolves the issues you were having ?

modules/backend/formwidgets/TagList.php
```diff
@@ -110,6 +110,10 @@ public function prepareVars()
      */
     public function getSaveValue($value)
     {
+        if (is_array($value)) {
+            $value = array_filter($value);
+        }
```
LukeTowers, line 115: Doesn't this change now allow for the case where an empty value is passed `""` when mode is MODE_RELATION and it doesn't get filtered out in either location?

mjauvin: No, this widget always returns an array in the POST data since its markup uses the array notation `[]` after the fieldName.

I should probably remove all check for `is_array()`

mjauvin: @LukeTowers I've streamlined the getSaveData() method.

LukeTowers: These changes, along with the array naming of the singular input now means that a field that previously stored a string value will attempt to store an array value instead which will break things.

mjauvin: No, the field won't STORE an array, it will still store a string... it's just the Form itself that returns an array in the POST data, that didn't change.

mjauvin: You can easily test all that with Winter.Test plugin under the Posts controller... this has been tested quite a bit already.

LukeTowers: Ah, I see now.

bennothommo (approved):

PR #1285, opened 2025-01-13
Table widget improvements

bennothommo:
- Style read-only columns with slightly darker grey background to indicate read-only state
- Prevent editors from being created for all column types if read-only (previously, only string columns would be rendered read-only by setting the `readonly` attribute, this is not ideal because it can be easily changed)
- Allow tab and arrow navigation for read-only columns
- Make `readOnly` option case-insensitive
- Allowed `searching` option to show search box even if `adding` and `deleting` buttons are disabled
- Fixed styling of toolbar if only the search box is shown (no background previously)
- Fixed styling of pagination
- Fixed broken search if client datasource is used

### In progress
- ~~Add support for `inspector` column type for data tables - allowing support for complex values for individual table records (ie. an object).~~ (will resubmit in another PR)


bennothommo:
Before

Screenshot From 2025-01-13 12-42-54

After

Screenshot From 2025-01-13 12-42-37

LukeTowers:
@bennothommo is this ready for review?

bennothommo:
@LukeTowers everything except for the Inspector column type. I could strip that out and do that as another PR.

LukeTowers:
> @LukeTowers everything except for the Inspector column type. I could strip that out and do that as another PR.

@bennothommo let's do that then. Would like to release 1.2.8 in the next week.

modules/backend/widgets/table/assets/js/table.datasource.client.js
```diff
@@ -86,6 +86,42 @@
         this.getRecords(offset, count, onSuccess)
     }
 
+        /*
```
LukeTowers:
```suggestion
    /*
```

LukeTowers (approved):

PR #1293, opened 2025-01-22
Add allowList functionality to Twig security policy

LukeTowers:

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -309,15 +309,7 @@ class Controller
         /*
          * The 'this' variable is reserved for default variables.
          */
-        $this->getTwig()->addGlobal('this', [
-            'page'        => $this->page,
-            'layout'      => $this->layout,
-            'theme'       => $this->theme,
-            'param'       => $this->router->getParameters(),
-            'controller'  => $this,
-            'environment' => App::environment(),
-            'session'     => App::make('session'),
-        ]);
+        $this->getTwig()->addGlobal('this', $this->getControllerGlobalVars());
 
         /*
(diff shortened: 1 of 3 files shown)


bennothommo:
@LukeTowers the docs state that people can put data to the session, as well forgetting data and flushing the entire session, as per https://wintercms.com/docs/v1.2/markup/this/session.

TBH, I'd be more included to remove `this.session` entirely, or [replacing the `App::make('session')` call here](https://github.com/wintercms/winter/blob/develop/modules/cms/classes/Controller.php#L319) with a limited proxy class that doesn't expose the Session class directly.

LukeTowers:
@bennothommo in that case then I'll add those methods in but leave the rest blocked. I'm not opposed to a potential proxy class but I'm not sure how much better that'll be than the allowList approach

bennothommo:
@LukeTowers I feel the Proxy class might be "cleaner", rather than having methods allowed several levels deep in a policy where it could be perhaps forgotten. Ideally I'd like a proxy class for the controller too.

LukeTowers:
eh, possibly; my main argument against a proxy class is what else is going to use said proxy class? If it's a proxy just for the sake of locking down that interface and is only used in that one case but requires a whole separate file & class for the functionality then I'd argue that it'd be more of a pain to maintain that vs maintaining the centralized security policy for Twig as a whole.

bennothommo:
@LukeTowers we could always use an anonymous class ;)

```php
$this->getTwig()->addGlobal('this', [
    // ...
    'session' => new class {
        public function get($key)
        {
            // ...
        }

        public function has($key)
        {
            // ...
        }
    }
    // ...
]);
```

jaxwilko (approved):
LGTM

PR #1294, opened 2025-01-23
Fix TagList returning an object in array mode

thienvu18:
After applying the `array_filter` method, the `$value` array lost its 0-based indexing, causing it to be treated as an object instead of an array. To resolve this, I’ve wrapped the result with `array_values` to reindex it and restore it as an array.

### DIFF
--- a/modules/backend/formwidgets/TagList.php
+++ b/modules/backend/formwidgets/TagList.php
@@ -114,7 +114,7 @@ class TagList extends FormWidgetBase
             $value = [$value];
         }
 
-        $value = array_filter($value);
+        $value = array_values(array_filter($value));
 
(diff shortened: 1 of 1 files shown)


mjauvin:
@thienvu18 the fix is valid.

I'm curious, did you get problems using it the way it was? If so, how?

thienvu18:
@mjauvin thank for accepting my pr.

When I use this form widget for a json column in Postgres, the value is stored as `{"1":"some_value"}` instead of `["some_value"]`. This makes `whereJsonLength` throw `Invalid parameter value: 7 ERROR:  cannot get array length of a non-array`

LukeTowers (approved):

PR #1295, opened 2025-01-23
Only make recordfinder clickable if not disabled

### LINKED ISSUE #1280: Record finder is clickable even when disabled
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

Hi,

As I said in the title, record finder is not working properly when in disabled state. It's clickable even when disabled and the popup is shown.

Thanks.

### Steps to replicate

1. Add a record finder field.
2. Disable it.
3. Click on it.

### Workaround

_No response_

mjauvin:
Fixes #1280


LukeTowers:
@mjauvin does this actually solve it? Shouldn't there be a change to the JS listener attachment logic as well?

mjauvin:
> @mjauvin does this actually solve it? Shouldn't there be a change to the JS listener attachment logic as well?

Yes, the data-api ajax request attributes are only added if the field is enabled.

LukeTowers:
Ahh, I see now; missed that the first glance

PR #1304, opened 2025-02-10
Add support for React

LukeTowers:
Adds:
- viteReactRefresh() Twig & \System\Classes\Asset\Vite::reactRefreshTag() functions for support hot reloading with React & Vite
- React as an asset bundle that can be passed to `vite:create` or `mix:create`

Still to add:
- Support for React in the theme scaffolding


LukeTowers:
@bennothommo do you know why the JS tests are failing?

PR #1311, opened 2025-02-16
Automatically set BackendMenu context from controller path

### LINKED ISSUE #1269: Automatically set BackendMenu context from controller path
It would be nice if the base `Backend\Classes\Controller` class set `BackendMenu::setContext($author.$plugin, $plugin, $controller)` in the constructor so that backend controller classes didn't have to explicitly call set context if their navigation followed the standard. 

Also adjust the scaffolding to reflect that.

sephyld:
Closes #1269 

I've added the three protected attributes `$author`, `$plugin` and `$sideMenuItemCode` in order to set the backend menu context in the base controller class. As per linked issue's description, I'm setting the values for these attributes during class generation.

The only reason left for the constructor to be overridden by the created controller was the chance of using the `--sidebar` option when using the `create:controller` artisan command. Passing this option resulted in setting the boolean `sidebar` property for it to be used in the template scaffold file. If this property is `true` the controller assigns the value `compact-container` to the `$bodyClass` attribute.

In order to get rid of it I thought it would be better to set the same value for the `$bodyClass` outside the constructor

### DIFF
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -8,6 +8,7 @@ use Flash;
 use Config;
 use Request;
 use Backend;
+use BackendMenu;
 use Redirect;
 use Response;
 use Exception;
@@ -116,6 +117,21 @@ class Controller extends ControllerBase
     protected $guarded = [];
 
     /**
+     * @var string Plugin author's name specified by the extending controller class in order to properly set the backend context
+     */
+    protected $author;
+
+    /**
+     * @var string Plugin's name specified by the extending controller class in order to properly set the backend context
+     */
+    protected $plugin;
+
+    /**
+     * @var ?string Side menu item code specified by the extending controller class in order to properly set the backend context
+     */
+    protected $sideMenuItemCode = null;
+
+    /**
      * Constructor.
      */
     public function __construct()
@@ -159,6 +175,8 @@ class Controller extends ControllerBase
         }
 
         $this->extendableConstruct();
+
+        BackendMenu::setContext("$this->author.$this->plugin", strtolower($this->plugin), $this->sideMenuItemCode);
     }
 
     public function __get($name)
--- a/modules/backend/console/scaffold/controller/controller.stub
+++ b/modules/backend/console/scaffold/controller/controller.stub
@@ -1,6 +1,5 @@
 <?php namespace {{ plugin_namespace }}\Controllers;
 
-use BackendMenu;
 use Backend\Classes\Controller;
 
 /**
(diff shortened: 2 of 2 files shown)


LukeTowers:
@sephyld we could probably set those values from the namespace of `get_class($this)` inside of the run method after the action has been run but before the view is rendered if `BackendMenu::getContext()` doesn't return anything (or just set it by default in the run method before the action is run so that it can be overridden). That way devs don't have to provide any values at all if they have the default approach of 'author.plugin.controller.action` for menu context but can still override it if for some reason they need a different backend context (usually because they're providing a controller that is going to be injected into another context's navigation menu, i.e. a Comments plugin that extends the Blog plugin and adds the Comments controller to the blog main menu item).

sephyld:
@LukeTowers hi

I see. I was thinking about something like this, please let me know if I'm on the right track

```php
        // run method
        // [...]
        /*
         * Set the admin preference locale
         */
        BackendPreference::setAppLocale();
        BackendPreference::setAppFallbackLocale();

        /*
         * Set the navigation context
         */
        $this->setNavigationContext();

        /*
         * Execute AJAX event
         */
        if ($ajaxResponse = $this->execAjaxHandlers()) {
            $result = $ajaxResponse;
        }
        // [...]
```

```php
    protected function setNavigationContext(): void
    {
        $context = \BackendMenu::getContext();
        $fullClassNameArray = explode('\\', get_class($this));

        $author = $fullClassNameArray[0];
        $plugin = $fullClassNameArray[1];
        $className = $fullClassNameArray[count($fullClassNameArray) - 1];

        $owner = $context->owner ?? "$author.$plugin";
        $mainMenuCode = $context->mainMenuCode ?? strtolower($plugin);
        $sideMenuCode = $context->sideMenuCode ?? strtolower($className);

        \BackendMenu::setContext($owner, $mainMenuCode, $sideMenuCode);
    }
```

This way we cuold get rid of the `$author`, `$plugin` and `$sideMenuItemCode` protected attributes. Devs could simply override the `setNavigationContext` method doing something like this in the controller classes

```php
    protected function setNavigationContext(): void
    {
         \BackendMenu::setContext("AnotherAuthor.AnotherPlugin", "anotherplugin", "items");
    }
```

or if `\BackendMenu::setContext` is called in the constructor the getContext call will return the previously given values.

Does it check out?

LukeTowers:
That's perfect @sephyld, thanks very much! I've implemented it in https://github.com/wintercms/winter/commit/5d0b9cabc1ae10ac7360833c9beb5d550f3babc7 as a part of #1312

PR #970, opened 2023-09-08
Calendar widget + CalendarController behavior (FullCalendar v6, Snowboard)

LukeTowers:
Need to rebuild on Snowboard / (maybe Vue) with FullCalendar v6.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable backend calendar with month, week, and day views.
  * Supports event search, filtering, tooltips, colors, time zones, recurring events, and all-day events.
  * Added event and date click actions for viewing records or creating events.
  * Added calendar controls, loading indicators, and cached event loading for faster navigation.
  * Supports read-only and preview calendar displays.
* **Documentation**
  * Added configuration and customization examples.
* **Tests**
  * Added coverage for display, filtering, recurrence, time zones, caching, and event formatting.

PR #1313, opened 2025-02-20
Add backend calendar controller

damsfx:
Original PR #970 

**FullCalendar** [docs](https://fullcalendar.io/docs)
- [x] update to fullcalendar v6.1.15
- [x] add all available locales
- [ ] make use of package.json 

> Notes : 
> - `firstDay` can be set by fullcalendar according to his configured locale [see docs](https://fullcalendar.io/docs/firstDay) but it doesn't seems to work in our case

**CalendarController**
- Config : 
    - [x] add widget theme color for buttons (default, primary or secondary)
    - [x] add widget initial view
    - [x] add widget first day of week
    - [x] add model attribute for all day event


modules/backend/widgets/Calendar.php
```diff
@@ -84,6 +89,16 @@ class Calendar extends WidgetBase
      */
     public array $availableDisplayModes = [];
 
+    /**
+     * Initial calendar view,  either mont, week, day or list
```
LukeTowers:
```suggestion
     * Initial calendar view, one of month, week, day, list
```

modules/backend/widgets/Calendar.php
```diff
@@ -153,7 +183,13 @@ public function init()
...
         $this->searchColumns = $columns;
 
-        $this->calendarVisibleColumns = [$this->recordTitle, $this->recordStart, $this->recordEnd];
+        $this->calendarVisibleColumns = [
+            $this->recordTitle,
+            $this->recordStart,
+            $this->recordEnd,
+        ];
+
+        // $this->validateModel();
```
LukeTowers, line 198: Why is this commented out?

modules/backend/widgets/calendar/assets/less/calendar.less
```diff
@@ -1,16 +1,30 @@
...
+    padding: 0 20px 1px;
     font-size: 15px;
 
     .loading-indicator{
         background-color: rgba(0, 0, 0, 0.02);
     }
 
     .calendar-control{
-        max-width: 1000px;
+        // max-width: 1000px;
```
LukeTowers: Commented out?

modules/backend/widgets/calendar/assets/less/calendar.less
```diff
@@ -1,16 +1,30 @@
+
+
+// Importing file while using 
+// @import "../../../../../system/assets/ui/less/button.variables.less";
```
LukeTowers: Commented out?

modules/backend/widgets/calendar/assets/less/calendar.less
```diff
@@ -1,16 +1,30 @@
+
+
```
LukeTowers:
```suggestion
```

LukeTowers:
@damsfx can you take a stab at pulling and compiling the assets with vite so that they're pulled in with package.json rather than including the vendor assets in the code directly? We're trying to move away from including the raw vendor assets as the source of truth (although we'd still be committing the compiled results and including those), it's mostly just to make it easier to update the dependencies in the future. Let me know if you need any help with that.

modules/backend/widgets/Calendar.php
damsfx: I tried to add it based on the ListController but I get an error.  
Need to take a closer look.

modules/backend/widgets/calendar/assets/less/calendar.less
damsfx: Yes, this way the widget can expend to the max width available.

modules/backend/widgets/calendar/assets/less/calendar.less
damsfx: Yes, my intention was to get the brand colors of Winter, but I think the assets needs to be compile instead of directly use the less file.

modules/backend/widgets/Calendar.php
damsfx: @LukeTowers If I try to validate the model (as in Lists widget), then this error in throw :   
`Typed property Backend\Widgets\Calendar::$model must not be accessed before initialization`.

Same if I try to add `model` to the `fillFromConfig()`method in plugin initialization.

damsfx:
I wouldn't mind a little help on this one.
I don't have much experience with Vite, I'm just starting to integrate it into my own plugins/themes.

modules/backend/widgets/calendar/assets/less/calendar.less
damsfx: The calendar css is now compiled with `winter:util compile less` command.  
The winter's button colors are used to style calendar widget buttons.

LukeTowers:
Going to merge this into #970 as @jaxwilko will have some time to work on this ❤️

PR #1314, opened 2025-02-21
Allow "winter:util compile" command to download Font Awesome if required

bennothommo:
For convenience, or if a user is unable to use NPM and/or `php artisan mix:install`, this will allow the `php artisan winter:util compile less` command to download Font Awesome if it does not exist, and place the necessary files in location in order to allow LESS compilation to work.

### DIFF
--- a/modules/system/console/WinterUtil.php
+++ b/modules/system/console/WinterUtil.php
@@ -3,7 +3,9 @@
 use Lang;
 use File;
 use Config;
+use DirectoryIterator;
 use Illuminate\Console\Command;
+use Illuminate\Support\Facades\File as Filesystem;
 use Illuminate\Support\Facades\Storage;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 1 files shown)


LukeTowers:
@bennothommo can we change the packages that are requiring that file to be compiled with mix or vite instead of adding this complexity for the old asset compilation system?

bennothommo:
@LukeTowers we already use Mix to compile the icons normally. This is just a workaround for people who can't use Mix/Vite to still be able to compile assets the old way if they need. It was to basically circumvent this issue:

image

mjauvin:
@LukeTowers this would greatly help me if it got merged.

helmutkaufmann:
Will this get committed?

LukeTowers:
@helmutkaufmann I'm not sure yet, it's a lot of extra code to include something that can already be accomplished by just running `artisan mix:install` and including the `module-system` package when asked.

mjauvin:
@LukeTowers we should be able to compile those assets without mix. I don't use mix myself (don't want to install npx)

modules/system/console/WinterUtil.php
```diff
@@ -114,6 +119,7 @@ protected function getOptions()
             ['debug', null, InputOption::VALUE_NONE, 'Run the operation in debug / development mode.'],
             ['projectId', null, InputOption::VALUE_REQUIRED, 'Specify a projectId for set project'],
             ['missing-files', null, InputOption::VALUE_NONE, 'Purge system_files records for missing storage files'],
+            ['include-icons', null, InputOption::VALUE_NONE, 'Include icons.less when compiling LESS files'],
```
LukeTowers:
```suggestion
```

PR #1315, opened 2025-02-23
Add failover transport to MailSetting

mjauvin:
image

### DIFF
--- a/modules/system/models/MailSetting.php
+++ b/modules/system/models/MailSetting.php
@@ -13,6 +13,7 @@ class MailSetting extends Model
 {
     use \Winter\Storm\Database\Traits\Validation;
 
+    const MODE_FAILOVER  = 'failover';
     const MODE_LOG       = 'log';
     const MODE_MAIL      = 'mail';
     const MODE_SENDMAIL  = 'sendmail';
@@ -72,9 +73,16 @@ class MailSetting extends Model
         $this->smtp_authorization = !!strlen($this->smtp_user);
     }
 
+    public function getFailoverMailersOptions()
+    {
+        $config = App::make('config');
+        return array_keys($config->get('mail.mailers'));
+    }
+
     public function getSendModeOptions()
     {
         return [
+            static::MODE_FAILOVER => 'system::lang.mail.failover',
             static::MODE_LOG      => 'system::lang.mail.log_file',
             static::MODE_MAIL     => 'system::lang.mail.php_mail',
             static::MODE_SENDMAIL => 'system::lang.mail.sendmail',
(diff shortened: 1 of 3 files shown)


modules/system/models/MailSetting.php
```diff
@@ -72,9 +73,16 @@ public function initSettingsData()
         $this->smtp_authorization = !!strlen($this->smtp_user);
     }
```
matteotrubini, line 76:
```suggestion
        $this->failover_mailers = implode(',', $config->get('mail.mailers.failover.mailers', []));
    }
```

modules/system/models/mailsetting/fields.yaml
```diff
@@ -80,6 +80,14 @@ tabs:
                 field: send_mode
                 condition: value[sendmail]
 
+        failover_mailers:
+            label: system::lang.mail.failover_mailers
+            placeholder: system::lang.mail.failover_mailers_placeholer
+            type: taglist
+            options: getFailoverMailersOptions
+            customTags: false
+            tab: system::lang.mail.general
```
matteotrubini, line 89:
```suggestion
            tab: system::lang.mail.general
            trigger:
                action: show
                field: send_mode
                condition: value[failover]
```

modules/system/models/MailSetting.php
```diff
@@ -72,9 +73,16 @@ public function initSettingsData()
...
 
+    public function getFailoverMailersOptions()
+    {
+        $config = App::make('config');
+        return array_keys($config->get('mail.mailers'));
```
matteotrubini:
```suggestion
        $mailers = App::make('config')->get('mail.mailers');
        unset($mailers['failover']);
        return array_keys($mailers);
```
To avoid having `failover` in the list of transports.

matteotrubini (changes requested):
add this validation rule:
`'failover_mailers' => 'required_if:send_mode,'.self::MODE_FAILOVER`

modules/system/models/MailSetting.php
```diff
@@ -91,6 +99,10 @@ public static function applyConfigValues()
         $config->set('mail.from.address', $settings->sender_email);
 
         switch ($settings->send_mode) {
+            case self::MODE_FAILOVER:
+                $config->set('mail.mailers.failover.mailers', explode(',', $settings->failover_mailers) ?? ['log']);
```
matteotrubini:
```suggestion
                $config->set('mail.mailers.failover.mailers', explode(',', $settings->failover_mailers));
```
`explode()` will never return `NULL`

matteotrubini (approved):
LGTM

LukeTowers:
@mjauvin are people able to configure the transports through the backend as well as use the failover transport?

mjauvin:
> @mjauvin are people able to configure the transports through the backend as well as use the failover transport?

No, they would need to configure the other mailers in the config

mjauvin:
As I wrote in discord:

<@164272636988227584> I'm thinking we should move away from the SettingsModel behavior for the MailSetting model and have an active record instead (for the default mailer). That would allow setting multiple records for the different transports to support the failover (and roundrobin in Laravel 11)

matteotrubini:
> > @mjauvin are people able to configure the transports through the backend as well as use the failover transport?
> 
> No, they would need to configure the other mailers in the config

All mailers can be configured through the backend as usual, and they are all saved together. Therefore, in theory, you should first configure the individual mailers, then select and configure the failover, and finally save everything in a single action.

From a user perspective, it would probably be more intuitive if the configurations of the selected failover mailers were visible, preventing them from being "hidden."

PR #1318, opened 2025-02-26
Add fix for colorpicker hex with alpha saving

### LINKED ISSUE #1316: Colorpicker value never passes validation with showAlpha = true
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

If a colorpicker is configured to allow for transparency option (which comes as an extra two hex digits when formats = hex or "all") the value of this field with opacity at the end can never be saved as it doesn't pass the built-in validation.

### Steps to replicate

1. Create a form field of type colorpicker, configure it with showAlpha: true.
2. Try to submit the form

### Workaround

No known workaround

jaxwilko:
Should fix https://github.com/wintercms/winter/issues/1316#issuecomment-2685144593.

 I tested all supported formats (`cmyk, hex, hsl, rgb`) with and without `showAlpha: true` and they all save correctly.

### DIFF
--- a/modules/backend/formwidgets/ColorPicker.php
+++ b/modules/backend/formwidgets/ColorPicker.php
@@ -64,7 +64,7 @@ class ColorPicker extends FormWidgetBase
      */
     protected array $validationPatterns = [
         'cmyk' => '/^cmyk\((\d{1,2}\.?\d{0,2}%,? ?){4}\)$/',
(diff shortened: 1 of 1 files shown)


mjauvin:
Looks good to me.

@lex0r can you test this and let us know if that resolves the issue?

PR #1344, opened 2025-04-30
Added AssetContainer class that can be shared between AssetMaker objects

jaxwilko:
This PR allows implementors of the AssetMaker trait to share "assets".

This fixes the issue where a component owning it's own assets aren't added to the parent controller's assets.


LukeTowers:
Not needed at the moment, we can revisit this approach in the future if there's a good enough reason to do so.

Thanks though! ❤️

Issue #1347, opened 2025-05-01
Extend Filter widget API to allow redefining a scope


lex0r:
### Package targeted

Winter CMS

### Description

Currently it is possible to remove a scope via `removeScope()`, but this is not enough in case when a scope needs some adjustment. `allScopes` is a protected property and direct access is not possible (and obviously not recommended).

### Will this change be backwards-compatible?

Adding a setter method will help in certain cases. It will be optional to use it.

mjauvin:
Are you using the event for this?

ref. https://wintercms.com/docs/v1.2/api/events/backend/filter/extendScopes

mjauvin:
If you can describe your use case in more details, we can identify if this is indeed a needed functionality.

lex0r:
@mjauvin yes I use `listFilterExtendScopes` on my controller instance. I can imagine why you think it's not "needed". I was even thinking about completely dropping the idea of modifying and extending the original `winter/user` plugin, something I decided to do because I thought I would get nice updates from the community, and because it seemed to be a "core" feature (which I think it de facto is but somehow not included in the core and optional). So, if I did drop the idea, I wouldn't be creating this PR and would save myself some time :) But here we are and yes, as one may have guessed, when you want to extend Winter.User you may be forced to go to such lengths as redefining certain scopes, and the only way to do that without creating a patch, is to implement a method like setScope.

mjauvin:
But why not remove the scope and add it back with the changes?

lex0r:
Because it's two actions, and one action is easier.

bennothommo:
@lex0r technically you can just use the `addScopes` method alone to modify an existing scope. Unless I missed it, there's no check for duplicate scope names when using that method, so you could just overwrite the existing scope with a new scope config.

PR #1348, opened 2025-05-01
Extend Filter with a method to redefine a scope. Addresses #1347

lex0r:


LukeTowers:
@lex0r can you provide a usage example for this?

PR #1354, opened 2025-05-13
Prevent cache collision across different themes when combining assets

Fl0Cri:
## Problem description
When providing an array to the `| theme` twig filter, some processing is done to figure out which actual files should be referenced, i.e. when inheriting from a parent theme.

The result is stored in cache, with the cache key being derived from the input paths only, which doesn't contain the active theme.

I encountered an issue when using 2 different themes (for different hostnames), each one having its own styles, but using a similar structure and thus the same file names.

In this case, when visiting my first theme, paths such as `assets/scss/theme.scss` get resolved to `/.../themes/myfirsttheme/assets/scss/theme.scss`. If I then visit my second theme, the cached result is used and I get the styles from my 1st theme. If I clear the cache, this time `assets/scss/theme.scss` resolves to `/.../themes/mysecondtheme/assets/scss/theme.scss` (but the 1st theme will now have the wrong styles).

I guess adding the theme directory to the cache key would make sense. There is no side effect I could think of.

## Steps to reproduce
1. Starting from a fresh installation
2. In `themes/demo/layouts/default.htm`, put square brackets around `assets/javascript/app.js` to make it use the combiner
3. Duplicate the `demo` theme, I will call the new one `demo2`
4. In `themes/demo/assets/javascript/app.js`, add an `alert('demo1')` somewhere
5. In `themes/demo2/assets/javascript/app.js`, add `alert('demo2')`

Now when you load a page, you should see the alert "demo1". In `config/cms.php` change the `activeTheme` to `demo2` and reload, you will still have "demo1" instead of "demo2".

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -1392,7 +1392,7 @@ class Controller
         $parentTheme = $this->getTheme()->getConfig()['parent'] ?? false;
         $themesPath = themes_path();
 
-        $cacheKey = __METHOD__ . '.' . md5(json_encode($url));
+        $cacheKey = __METHOD__ . '.' . $themeDir . '.' . md5(json_encode($url));
 
         if (!($assets = Cache::get($cacheKey))) {
             $assets = [];
--- a/modules/cms/tests/classes/ControllerTest.php
+++ b/modules/cms/tests/classes/ControllerTest.php
@@ -96,6 +96,22 @@ class ControllerTest extends TestCase
         $this->assertTrue(in_array('style2.css', $files));
     }
 
+    public function testThemeCombineAssetsInDifferentThemes(): void
+    {
+        $themeA = Theme::load('test');
+        $controllerA = new Controller($themeA);
+
+        $themeB = Theme::load('childtest');
+        $controllerB = new Controller($themeB);
+
+        // Generate a url for the same file in both themes.
+        // Because the files are different, the urls should be too.
+        $this->assertNotEquals(
+            $controllerA->themeUrl(['assets/css/style2.css']),
+            $controllerB->themeUrl(['assets/css/style2.css']),
+        );
+    }
+
     public function testPageUrl()
     {
         $theme = Theme::load('test');


LukeTowers:
@Fl0Cri the best way to handle this is to set your cache key prefix in your config/cache.php based on the app URL or hostname; there's likely to be other things that make use of the cache (i.e. components) inside of your theme that would also need this separation. Let me know if you disagree or if you have any questions about how to set that up and we can revisit this.

Fl0Cri:
I encountered this issue when migrating an app from October 1 to Winter. I didn't experience any cache-related issues in October, not sure if that was intended support or if I simply wasn't using any affected feature.

However, if you don't feel confident about supporting this, I agree that changing the prefix is a better solution.

When I tried to do this, it didn't work because the file cache driver doesn't support prefixes (that's a Laravel thing).

After some more digging, it looks like you already solved my problem 2 years ago here https://laracasts.com/discuss/channels/laravel/unable-to-prefix-cache-in-multi-tenant-app ;-)
Thanks for the idea to change the cache path instead.

One drawback of this approach though is that `php artisan cache:clear` (for example in a deployment script) will only clear the cache of the "primary" site. But that's something I can work around, hopefully by hooking to an event of the cache:clear command.

P.S. Do you have any opinion against changing the cache config dynamically from a plugin in the context of WinterCMS? There will be 3 "cache namespaces" (one for each site and one before the plugin registers), but I don't think it's an issue.

The other approach would be to use a different environment by hostname, set using `config/environment.php`, but I would prefer to avoid this: we have queues that need to run code in the context of a specific site, for example to swap mail config, and I fear that using a different code path for this could make issues harder to detect. (A broken frontend can be quickly spotted by basic QA testing, scheduled notifications in a background process cannot)

LukeTowers:
@Fl0Cri right, I remember running into that very same issue. Good to know past me is still making present me's life easier sometimes 😂 

Changing the cache dynamically from a plugin isn't a huge deal; in theory if a ton of different plugins were all dynamically changing the configs you could get headaches, but honestly with how Winter projects are typically setup those sorts of changes usually only happen in a project specific plugin anyways so you should be fine.

Using config/environment.php and config/$environment/$config.php is a fine approach to, it makes it slightly more obvious what is happening than doing everything dynamically deep inside of a plugin; but ultimately either approach should work fine. Up to you which way you want to go.

Issue #1247, opened 2024-11-04
Backend anchors have extra styling in firefox


nmiyazaki-chapleau:
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Issue description

Firefox adds a blue outline to anchor elements on :focus (which honestly looks horrible) and interferes with the look & feel of WinterCMS.
Some example of problematic places include sections with tabs - on click, which makes the element focused, it add a small blue outline on the left and right side of the element.
Example in the CMS tabs: 
![Image](https://github.com/user-attachments/assets/f7867c77-278a-4003-a85e-df5ac637ef51)
Example in the underlying section:
![Image](https://github.com/user-attachments/assets/8b5db0f8-e2d1-451c-ac44-8fecead46651)

### Steps to replicate

1. Open firefox
2. Go to backend
3. Click on anything that is an anchor

### Workaround

Can be fixed using `a:focus, a:focus-visible { outline: none; }`. However, I do not know if this is intended behavior.

damsfx:
Same for form primary tabs 
![Image](https://github.com/user-attachments/assets/3f5d1e01-e429-4547-9008-eb4dcb45cd3a)

LukeTowers:
So long as the focused element has some form of styling to distinguish it for usability I'd be more than happy to merge any PR suggested. Let me know if you need help submitting one!

bennothommo:
I remember this being added back in the October days. It's horrific. I would very much like a more elegant solution.

IsaiahPaget:
Hey there Luke,

I am a little later than I thought I would be to this. On the same day you reached out to me, I was moved to the second stage of an interview process for a full-time gig, so I had to make sure I studied for that.

This looks like it's something fairly easy to accomplish as a first PR, I loaded up the dev branch and opened the application in Firefox, and lo and behold, it's there. So, from reading the previous comments, it sounds like it's important that the user knows that it's focused, but when using a mouse you get that feedback when the tab is set to active because it comes to the foreground. But there is that scenario when you use the Tab key, and then you can select it that way.

One thing I'd maybe clarify is that: Should the focus ring of all anchor tags be beautified or is it that just the tabs look ugly with the focus ring? As far as I can tell, the real difference between the two browsers is that the focus ring just doesn't show up when you click on the tab in Chrome, but it does when you click on the tab in Firefox.

For reference this is what the focus ring looks like on Chrome

[...]

PR #1361, opened 2025-06-06
Fix anchor tag styling on Firefox

IsaiahPaget:
I have added some CSS to fix this bug: https://github.com/wintercms/winter/issues/1247.

The issue was that, in Firefox only, clicking on an anchor tag would show an accessibility outline. This should only happen when navigating with the keyboard. To better align the behavior with Chrome, I added CSS that targets all anchor tags that are :focus but not :focus-visible, and sets their outline to none. But only if the browser is Firefox.

This means the outline will not appear unless the user is navigating with a keyboard. This improves the visual experience, without breaking any existing features, it only affects anchor tags and is purely a visual adjustment.

I figured I'd put this in the layout.less file because this seems like the most global file, I am not sure if this is the best place, happy to make changes as seen fit.

I initially thought this change would be simple and quick, but it took me a second to trace where all the CSS was coming from. Factoring in the time spent learning this small part of the codebase, I figured it would take me the better part of an afternoon, and that's about how long it took me.

### DIFF
--- a/modules/backend/assets/less/layout/layout.less
+++ b/modules/backend/assets/less/layout/layout.less
@@ -22,6 +22,12 @@ body.no-select {
     cursor: default !important;
 }
 
+// Remove focus outline for mouse clicks, keep for keyboard navigation (Only needs to happen on Firefox)
+@supports (-moz-appearance: none) {
+    a:focus:not(:focus-visible) {
+        outline: none;
+    }
+}
 //
 // Layout canvas
 //
--- a/modules/backend/assets/less/layout/layout.less
+++ b/modules/backend/assets/less/layout/layout.less
@@ -22,12 +22,6 @@ body.no-select {
     cursor: default !important;
 }
 
-// Remove focus outline for mouse clicks, keep for keyboard navigation (Only needs to happen on Firefox)
-@supports (-moz-appearance: none) {
-    a:focus:not(:focus-visible) {
-        outline: none;
-    }
-}
 //
 // Layout canvas
 //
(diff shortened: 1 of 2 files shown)


modules/backend/assets/less/layout/layout.less
```diff
@@ -22,6 +22,12 @@ body.no-select {
     cursor: default !important;
 }
 
+// Remove focus outline for mouse clicks, keep for keyboard navigation (Only needs to happen on Firefox)
+@supports (-moz-appearance: none) {
+    a:focus:not(:focus-visible) {
+        outline: none;
+    }
+}
```
LukeTowers: Let's move this to the bottom of the file under a section called `Browser specific` like the `Screen specific` section we have right now, and then recompile the assets with `php artisan mix:install` & `php artisan winter:util compile less`.

IsaiahPaget:
@LukeTowers Hey, so here is the update. I'll write this down in the google doc, but I think I was a little thrown off by the fact that I have to include the compiled css in the commit. I am not sure if this is already in the docs but maybe in the contributing.md it could be worth it to put a little excerpt about that fact that you have to stage the compiled css. I know it says not to mess with it directly, but I didn't intuit that you also have to commit it.

I am also wondering: some of the config files got updated when I first 'built' the project, do you guys always just stage the specific files that you made changes to, or is there some config I add to my .gitconfig or something that will avoid commiting instance specific changes, like the package.json or something of that nature?

LukeTowers:
@IsaiahPaget the .gitignore for core development is kind of tricky, as the values we'd use for core development (i.e. tracking changes in modules, ignoring instance-specific files like package.json, the lock files, etc) is different from the gitignore that we need to ship to our end users (i.e. people that consume the `wintercms/winter` package). I know we already have some level of configuration to strip out files for the people consuming the package itself through composer `create-project`, so if there was some way to switch between a "I'm working on the project as a core contributor right now" and a "this is the .gitignore the project recommends for anyone building a Winter CMS project"; that would be a fantastic improvement. 

I haven't looked into doing that at all but if that's something you'd be interested in investigating that'd be very helpful 😄

PR #1371, opened 2025-06-22
Refactor: properly handle callable array even when options are missing

matteotrubini:
Improved handling for cases where `$callable` is an array but does not include the `$callable['options']` key. In such cases, the default options are preserved and no errors occur.

Also streamlined the normalization logic by explicitly checking for valid entries (`!empty()`)

⚠️ If this logic proves to be correct and stable, it should be applied to the `makeTwigFilters()` function as well for consistency.
Eventually, both implementations could be unified into a shared helper, as they follow the same logic.

### DIFF
--- a/modules/system/classes/MarkupManager.php
+++ b/modules/system/classes/MarkupManager.php
@@ -323,7 +323,7 @@ class MarkupManager
             if (!is_callable($callable)) {
                 throw new SystemException(sprintf(
                     'The markup %s (%s) for %s is not callable.',
-                    strtolower(class_basename($twigClass)),
+                    $twigClass,
                     json_encode($callable),
                     $name
                 ));
--- a/modules/system/classes/MarkupManager.php
+++ b/modules/system/classes/MarkupManager.php
@@ -165,10 +165,10 @@ class MarkupManager
 
     /**
      * Returns a list of the registered Twig extensions of a type.
-     * @param $type string The Twig extension type
+     * @param string $type The Twig extension type
      * @return array
      */
-    public function listExtensions($type)
+    public function listExtensions(string $type): array
     {
         if ($this->items === []) {
             $this->loadExtensions();
@@ -181,7 +181,7 @@ class MarkupManager
      * Returns a list of the registered Twig filters.
      * @return array
      */
-    public function listFilters()
+    public function listFilters(): array
     {
         return $this->listExtensions(self::EXTENSION_FILTER);
     }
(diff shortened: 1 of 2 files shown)


LukeTowers:
@matteotrubini can you add some test cases for this?

matteotrubini:
@LukeTowers, once I get your green light, I’ll proceed with makeTwigFilters() as described in the first comment.

LukeTowers:
@matteotrubini ideally the options logic that's present everywhere in Winter could be centralized to a helper as much as possible. Should be good to proceed with the makeTwigFilters

matteotrubini:
I have finished my review and the tests are green. I would squash the commits before marking the PR as ready - what do you think, @LukeTowers?

LukeTowers:
@matteotrubini I always squash when merging, no need to worry about squashing on the PR itself.

modules/system/classes/MarkupManager.php
```diff
@@ -374,4 +293,77 @@ protected function isWildCallable($callable, $replaceWith = false)
...
+
+    /**
+     * Shared logic for creating Twig callable (functions/filters).
+     * @param array $definitions
+     * @param string $twigClass
+     * @param array $defaultOptions
+     * @param array $collection
+     * @return array
+     */
+    private function makeTwigCallable(array $definitions, string $twigClass, array $defaultOptions, array $collection = []): array
```
LukeTowers:
```suggestion
    private function makeTwigCallables(array $definitions, string $twigClass, array $defaultOptions, array $collection = []): array
```

LukeTowers:
Is the intention behind this to allow definitions using October v4's syntax? (i.e. https://github.com/OFFLINE-GmbH/oc-responsive-images-plugin/commit/741e145a71df7838b1526747da43d522141ed02d, https://github.com/octobercms/october/blob/8b8d2f7444ea98da488014504f51e12f4150465a/modules/system/classes/MarkupExtensionItem.php#L80-L98). If that's the case I think we need to be a bit more explicit about how we're supporting that to make sure that we're also properly parsing the options.

matteotrubini:
The purpose of this PR is the one described in the first comment, and it was achieved with the initial commit.
Consequently, Offline.ResponsiveImages v2.9.0 has become functional.
The subsequent commits focus on unifying shared logic around the class.
All changes were made without access to the October v4 source code, and therefore without considering any change that may have been introduced.

PR #1382, opened 2025-07-08
Run a more robust count query that supports "group by"

### LINKED ISSUE #454: When using listExtendQuery with groupby, paginate seems not work
### Winter CMS Build

Other (please specify below)

### PHP Version

7.4

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

Hello everyone.
I have a model listing, where I need to override the standard querie, using a groupBy that returns the desired results, according to the business rule. The values ​​that appear as "totals" in the pagination are correct.

The problem is, when doing this:
 public function listExtendQuery($query)
    {
     $query->groupBy('column');
    }
    
  And when trying to move from one page to another, it just doesn't load. I get stuck on the first page.
  See: the pagination widget is loaded, but it doesn't work when I click next page...
  
  Does anyone have any ideas?

### Steps to replicate

• Use any model with some data
• In the controller of above's model, use the listExtendQuery to override default querie using: $query->groupBy('some_column')
• Try to click in the next step indicator

### Workaround

Nothing yet.

lex0r:
Fixes https://github.com/wintercms/winter/issues/454

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -785,7 +785,7 @@ class Lists extends WidgetBase
         $currentPageNumber = intval($currentPageNumber);
 
         if ($currentPageNumber > 1) {
-            $count = $query->count();
+            $count = $query->toBase()->getCountForPagination();
 
             // If the current page number is higher than the amount of available pages, go to the last available page
             if ($count <= (($currentPageNumber - 1) * $this->recordsPerPage)) {


mjauvin:
@lex0r thanks for this PR.

Can you give a bit more details on exactly what those methods do, I am not familiar with them.

lex0r:
@mjauvin both the original `count()` method and `getCountForPagination()` are available in `Illuminate\Database\Query\Builder`, but the main difference is that `count()` is much simpler and I suppose was never meant to be used for pagination, while the second one has support for more complex queries that may have grouping.

mjauvin:
@lex0r what about `toBase()`?

Also, unit test for this would be very appreciated.

damsfx:
> @lex0r what about `toBase()`?

`toBase()` will retrieve the data from the database but it will not prepare the associated Eloquent models.  
If you are not really going to use features like Mutators, Relationships of Eloquent and loading a large amount of data, then the toBase function can be a game changer in performances.

mjauvin:
I'm worried that the code here will have a hard time because of this:

https://github.com/wintercms/winter/blob/054fa0552f1362decbff67abd2d462384700622e/modules/backend/widgets/Lists.php#L738

lex0r:
> > @lex0r what about `toBase()`?
> 
> `toBase()` will retrieve the data from the database but it will not prepare the associated Eloquent models. If you are not really going to use features like Mutators, Relationships of Eloquent and loading a large amount of data, then the toBase function can be a game changer in performances.

The good news is that we only need to run a count query, so not interested in any actual record data.

> I'm worried that the code here will have a hard time because of this:
> 
> https://github.com/wintercms/winter/blob/054fa0552f1362decbff67abd2d462384700622e/modules/backend/widgets/Lists.php#L738

@mjauvin not sure I'm following your concern. The code I suggested is absolutely pagination friendly, it's even in the name of the method. Surely it will work whatever pagination method is used?

lex0r:
> @lex0r what about `toBase()`?
> 
> Also, unit test for this would be very appreciated.

I would expect a unit test that checks for what records are returned based on the requested page to already exist. If it doesn't exist then it's probably not very important, but even if it is, I can't promise I will have time to look into creating a test for pagination unless there's some ground work that has been done and it won't take long to implement :)

LukeTowers:
@lex0r if you aren't able to create a unit test case for this can you create a PR in the [test plugin](https://github.com/wintercms/wn-test-plugin) that replicates the original issue and provides instructions for testing this fix?

PR #1383, opened 2025-07-09
Add fix for error not being correctly reported

jaxwilko:
A misconfiguration can cause an issue that is currently being incorrectly reported.
image
With this patch, we only use `$formConfig` if defined, if not we default to a js back. This results in the correct information being presented to the user:
image


LukeTowers:
What conditions trigger this?

jaxwilko:
I messed up a relation configuration, but it didn't tell me because it was masked by the undefined var error

LukeTowers (approved):

PR #1385, opened 2025-07-10
remove columns limit query

wpjscc:
For example, if a user has two payment orders, only one payment order will be counted when there is a `limit`. Now, the `limit` is removed, but we can still restrict it by setting `limit: 1` in the `conditions`.

```yaml
user_orders_real_amount:
      label: real_amount
      relations: user_orders
      limit: 1
      select: 'sum(real_amount)'
```

```yaml
user_orders_max_id:
      label: user_orders_max_id
      relations: user_orders
      limit: 1
      select: 'user_orders.id'
      conditions:  order by  user_orders.id desc limit 1
```

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -575,10 +575,6 @@ class Lists extends WidgetBase
                     $joinQuery->whereRaw(DbDongle::parse($column->config['conditions']));
                 }
 
-                if ($limit) {
-                    $joinQuery->limit($column->config['limit']);
-                }
-
                 $joinSql = $joinQuery->toSql();
 
                 $selects[] = DB::raw("(" . $joinSql . ") as " . $alias);


LukeTowers:
@wpjscc I don't understand the reason for this PR, why are you removing the feature?

wpjscc:
for example 

Model: User
table: users

| id  | name  |
|------|---------|
| 1    |  John       |


User hasMany
```
public $hasMany=['user_orders' => [UserOrder::class ]];
```

Model: UserOrder
 table: user_orders

| id  | user_id | real_amount |
|------|---------|-------------|
| 1    | 1     | 50.00       |
| 2    | 1    | 50.00       |

```
user_orders_real_amount:
      label: real_amount
      relations: user_orders
      limit: 1
      select: 'sum(real_amount)'

```
generate sql is 
```
(select sum(real_amount) from user_orders where users.id =  user_orders.user_id limit 1) as user_orders_real_amount
```
the user_orders_real_amount result is `50`,it should is `100`

the PR  generate sql is 

```
(select sum(real_amount) from user_orders where users.id =  user_orders.user_id) as user_orders_real_amount
```

it result is `100`

LukeTowers:
@wpjscc Yes, but why are you even setting a limit value if you don't want your query limited? That's the part that I don't understand, you don't need to remove the feature for everyone when you can just remove that line from your YAML.

wpjscc:
if do not have `limit : 1`

it will generate sql

```
(select group_concat(sum(real_amount) separator ', ') from user_orders where users.id =  user_orders.user_id) as user_orders_real_amount
```

There is a contradiction. If we use relations, then we must use limit:1 to call [DbDongle::raw($sqlSelect)](https://github.com/wintercms/winter/blob/52661d683996017c5c6fa834b4ad835e6f66cf69/modules/backend/widgets/Lists.php#L568). In this case, the order result is 50, but we expect it to be 100

So how can we generate such SQL when using relations

```
(select sum(real_amount) from user_orders where users.id =  user_orders.user_id) as user_orders_real_amount
```

rather than

```
(select group_concat(sum(real_amount) separator ', ') from user_orders where users.id =  user_orders.user_id) as user_orders_real_amount
```

PR #1386, opened 2025-07-13
Fix: prevent null/undefined in Snowboard.request createFormData

phpiando:
### Problem

When using `Snowboard.request()` with `data` that includes `null` or `undefined` values, a `TypeError` is thrown due to `Object.entries()` not accepting such values.

### Fix

A simple check for `null` or `undefined` was added at the beginning of the `createFormData()` function to prevent the error and silently skip those fields.

### Reproduction Example

```js
Snowboard.request('onSomething', {
    data: {
        name: 'phpiando',
        optionalField: null // This causes an error in current version
    }
});
```

This PR prevents that error.

### DIFF
--- a/modules/system/assets/js/snowboard/ajax/Request.js
+++ b/modules/system/assets/js/snowboard/ajax/Request.js
@@ -815,6 +815,10 @@ export default class Request extends PluginBase {
      * @returns {void}
      */
     createFormData(formData, data, prefix = '') {
+        if (data === null || data === undefined) {
+            return;
+        }
+
         if (typeof data !== 'object') {
             formData.append(prefix, data);
             return;


LukeTowers:
@phpiando so with this code change your example code would actually just send a request with `name` defined in the form data and not provide anything for `optionalField` at all? I feel like that's not necessarily what I would expect to happen. @bennothommo any thoughts on this one?

phpiando:
> @phpiando so with this code change your example code would actually just send a request with `name` defined in the form data and not provide anything for `optionalField` at all? I feel like that's not necessarily what I would expect to happen. @bennothommo any thoughts on this one?

Yep! this change would result in fields with null or undefined values being skipped entirely in the FormData.
In our use case, that behavior is preferred to avoid errors like:
```
TypeError: Cannot convert undefined or null to object
```

But!
Maybe a better solution would be to treat null and undefined as empty strings when appending to FormData, like:
```
if (data === null || data === undefined) {
    formData.append(prefix, '');
    return;
}
```

That way, the key is still sent, but it doesn't cause errors.

LukeTowers:
That could work, but I'll wait for @bennothommo's thoughts on this

bennothommo:
@phpiando @LukeTowers apologies for the delay in responding to this one.

So, as far as I know, this problem will only exhibit with a `null` value because `typeof null` equals an object, but the `Object.entries` method needs an actual object, hence the type error. With `undefined`, it would be not be an object and therefore would go to `formData.append()`, which I believe will convert it into an empty string as per [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append#value).

As for how to handle `null` and `undefined` - I consider both to be the *absence* of a value, and thus should result in that "field" not being included at all, so @phpiando's initial fix is probably the correct one.

The only thing I would ensure is that the following data definition:

```js
    data: {
        myField: {
            inner: {
                child: {
                    field: undefined,
                }
            }
        }
        myField2: 'value',
    },
```

Results only in the `myField2` field being sent across to Winter. If you can confirm that works with your fix, I'd be happy with that.

phpiando:
@bennothommo @LukeTowers Hi!

Hi again,

I just ran two tests to verify the behavior:

1. Only undefined value:
```
payload = {
    myField: {
        child: {
            inner: {
                field: undefined
            }
        }
    },
    myField2: 'value'
};
```
✅ Result: only myField2 is included in the FormData.
myField is entirely skipped, as it only contains undefined — exactly as expected.

2. Mixed values (undefined + valid):

```
payload = {
    myField: {
        child: {
            inner: {
                field: undefined,
                value: 'present'
            }
        }
    },
    myField2: 'value'
};
```
✅ Result: both myField2 and myField[child][inner][value] are included in the FormData.
Only the undefined field is ignored — the rest is preserved.

So yes, the fix behaves exactly as intended:
undefined and null are omitted
valid values inside nested objects are still captured correctly

Let me know if you'd like anything else updated in the PR!

LukeTowers:
Thanks for the contribution @phpiando! ❤️

PR #1300, opened 2025-01-25
Repeater: Always send a value in the Form data

mjauvin:
Fixes #1290

Issue #1290, opened 2025-01-20
1.2.7 cannot save empty repeater


AIC-BV:
### Winter CMS Build

dev-develop

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Issue description

Since 1.2.7 it is no longer possible to save an empty repeater.
Removing all items and saving, will result in no changes at all.
Removing 10 items, leaving one and saving, will result in one item in the repeater. Removing the last item will not work.

### Steps to replicate

1. Create a repeater with content
2. Save
3. Remove the content
4. Save
=> The content is still there

### Workaround

Delete the data through phpmyadmin/sql workbench/...

AIC-BV:
I figured when downgrading backend module to 1.2.6 it works again: ```"winter/wn-backend-module": "1.2.6"```
(and Tailwind UI to dev-main#42dabcdf7e0daaa2b0dddd5fd5117412e8577e1a because latest version depends on Vite which was added in backend-module 1.2.7)

AIC-BV:
I think it is because of this
https://github.com/wintercms/wn-backend-module/commit/9cc2bd87aad56d71c38fcbdc314d5f97516dbdbe

But reverting this also seems to break the repeater (all empty repeaters get an empty row)
@LukeTowers any idea?

LukeTowers:
@AIC-BV is this still an issue on the latest develop branch? @mjauvin any thoughts?

mjauvin:
> [@AIC-BV](https://github.com/AIC-BV) is this still an issue on the latest develop branch? [@mjauvin](https://github.com/mjauvin) any thoughts?

Submitted a PR.

LukeTowers:
@AIC-BV can you test #1300 and let me know if that solves the issue?

AIC-BV:
@LukeTowers Solved! ❤

PR #1390, opened 2025-07-25
Preserve clearing of Repeater form widgets on model save

sniemz:
Issue:
When a user removes all values from a Repeater widget, the widget submits no POST data. This causes the form save process to skip the field entirely, unintentionally preserving old data.

Suggested Improvement:
Detect when a widget was rendered (`_loaded` flag present) but submitted no data. In these cases, interpret the absence of data as the user intentionally clearing the field, and persist this empty state in the database.

Example:
A backend form has a Repeater widget to manage a list of links:

```
#  Form Field Definitions


links:
    label: Useful Links
    type: repeater
    form:
        fields:
            label:
                label: Label
                type: text
            url:
                label: URL
                type: text
```
Adding one or multiple items works as expected.

Removing a few items also works as expected.

Removing **all** items previously did nothing and the old data was retained.

With this change, removing all items should now correctly clear the field on model save.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -215,10 +215,6 @@ class Repeater extends FormWidgetBase
      */
     public function getSaveValue($value)
     {
-        if (empty($value)) {
-            return [];
-        }
-
         return $this->processSaveValue($value);
     }
 
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -215,6 +215,10 @@ class Repeater extends FormWidgetBase
      */
     public function getSaveValue($value)
     {
+        if (empty($value)) {
+            return [];
+        }
+
         return $this->processSaveValue($value);
     }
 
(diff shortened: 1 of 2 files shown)


sniemz:
Hey @LukeTowers , just double-checking you're not waiting on anything from me here?
Cheers

LukeTowers:
@sniemz I was trying to replicate this issue with the Record model in the [Test Plugin](https://github.com/wintercms/wn-test-plugin) but was unable to as deleting the repeater items seemed to work fine. I did however come across another somewhat related issue where the nestedform in the repeater was having it's value be polluted from the root level `name` field instead of properly scoping the value retrieval to the current field level (of being inside the repeater) when getting the values for the field from the model.

Could you provide me replication instructions using the test plugin?

My main concern with this PR right now is it's applying a change specific to the Repeater formwidget inside of the Form widget class, which makes the parent somewhat dependent on the child and is a bit of a code smell / likely to cause issues in the future.

sniemz:
This is a duplicate of https://github.com/wintercms/winter/issues/1290
and has been fixed in 1.2.8 by this PR https://github.com/wintercms/winter/pull/1300/files

LukeTowers:
This was fixed in 1.2.8

PR #1397, opened 2025-08-15
Add fix to inform client when application is not set up

jaxwilko:
If the application has not been correctly set up, we are currently `exit(1)` and then returning a 200.

This change ensures that if the context of the request is web, we inform the client that things are not okay.


jaxwilko:
Thanks GitHub, I didn't close this I merged a PR in a private repo that mentioned this PR... GJ

LukeTowers (approved):

PR #1426, opened 2025-12-06
Add php override file used by many shared hosting

mjauvin:
## Summary by CodeRabbit

* **Chores**
  * Updated system configuration to include additional files in the mirrored assets list, enhancing the platform's file distribution and availability handling.

### DIFF
--- a/modules/system/console/WinterMirror.php
+++ b/modules/system/console/WinterMirror.php
@@ -33,6 +33,7 @@ class WinterMirror extends Command
      */
     protected $files = [
         '.htaccess',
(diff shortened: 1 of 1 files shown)


LukeTowers:
Why do you want that file mirrored to the public directory @mjauvin? That seems like the kind of file one wouldn't want mirrored at all.

mjauvin:
Because it's required when the site uses the mirror for the overrides to be seen by the webserver.

It's the equivalent of `php.ini`

LukeTowers (approved):

Issue #431, opened 2022-02-03
Switch code editor to Monaco


bennothommo:
![Switch to Monaco](https://user-images.githubusercontent.com/15900351/152263359-7f6a36d1-34e5-4fa3-9aad-b6fd35ac43b6.png)

We are planning to update the code editor to use the [Monaco](https://microsoft.github.io/monaco-editor/) that powers [Visual Studio Code](https://github.com/microsoft/vscode). With this, we will gain an exceptional amount of features and control over the experience of editing code within Winter CMS, using an interface that users are comfortable with.

We will be able to take advantage of the extension work done in our [official VSCode extension](https://github.com/wintercms/vscode-extension/issues/9) to provide the same level of functionality in the Monaco editor.

PR #1431, opened 2025-12-20
Add blade support for cms/backend partials

mjauvin:
### from twig:
```twig
{% partial "my_partial.blade" %}
```

### from PHP:
```php
$cmsController->renderPartial("my_partial.blade");
```
Actual partial file path should be: `/partials/my_partial.blade.php` 

```php
$backendController->makePartial("my_partial.blade", array $vars);
```
Actual partial file path should be: `_my_partial.blade.php` 

## Summary by CodeRabbit

* **New Features**
  * Added support for Blade-style partials in the template system, enabling additional template rendering options alongside existing partials.

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -25,6 +25,7 @@ use Winter\Storm\Exception\AjaxException;
 use Winter\Storm\Exception\ValidationException;
 use Winter\Storm\Parse\Bracket as TextParser;
 use Illuminate\Http\RedirectResponse;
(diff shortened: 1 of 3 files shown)


LukeTowers:
@mjauvin off the top of my head I think we'd need to restrict this with either a config, or a property in theme.yaml, or both; since it would allow PHP code to be written in the markup section of a CMS object which would bypass the safemode check. Additionally the CMS / Theme Editor would need blade support added (would be a lot easier if we were to get the #431 PR merged (swapping Ace with Monaco).

PR #1433, opened 2025-12-28
Improve grouped repeater UX by adding search and multiple columns

LukeTowers:
Screenshot 2025-12-28 at 4 18 03 PM
Screenshot 2025-12-28 at 4 18 09 PM

## Summary by CodeRabbit

* **New Features**
  * Added client-side search UI with live filtering by title or description, clear button, no-results messaging, and autofocus behavior.
  * Redesigned repeater items from a vertical list to a responsive grid of clickable cards with improved hover styling and visual hierarchy.

### DIFF
--- a/modules/backend/formwidgets/repeater/assets/js/repeater.js
+++ b/modules/backend/formwidgets/repeater/assets/js/repeater.js
@@ -139,6 +139,78 @@
             })
 
         $('[data-repeater-add]', $container).data('request-form', $form)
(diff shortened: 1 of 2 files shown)


jaxwilko (approved):
It is better than what's there but also less than ideal. The inline styles are horrible but hopefully soon we can convert all of this to Tailwind and then this will not be a problem :)

PR #1435, opened 2026-01-08
Add fieldset formwidget

mjauvin:
This formwidget wraps form elements with a `<fieldset>` HTML tag.

The fields are visually grouped but saved as if they were regular fields at the parent level.

The field's label config is used for the `<fieldset>` legend.

The Form widget's `getSaveData()` delegates to the FieldSet's own `getSaveData()`, which reuses the nested form's per-field save logic (number casting, nested-widget `getSaveValue()`, `NO_SAVE_DATA` exclusion, and disabled/hidden handling) to pull the grouped fields up to the parent form's level.

image

The field config is akin to a nestedform:
```yaml
myFieldSet:
  type: fieldset
  label: Grouped Fields
  fields:
      ...
```

## Summary by CodeRabbit

* **New Features**
  * Adds a FieldSet form widget to embed a nested form section with an optional legend and an option to hide labels.

* **Style**
  * New themed styling for the FieldSet, including dark-mode support and improved tab visuals.

* **Bug Fixes**
  * Nested widget fields are now collected and saved correctly during form submission.


damsfx:
@mjauvin How is the responsive design being managed?  
Does the span left/right/auto or custom Css with `span:storm` still works ?

mjauvin:
> @mjauvin How is the responsive design being managed? Does the span left/right/auto or custom Css with `span:storm` still works ?

It's pretty much derived from a nestedform ... so it renders a form ***within*** the `<fieldset>` tag.

The main difference is that fields are exported to the parent when saving (does not use a JSON field).

modules/backend/formwidgets/FieldSet.php
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * @inheritDoc
+     */
+    protected $defaultAlias = 'fieldset';
+
+    /**
+     * @var array Form configuration
+     */
+    public $form;
+
```
LukeTowers:
```suggestion
```

LukeTowers:
@mjauvin can you make a PR to the docs and include examples in the nestedform and fieldset sections as to when you should pick one vs the other?

mjauvin:
> @mjauvin can you make a PR to the docs and include examples in the nestedform and fieldset sections as to when you should pick one vs the other?

Done in https://github.com/wintercms/docs/pull/258

modules/backend/formwidgets/fieldset/assets/less/fieldset.less
```diff
@@ -0,0 +1,52 @@
...
+            > li > a > span.title {
+              &:before, &:after {
+                background-color: @panel-bg;
+                @media (prefers-color-scheme: dark) {
+                  background-color: @panel-bg-dark;
+                }
+              }
+            }
+        }
+    }
```
LukeTowers, line 51: Can you fix the indentation to be a consistent 4 spaces in this file?

modules/backend/formwidgets/fieldset/partials/_fieldset.php
```diff
@@ -0,0 +1,9 @@
+<?php $label = object_get($this->config, 'label'); ?>
+
+<fieldset class="fieldset">
+    <?php if ($label): ?>
+        <legend><?= trans($label) ?></legend>
```
LukeTowers:
```suggestion
        <legend><?= e(trans($label)) ?></legend>
```

modules/backend/formwidgets/FieldSet.php
```diff
@@ -0,0 +1,88 @@
+<?php
+namespace Backend\FormWidgets;
```
LukeTowers, line 3:
```suggestion

namespace Backend\FormWidgets;
```

modules/backend/widgets/Form.php
```diff
@@ -1221,6 +1221,23 @@ public function getSaveData(): array
...
+                         * Number fields should be converted to integers
+                         */
+                        if ($field->type === 'number') {
+                            $value = !strlen(trim($value)) ? null : (float) $value;
+                        }
+
+                        $this->dataArraySet($result, $parts, $value);
+                    }
+                }
+            }
```
LukeTowers, line 1242: Does this work with formwidgets that are embedded in the fieldset's form? Do they get added to the $this->formWidgets of the parent form or should we refactor this a bit so that it's a separate method / inline function that can be called again on every fieldset form inside of the the main form recursively?

LukeTowers (changes requested):

modules/backend/widgets/Form.php
mjauvin: Yeah, I don't think that works for embeded forwidgets.

I don't really have time to fix this today, so you may have to leave this out of 1.2.12 release, unfortunately.

mjauvin: @LukeTowers actually, formwidgets with nested fields should NOT have their internal fields added to the parent form... that part is fine.

But I tested with a tag formwidget and it's giving an error, so this is not 100% yet anyway, we should postpone.

mjauvin: On a sidenote, the main purpose of the fieldset formwidget is to group regular fields (mostly text/tel/url/email/checkboxes/radio)... so having other formwidget types could be useful, but this is not important in my opinion.

LukeTowers: If we don't want to support FormWidgets just yet then we'll need to throw a SystemException if any are detected when the form is initialized for the fieldset and be explicit that we don't support formwidgets in the fieldset yet in the documentation. Then we could be good to merge it.

mjauvin:
This is now working great with forwidgets inside the fielset. It won't work at multiple levels, though (e.g. widget inside a nestedform inside the fieldset).

modules/backend/widgets/Form.php
mjauvin: formwidgets are now supported.

LukeTowers:
@mjauvin if it won't work under certain conditions can we detect that and throw a systemexception when an unsupported configuration is detected?

LukeTowers:
Given the changes to the Form widget getSaveData I'd really like to see some unit testing around that, do we have anything there right now or do we need to add coverage from scratch?

mjauvin:
> Given the changes to the Form widget getSaveData I'd really like to see some unit testing around that, do we have anything there right now or do we need to add coverage from scratch?

It doesn't change Form's getSaveData(), it adds a specific case for the FieldSet widget if present. Current behavior is not changed if no FieldSet is used.

mjauvin:
> @mjauvin if it won't work under certain conditions can we detect that and throw a systemexception when an unsupported configuration is detected?

Not sure it's worth processing all fields recursively to find potential issues down below...

LukeTowers:
@mjauvin wouldn't all those fields have to be processed recursively anyways? What specific conditions cause the issue that you were talking about and what is the impact of that issue? Perhaps I didn't understand you fully the first time

mjauvin:
There is no more issues, they were resolved by calling widget's getSaveValue() when going through the FieldSets' internal fields (see below):

```php
            if ($widget instanceof FieldSet) {
                foreach ($widget->getFormFields() as $field) {
                    $parts = HtmlHelper::nameToArray($field->fieldName);
                    if (($value = $this->dataArrayGet($data, $parts)) !== null) {
                        switch ($field->type) {
                            case 'number':
                                $value = !strlen(trim($value)) ? null : (float) $value;
                                break;
                            case 'widget':
                                $value = $widget->getFormWidget($field->fieldName)->getSaveValue($value);
                                break;
                        }
                        $this->dataArraySet($result, $parts, $value);
                    }
                }
                continue;
            }
```

The only remaining "gap" is if og of those widgets itself has widgets, but the FieldSet shouldn't be used for deeply nested fields anyway.

mjauvin:
@LukeTowers anything preventing this to get merged?

LukeTowers:
@mjauvin if it only supports `fields` under `form` then we should just perform that merge ourselves so that the definition only provides `fields.*` instead of `form.fields.*`. If you can make that change to the code and docs then I should be able to review and merge.

mjauvin:
> @mjauvin if it only supports `fields` under `form` then we should just perform that merge ourselves so that the definition only provides `fields.*` instead of `form.fields.*`. If you can make that change to the code and docs then I should be able to review and merge.

Why would it be different than with the nestedform widget?

LukeTowers:
@mjauvin because a nestedform does support tabs and the other config items that a form widget itself does; this one only supports fields.

mjauvin:
@LukeTowers I made the changes in here as well as in the docs

PR #1441, opened 2026-01-18
Add current form data as second argument to RecordFinder scope method

mjauvin:
It can be useful if the record finder needs some other fields in the form to filter its values.

Related: https://github.com/wintercms/docs/pull/256

E.g. fields definition:

```yaml
country:
    type: dropdown

state:
    type: recordfinder
    list: $/author/plugin/models/state/columns.yaml
    scope: statesByCountry
```

```php
class State extends model
{
    public function scopeStatesByCountry($query, $model, $formData)
    {
        if ($country_id = array_get($formData, 'country')) {
            $query->where('country_id', $country_id);
        }
    }
}
```

### DIFF
--- a/modules/backend/formwidgets/RecordFinder.php
+++ b/modules/backend/formwidgets/RecordFinder.php
@@ -353,7 +353,8 @@ class RecordFinder extends FormWidgetBase
         }
         elseif ($scopeMethod = $this->scope) {
             $widget->bindEvent('list.extendQueryBefore', function ($query) use ($scopeMethod) {
-                $query->$scopeMethod($this->model);
+                $formData = $this->getParentForm()->getSaveData();
+                $query->$scopeMethod($this->model, $formData);
             });
         }
         else {


AIC-BV:
Works perfectly! Thanks!!!
```php
public function scopeWhereProduct($query, $model, $formData)
{
    if ($productId = array_get($formData, 'product')) {
        return $query->where('product_id', $productId);
    }
    return $query->where('product_id', null); // returns an empty list in my case, so they will be forced to select a product first ;)
}
```

AIC-BV:
@mjauvin can this be merged?
I had trouble because I ran composer update

mjauvin:
@AIC-BV did you test this PR as well?

AIC-BV:
> @AIC-BV did you test this PR as well?

I copied the change to my local file and deployed it in production since Jan 18
Works perfectly

mjauvin:
@LukeTowers any objection in merging this? There's a PR to the docs also in the description.

JonasPardon:
@LukeTowers  Also running into this, could this be merged please? 🙏🏻

LukeTowers:
@mjauvin does this match any other scope method calls in the core? Seems somewhat arbitrary for your specific use case. @JonasPardon @AIC-BV can you provide examples of how you're using it?

AIC-BV:
@LukeTowers concrete case from our webshop plugin (running patched in production since January):

An `OrderProduct` has a `product` recordfinder and a `variant` recordfinder. A variant only exists within a product, so the variant list has to be filtered by the product that was just picked — in the same form, which isn't saved yet. `$this->model` alone can't do it: on a create form `product_id` is still null, so you either list every variant in the catalogue or none.

```yaml
product:
    type: recordfinder
    list: ~/plugins/aic/webshop/models/product/columns.yaml

variant:
    type: recordfinder
    list: ~/plugins/aic/webshop/models/variant/columns.yaml
    scope: whereProduct
```

```php
public function scopeWhereProduct($query, $model, $formData = [])
{
    $productId = array_get($formData, 'product') ?: $model->product_id;

    return $query->where('product_id', $productId); // null => empty list, forces picking a product first
}
```

Same shape as Marc's country → state: any parent → child recordfinder pair hits this. And since PHP ignores extra arguments on userland methods, existing single-argument scopes keep working unchanged.



In short, picking a product fills in the variant picker
image

PR #1447, opened 2026-02-05
Move Vite rendering to `styles` twig directive

damsfx:
If the recommendations are followed, the page architecture should be: 
```html
<!DOCTYPE html>
<html lang="en">
<head>
    /* Website metas */

    {% styles %}
</head>
<body>
    /* Website content */

    {% scripts %}
</body>
</html>
```
This avoids FOUC (Flash of Unstyled Content), the browser renders the content progressively, and the experience is better, scripts doesn't block content rendering and all nodes are availables for manipulation.

If a component uses $this->addVite(...) to include the required resources, those resources are rendered in the site's Twig tag {% scripts %}.
In this case, this can cause FOUC issues because the styles are rendered after the entire DOM.

Proposed fix: 
The idea is to render Vite's `<style />` and `<script />` elements from the `{% styles %}` twig directive: 
- No more FOUC
- No issues with scripts that could be blocking because the `type=“module”` attribute is added automatically and acts like the `defer` attribute.
 

## Release Notes

**Refactor**
* Modified asset loading sequences for stylesheets and scripts to streamline the resource compilation pipeline
* Adjusted asset processing order during page generation for improved resource management and efficiency

### DIFF
--- a/modules/cms/twig/ScriptsNode.php
+++ b/modules/cms/twig/ScriptsNode.php
@@ -26,7 +26,6 @@ class ScriptsNode extends TwigNode
         $compiler
             ->addDebugInfo($this)
             ->write("echo \$this->env->getExtension('Cms\Twig\Extension')->assetsFunction('js');\n")
-            ->write("echo \$this->env->getExtension('Cms\Twig\Extension')->assetsFunction('vite');\n")
             ->write("echo \$this->env->getExtension('Cms\Twig\Extension')->displayBlock('scripts');\n")
         ;
     }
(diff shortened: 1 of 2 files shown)


LukeTowers:
@damsfx could you also make the changes to how vite is rendered in the backend?

damsfx:
@LukeTowers I can't find any use of Vite in the backend module.   
Maybe I'm missing something, what are you talking about?

damsfx:
@LukeTowers ping

LukeTowers (approved):

PR #1465, opened 2026-03-03
Remove unnecessary closure argument passed to array_first()

Pasquale-CS:
The array_first function takes only one parameter: https://www.php.net/manual/en/function.array-first.php

## Summary by CodeRabbit

* **Refactor**
  * Internal code optimization to simplify permission redirect logic. No user-visible changes.


LukeTowers (approved):

LukeTowers:
This was an artifact from when array_first() mapped to Arr::first() and likely required the closure argument.

PR #1478, opened 2026-04-07
Fix infinite loop in form widget with circular dependsOn declarations

### LINKED ISSUE #421: dependsOn triggers infinite loop
### Winter CMS Build

1.1

### PHP Version

7.4

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

When a model has two fields that are dependent upon one another, an infinite loop occurs when the dependants are refreshed (line ~189 of winter.form.js: `$('[data-field-name="' + toRefresh.fields[0] + '"]').trigger('change')`

Appears to be related to this update: https://github.com/wintercms/winter/commit/712843d3c231432cefcc3fd5728518e43f73cd33#diff-cf4c08be53eb135693cb4e0424261e5d15c7475d92abc4c87aff32cb81266413

### Steps to replicate

Model setup:
```
alpha_field_one:
    type: switch
    dependsOn:
        - alpha_field_two
alpha_field_two:
    type: switch
    dependsOn:
        - alpha_field_one
```

### Workaround

_No response_

austinderrick:
## Summary

When two or more form fields have circular `dependsOn` declarations (e.g., field A depends on B and B depends on A), the `onRefreshDependants` method triggers change events that cascade infinitely, freezing the browser tab.

This fix adds cascade chain tracking through the jQuery event object. Each time a field triggers its dependants, the current field name is appended to a `cascadeChain` array on the event. Before refreshing, the method checks whether the triggering field already appears in the chain. If it does, a cycle has been detected and the cascade stops.

This preserves legitimate transitive cascading (A → B → C) while preventing circular loops (A → B → A) of any depth.

## How it works

The `event` parameter was already being passed by jQuery as the third argument to the handler (after the two preset arguments from `$.proxy`), so no changes to the event binding in `bindDependants` were needed. The only changes are in `onRefreshDependants`:

1. Accept the `event` parameter and extract `cascadeChain` (defaults to `[]` for user-initiated changes)
2. If `fieldName` is already in `cascadeChain`, return early (cycle detected)
3. In the success callback, append `fieldName` to the chain and pass it through the triggered change events

## Tests

Added 4 Jest tests covering:
- Basic dependency refresh (A → B)
- Circular dependency detection (A → B → A, stops at 2 requests)
- Transitive cascading preservation (A → B → C, both refresh correctly)
- Three-field circular chain detection (A → B → C → A, stops at 3 requests)

All 44 tests pass (7 suites including the new one).

```
PASS tests/js/cases/framework/FormWidgetDependants.test.js
  Form Widget dependsOn
    ✓ refreshes dependent fields when a field changes (556 ms)
    ✓ prevents infinite loop with circular dependsOn declarations (2015 ms)
    ✓ allows transitive cascading (A -> B -> C) without blocking (1514 ms)
    ✓ stops cycle in a three-field circular chain (A -> B -> C -> A) (2016 ms)

Test Suites: 7 passed, 7 total
Tests:       44 passed, 44 total
```

Fixes #421

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented infinite refresh loops for dependent form fields by tracking the refresh chain and detecting circular dependencies before triggering further updates.

* **Tests**
  * Added Jest tests covering dependency cascades: circular prevention, transitive chains, and request counts/timing.
  * Added test fixtures/stubs to simulate the widget environment so tests run reliably in isolation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wintercms/winter/pull/1478?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/modules/system/tests/js/cases/framework/FormWidgetDependants.test.js
+++ b/modules/system/tests/js/cases/framework/FormWidgetDependants.test.js
@@ -6,7 +6,7 @@ describe('Form Widget dependsOn', function () {
     /**
      * Build a FakeDom with jQuery, WinterCMS foundation, stubs, and the form widget.
      * The FormWidgetStubs fixture provides minimal implementations of ocJSON,
-     * $.fn.render, $.fn.request (synchronous success), and $.fn.loadIndicator.
+     * $.fn.render, $.fn.request (synchronous success), and $.fn.loadIndicator
      */
     function buildDom(html) {
         return FakeDom
--- a/modules/system/tests/js/cases/framework/FormWidgetDependants.test.js
+++ b/modules/system/tests/js/cases/framework/FormWidgetDependants.test.js
@@ -1,0 +1,149 @@
+import FakeDom from '../../helpers/FakeDom';
+
+jest.setTimeout(5000);
+
+describe('Form Widget dependsOn', function () {
+    /**
+     * Build a FakeDom with jQuery, WinterCMS foundation, stubs, and the form widget.
+     * The FormWidgetStubs fixture provides minimal implementations of ocJSON,
+     * $.fn.render, $.fn.request (synchronous success), and $.fn.loadIndicator.
+     */
+    function buildDom(html) {
+        return FakeDom
+            .new()
+            .addScript([
+                'modules/backend/assets/js/vendor/jquery.min.js',
+                'modules/system/assets/ui/js/foundation.baseclass.js',
+                'modules/system/assets/ui/js/foundation.controlutils.js',
+                'modules/system/tests/js/fixtures/formWidget/FormWidgetStubs.js',
+                'modules/backend/widgets/form/assets/js/winter.form.js',
+            ])
+            .render(html);
+    }
+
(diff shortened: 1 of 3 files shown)


modules/system/tests/js/cases/framework/FormWidgetDependants.test.js
```diff
@@ -0,0 +1,149 @@
+import FakeDom from '../../helpers/FakeDom';
+
+jest.setTimeout(5000);
+
+describe('Form Widget dependsOn', function () {
+    /**
+     * Build a FakeDom with jQuery, WinterCMS foundation, stubs, and the form widget.
+     * The FormWidgetStubs fixture provides minimal implementations of ocJSON,
+     * $.fn.render, $.fn.request (synchronous success), and $.fn.loadIndicator.
```
LukeTowers:
```suggestion
     * $.fn.render, $.fn.request (synchronous success), and $.fn.loadIndicator
```

PR #1481, opened 2026-05-06
Added support for returning vite assets in their correct group

jaxwilko:
This PR adds support for returning vite assets in their correct groups.

This is done so that CSS/JS added via `->addVite()` is correctly rendered via the `{% styles %}` & `{% scripts %}` tags.

The `getAssetType()` method is a little fluffy as it will probably be extended later.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced asset handling with tighter Vite integration: CSS and JS entrypoints are correctly grouped and emitted to prevent cross-leakage between styles and scripts; styles now emit the appropriate CSS entrypoints alongside Vite-managed assets.

* **Tests**
  * Expanded test coverage for asset handling and Vite integration to verify CSS/JS emission, grouping behavior, and to guard against regressions.

### DIFF
--- a/modules/system/tests/traits/AssetMakerTest.php
+++ b/modules/system/tests/traits/AssetMakerTest.php
@@ -2,10 +2,12 @@
 
 namespace System\Tests\Traits;
 
+use System\Classes\Asset\PackageManager;
 use System\Tests\Bootstrap\TestCase;
 use System\Traits\AssetMaker;
 use System\Traits\EventEmitter;
 use System\Traits\ViewMaker;
+use Winter\Storm\Support\Facades\File;
 use Winter\Storm\Support\Facades\Url;
 
 class AssetMakerStub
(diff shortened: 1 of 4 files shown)


modules/system/tests/traits/AssetMakerTest.php
```diff
@@ -118,4 +172,116 @@ public function testAssetOrdering(): void
...
+            'foo.js' => 'js',
+            'assets/javascript/theme.js' => 'js',
+            'foo.css' => 'css',
+            'assets/css/theme.css' => 'css',
+            'foo.txt' => null,
+            'foo' => null,
+            'js' => null,
+            // str_ends_with is case-sensitive; the helper only matches lowercase extensions.
+            'foo.JS' => null,
+            'foo.CSS' => null,
```
jaxwilko:
```suggestion
            'foo.JS' => 'js',
            'foo.CSS' => 'css',
```

damsfx:
It works as expected for me. 👍

LukeTowers (approved):

PR #1482, opened 2026-05-06
Added basic viewer for event log details

### LINKED ISSUE #1475: Contextual data array not displayed in backend Event Log despite documentation
### Winter CMS Build

1.2

### PHP Version

8.1

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Issue description

According to the [official Winter CMS documentation](https://wintercms.com/docs/v1.2/docs/services/error-log), passing an array of contextual data to the log methods should be "formatted and displayed with the log message":

php
Log::info('User failed to login.', ['id' => $user->id]);

However, when viewing the log entry in the backend System → Logs → Event Log, only the message string is visible. The contextual array does not appear anywhere in the backend interface.

### Steps to replicate

Add the following code anywhere in a plugin or theme:

php
Log::info('Test contextual log', ['status' => 200, 'body' => 'Test response']);
Trigger the code (e.g., visit a page or run a component)

Go to Backend → System → Logs → Event Log

Find the newly created log entry and click to view it

Observe that only "Test contextual log" is visible — no status or body data

### Workaround

_No response_

jaxwilko:
This PR adds a very basic viewer for event log details.

image

Should resolve https://github.com/wintercms/winter/issues/1475.

I'd like to add better support for the new log viewer, but don't have time currently. To anybody reading this you can use:
```php
$message = 'My Error Message';
$details = (new EventLog())->getDetails(new \Exception());
$details['exception']['message'] = $message;
$details['exception']['snippet'] = [
    'Something terrible has happened',
];

Log::error($message, $details);
```
To force in custom logs in the new style with fancy stack traces, but use at your own risk because it's fairly easy to break the renderer while doing this.

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Event Logs field details now include a visible "DETAILS" block displaying the field value when available, providing more contextual information directly in the Event Logs view. No other behavior was changed.

### DIFF
--- a/modules/system/controllers/eventlogs/_field_details.php
+++ b/modules/system/controllers/eventlogs/_field_details.php
@@ -1,5 +1,11 @@
 <?php
 if (!isset($value['logVersion']) || $value['logVersion'] !== 2) {
+    if ($value ?? false) {
+        printf(
+            '<h4>DETAILS</h4><pre style="background: #fff; padding: 10px; border: 1px solid #ddd;">%s</pre>',
+            print_r($value, true)
+        );
+    }
     return;
 }
 
--- a/modules/system/controllers/eventlogs/_field_details.php
+++ b/modules/system/controllers/eventlogs/_field_details.php
@@ -3,7 +3,7 @@ if (!isset($value['logVersion']) || $value['logVersion'] !== 2) {
     if ($value ?? false) {
         printf(
             '<h4>DETAILS</h4><pre style="background: #fff; padding: 10px; border: 1px solid #ddd;">%s</pre>',
-            print_r($value, true)
+            e(print_r($value, true))
         );
     }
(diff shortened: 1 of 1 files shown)


modules/system/controllers/eventlogs/_field_details.php
```diff
@@ -1,5 +1,11 @@
 <?php
 if (!isset($value['logVersion']) || $value['logVersion'] !== 2) {
+    if ($value ?? false) {
+        printf(
+            '<h4>DETAILS</h4><pre style="background: #fff; padding: 10px; border: 1px solid #ddd;">%s</pre>',
+            print_r($value, true)
```
jaxwilko:
```suggestion
            e(print_r($value, true))
```

PR #1490, opened 2026-06-12
Fix deprecation notices in the backend module

austinderrick:
## Summary

Fixes two families of deprecation notices in the backend module, found while auditing PHP 8.4/8.5 readiness (follow-up to #1489):

**1. `TranscodeFilter` is missing the return types declared by `php_user_filter`.** Loading the class raises three deprecations on PHP 8.1+:

```
Deprecated: Return type of ...TranscodeFilter::filter($in, $out, &$consumed, $closing) should either be compatible with php_user_filter::filter($in, $out, &$consumed, bool $closing): int, or the #[\ReturnTypeWillChange] attribute should be used...
Deprecated: Return type of ...TranscodeFilter::onCreate() should either be compatible with php_user_filter::onCreate(): bool, ...
Deprecated: Return type of ...TranscodeFilter::onClose() should either be compatible with php_user_filter::onClose(): void, ...
```

Since the repo's PHP floor is 8.1, this adds the real return types (`: int`, `: bool`, `: void`) rather than `#[\ReturnTypeWillChange]`. The existing bodies already return the right types (`PSFS_PASS_ON`, booleans, nothing).

**2. `displayAs()` passes a null render mode into `strtolower()`.** `Lists::makeListColumn()` and `Filter::makeFilterScope()` pass `$config['type'] ?? null` into `displayAs()`, and `FormField`, `ListColumn`, and `FilterScope` each do `strtolower($type) ?: $this->type`, so any column/scope/field config without an explicit `type` raises:

```
Deprecated: strtolower(): Passing null to parameter #1 ($string) of type string is deprecated
```

Fixed by typing the parameter natively (`?string $type`) in all three classes and coalescing inside (`strtolower($type ?? '')`), keeping identical semantics: null and `''` both fall back to the existing type. Real argument types rather than a docblock-only change, per the direction from the storm#229 review. All in-repo callers already pass string-or-null (`Form::makeFormField()` validates the type before calling), and untyped subclass overrides remain legal (widening).

## Verification

Reproduced all four deprecations on a clean `develop` checkout under PHP 8.5.5 with a small harness (load `TranscodeFilter`, call each `displayAs(null)`); after this change the same harness emits none. `modules/backend` PHPUnit suite shows an identical result before and after the change on the same machine (the handful of locally-failing tests are environment-related and untouched by this diff: FileUploadScoping, NavigationManager, WidgetManager).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented missing “type” configs from causing rendering issues by defaulting to sensible values (e.g., text for fields/columns and group for filter scopes).
* **Refactor**
  * Strengthened type-safety across backend display helpers by requiring non-nullable string inputs.
  * Added explicit return/parameter typing for filter lifecycle methods to make behavior more predictable.

### DIFF
--- a/modules/backend/behaviors/importexportcontroller/TranscodeFilter.php
+++ b/modules/backend/behaviors/importexportcontroller/TranscodeFilter.php
@@ -17,7 +17,7 @@ class TranscodeFilter extends php_user_filter
 
     protected $encodingTo;
 
-    public function filter($in, $out, &$consumed, $closing)
+    public function filter($in, $out, &$consumed, bool $closing): int
     {
         while ($resource = stream_bucket_make_writeable($in)) {
             if (in_array($this->encodingFrom, mb_list_encodings())) {
@@ -42,7 +42,7 @@ class TranscodeFilter extends php_user_filter
         return PSFS_PASS_ON;
     }
 
-    public function onCreate()
+    public function onCreate(): bool
     {
         if (strpos($this->filtername, self::FILTER_NAME) !== 0) {
             return false;
@@ -70,7 +70,7 @@ class TranscodeFilter extends php_user_filter
         return true;
     }
 
-    public function onClose()
+    public function onClose(): void
     {
         setlocale(LC_CTYPE, $this->params['locale']);
     }
--- a/modules/backend/classes/FilterScope.php
+++ b/modules/backend/classes/FilterScope.php
@@ -101,12 +101,12 @@ class FilterScope
      * Specifies a scope control rendering mode. Supported modes are:
      * - group - filter by a group of IDs. Default.
      * - checkbox - filter by a simple toggle switch.
-     * @param string $type Specifies a render mode as described above
+     * @param string|null $type Specifies a render mode as described above
      * @param array $config A list of render mode specific config.
      */
-    public function displayAs($type, $config = [])
+    public function displayAs(?string $type, $config = [])
     {
-        $this->type = strtolower($type) ?: $this->type;
+        $this->type = strtolower($type ?? '') ?: $this->type;
         $this->config = $this->evalConfig($config);
         return $this;
     }
--- a/modules/backend/classes/FilterScope.php
+++ b/modules/backend/classes/FilterScope.php
@@ -101,12 +101,12 @@ class FilterScope
      * Specifies a scope control rendering mode. Supported modes are:
      * - group - filter by a group of IDs. Default.
      * - checkbox - filter by a simple toggle switch.
-     * @param string|null $type Specifies a render mode as described above
+     * @param string $type Specifies a render mode as described above
      * @param array $config A list of render mode specific config.
      */
-    public function displayAs(?string $type, $config = [])
+    public function displayAs(string $type, $config = [])
     {
-        $this->type = strtolower($type ?? '') ?: $this->type;
+        $this->type = strtolower($type) ?: $this->type;
         $this->config = $this->evalConfig($config);
         return $this;
     }
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -711,7 +711,7 @@ class Filter extends WidgetBase
     protected function makeFilterScope($name, $config)
     {
         $label = $config['label'] ?? null;
-        $scopeType = $config['type'] ?? null;
+        $scopeType = $config['type'] ?? 'group';
 
         $scope = new FilterScope($name, $label);
         $scope->displayAs($scopeType, $config);
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -850,8 +850,8 @@ class Form extends WidgetBase
          * Defined field type
          */
         else {
-            $fieldType = $config['type'] ?? null;
-            if (!is_string($fieldType) && $fieldType !== null) {
+            $fieldType = $config['type'] ?? 'text';
+            if (!is_string($fieldType)) {
                 throw new ApplicationException(Lang::get(
                     'backend::lang.field.invalid_type',
                     ['type' => gettype($fieldType)]
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -1111,7 +1111,7 @@ class Lists extends WidgetBase
             $config['searchable'] = false;
         }
 
-        $columnType = $config['type'] ?? null;
+        $columnType = $config['type'] ?? 'text';
 
         $column = new ListColumn($name, $label);
         $column->displayAs($columnType, $config);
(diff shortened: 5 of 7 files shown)


modules/backend/behaviors/importexportcontroller/TranscodeFilter.php
```diff
@@ -17,7 +17,7 @@ class TranscodeFilter extends php_user_filter
 
     protected $encodingTo;
 
-    public function filter($in, $out, &$consumed, $closing)
+    public function filter($in, $out, &$consumed, $closing): int
```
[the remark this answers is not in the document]
austinderrick: Applied in b545a05 — `bool $closing` now matches the parent declaration exactly. One precision on the rationale: the type isn't required to clear the deprecation (omitting a parameter type in an override is legal widening, and the tentative-return-type check only concerns the return — verified empirically with an E_ALL harness before adding it). It's included as signature hygiene to mirror `php_user_filter::filter()` verbatim.

modules/backend/classes/FilterScope.php
```diff
@@ -101,12 +101,12 @@ public function __construct($scopeName, $label)
...
      * - checkbox - filter by a simple toggle switch.
-     * @param string $type Specifies a render mode as described above
+     * @param string|null $type Specifies a render mode as described above
      * @param array $config A list of render mode specific config.
      */
-    public function displayAs($type, $config = [])
+    public function displayAs(?string $type, $config = [])
     {
-        $this->type = strtolower($type) ?: $this->type;
+        $this->type = strtolower($type ?? '') ?: $this->type;
```
LukeTowers: Why would this be allowed to be called without an argument for type present?

LukeTowers (changes requested):

modules/backend/classes/FilterScope.php
austinderrick: The null comes from the config-driven builders. When a field, column, or scope whose config omits type calls displayAs(null), the method treats that as "no override, keep the default render mode".

So there are two ways we can go:
1) We can keep accepting null and make that contract explicit. That stays backward compatible, since displayAs is public API and a plugin may pass a null type.

2) Or we can reject null: default the type at the three callers (config['type'] ?? 'group', and so on) and make displayAs require a real string. That reads cleaner, but it breaks anyone calling displayAs directly.

Which do you prefer?

LukeTowers: Anybody calling `displayAs(null)` deserves to have their code break 😛 Display as what? Hopes and dreams? 😂

Let's go with option 2 @austinderrick; thanks!

austinderrick: Haha, sounds great! I'll go that route. :)

austinderrick: Updated!

modules/backend/widgets/Filter.php
```diff
@@ -711,7 +711,7 @@ public function removeScope($scopeName)
     protected function makeFilterScope($name, $config)
     {
         $label = $config['label'] ?? null;
-        $scopeType = $config['type'] ?? null;
+        $scopeType = $config['type'] ?? 'group';
```
LukeTowers, line 714: Where was this default being inferred previously?

modules/backend/widgets/Form.php
```diff
@@ -850,8 +850,8 @@ protected function makeFormField($name, $config = [])
          * Defined field type
          */
         else {
-            $fieldType = $config['type'] ?? null;
-            if (!is_string($fieldType) && $fieldType !== null) {
+            $fieldType = $config['type'] ?? 'text';
```
LukeTowers, line 853: Where was this default being inferred previously?

modules/backend/widgets/Lists.php
```diff
@@ -1111,7 +1111,7 @@ protected function makeListColumn($name, $config)
             $config['searchable'] = false;
         }
 
-        $columnType = $config['type'] ?? null;
+        $columnType = $config['type'] ?? 'text';
```
LukeTowers, line 1114: Where was this default being inferred previously?

modules/backend/classes/FilterScope.php
```diff
@@ -104,7 +104,7 @@ public function __construct($scopeName, $label)
      * @param string $type Specifies a render mode as described above
      * @param array $config A list of render mode specific config.
      */
-    public function displayAs($type, $config = [])
+    public function displayAs(string $type, $config = [])
     {
         $this->type = strtolower($type) ?: $this->type;
```
LukeTowers, line 109: Does this only fallback to $this->type now when calling `displayAs('')`?

modules/backend/widgets/Filter.php
austinderrick: It was inferred inside `FilterScope::displayAs()`. Previously `makeFilterScope()` passed `$config['type'] ?? null`, and `displayAs()` does `$this->type = strtolower($type) ?: $this->type`. With an absent type, `strtolower(null)` returns `''` (the deprecation we're fixing), then `'' ?: $this->type` falls through to the `FilterScope::$type` property default, which is `'group'` (`FilterScope.php:47`).

So `?? 'group'` just hoists that same property default up to the caller. The resolved scope type is identical; the only change is that a real string now reaches `displayAs()`, so the `strtolower(null)` deprecation never fires.

modules/backend/widgets/Form.php
austinderrick: Same mechanism: `FormField::$type` defaults to `'text'` (`FormField.php:81`). Previously `makeFormField()` passed `$config['type'] ?? null` into `FormField::displayAs()`, which does `$this->type = strtolower($type) ?: $this->type` — so `null` became `strtolower(null)` → `''` → fall back to `'text'`. `?? 'text'` resolves to that same value while keeping a real string in the signature.

I also dropped the now-dead `&& $fieldType !== null` from the guard on the next line, since `$fieldType` can no longer be `null` once it defaults to `'text'`.

modules/backend/widgets/Lists.php
austinderrick: Same as the other two: `ListColumn::$type` defaults to `'text'` (`ListColumn.php:28`), and `ListColumn::displayAs()` does `$this->type = strtolower($type) ?: $this->type`. An absent type previously fell back to `'text'` via `strtolower(null)` → `''` → `$this->type`. `?? 'text'` makes that default explicit at the caller without changing the resolved column type.

modules/backend/classes/FilterScope.php
austinderrick: Yes. Now that `$type` is typed `string`, `null` can't reach it, so `strtolower($type) ?: $this->type` only falls back to `$this->type` when `strtolower($type)` is falsy — i.e. `displayAs('')` (and the pathological `displayAs('0')`, which was already falsy here before this change). Any non-empty type string is applied as-is.

With the three in-repo callers now always passing a non-empty default, that fallback is effectively only reachable by an external/plugin caller that explicitly passes `''`.

PR #1499, opened 2026-07-06
RelationController behavior: support for deeply nested relations

mjauvin:
We can now use this in config_relation.yaml:

```yaml
items:
    label: Items
    view:
        toolbarButtons: create|delete
        list: $/acme/nested/models/item/columns.yaml
    manage:
        form: $/acme/nested/models/item/fields.yaml

items[taxes]:
    label: taxes
    view:
        toolbarButtons: create|delete
        list: $/acme/nested/models/tax/columns.yaml
    manage:
        form: $/acme/nested/models/tax/fields.yaml
```
Which would create a popup to manage Items in an Order, along with another nested popup to manage Taxes within that Items popup (see screenshot):

Note: deeper nesting can be achieved as well (e.g. items[taxes][other])

image

The nested popups now have an offset which makes it easier to see the nesting.

## Summary

- **New Features**
  - Nested relation fields specified with bracket paths (multi-level/repeated) are supported more reliably.
  - Multiple simultaneously open popups now cascade with a small visual offset to reduce overlap.

- **Bug Fixes**
  - Relation actions (create/update/remove/unlink/delete) now target the correct nested record and stay isolated from stale request context.
  - Popup callbacks consistently report the intended relation field name, improving nested UI routing and preventing DOM/session collisions.

- **Tests**
  - Added a comprehensive nested relation controller test suite covering recursive, ambient, and regression scenarios.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -536,8 +536,13 @@ class RelationController extends ControllerBehavior
      *  a) AMBIENT - the nesting stack is non-empty, meaning a manage
      *     form is actively rendering right now (see
      *     relationMakePartial()) and this field is nested directly
-     *     under whatever's on top of it. $model is trusted as-is - it's
-     *     the widget's own bound model.
+     *     under whatever's on top of it. $model canNOT be trusted
+     *     here: initRelation() is reached as
+     *     relationRender() -> validateField() -> initRelation($this->model),
+     *     and $this->model is the record of the ENCLOSING relation,
+     *     not of the manage form actually rendering. Those coincide at
+     *     one level of nesting and diverge at two, so the parent is
+     *     walked from the root exactly as in (b).
      *  b) RECONSTRUCTED - $field is already a bracket path (e.g.
      *     "items[5][taxes]"), walked from the root to find the actual
      *     model. This is the common case for anything other than an
@@ -574,7 +579,9 @@ class RelationController extends ControllerBehavior
 
         if ($ambientPath !== null) {
             $this->nestedField = $ambientPath . '[' . $field . ']';
-            return [$model, $field, true];
+            [$hops] = $this->parseBracketField($this->nestedField);
+
+            return [$this->resolveModelForHops($this->rootModel, $hops), $field, true];
         }
 
         if (strpos($field, '[') !== false) {
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -666,10 +666,21 @@ class RelationController extends ControllerBehavior
                 );
             }
 
-            $related = $model->{$relationName}()->getRelated();
+            $relation = $model->{$relationName}();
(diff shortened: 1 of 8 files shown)


AIC-BV:
AI dump, hopefuly this makes any sense to you

Tested this branch against a real three-level config in our own backend (`Order → closets → paslatten → closet`). Two levels work well — correct parent scoping, sibling isolation, DOM/session isolation via the hashed alias, and the popup offset all behaved. Three levels hit one bug.

**Symptom**

A form yaml rendered as a nested manage form, whose own partial calls `relationRender('closet')`, throws:

```
BadMethodCallException: Call to undefined method Aic\Account\Models\Closet::closet()

#6 RelationController.php(450): Model->__call('closet', Array)
#7 RelationController.php(304): RelationController->initRelation(Object(Closet), 'closet')
#8 RelationController.php(931): RelationController->validateField('closet')
#9 RelationController->relationRender('closet')
```

`closet` is a relation on `Paslatten`, not on `Closet` — the leaf resolved against the grandparent.

**Cause**

The ambient branch in `resolveNestedContext()` trusts `$model`, per its docblock ("it's the widget's own bound model"). But the call arrives as `relationRender() → validateField() → initRelation($this->model)`, and `$this->model` is the record of the *enclosing* relation. At one level of nesting that is the parent, so it works; at two it is the grandparent.

**Patch**

```diff
 if ($ambientPath !== null) {
     $this->nestedField = $ambientPath . '[' . $field . ']';
-    return [$model, $field, true];
+    [$hops] = $this->parseBracketField($this->nestedField);
+
+    return [$this->resolveModelForHops($this->rootModel, $hops), $field, true];
 }
```

This is what the reconstructed branch two lines below already does, so both cases share one resolution mechanism and the ambient path keeps `resolveModelForHops()`'s `$relation->find($id)` parent check at every depth.

All 22 existing tests still pass unmodified — depth 2 resolves to the same model either way, which is also why the suite doesn't currently catch this. I added `testAmbientFieldTwoLevelsDeepResolvesInnermostAncestorNotEnclosingRecord` using your self-referencing `Item → items → Item` fixture; it fails without the patch (resolves Item A instead of the sub-item) and passes with it. Full backend suite diffed both ways shows no other change.

After the patch, three levels resolve correctly in the browser: closet A's paslat renders closet A, closet B's renders closet B.

Happy to open this as a PR against `nested-relation-controller` if useful.

Two smaller notes:

1. The branch is diverged rather than just ahead (12 commits behind `develop` on backend, 7 on system), so installing it removes the FieldSet formwidget, the Twig SecurityPolicy hardening and a few tests. Worth merging `develop` to make it testable against a current tree.
2. `aliasNestedConfig()` throws a hard `ApplicationException` when a nested path has no config entry. That means any shared form yaml calling `relationRender()` breaks the popup unless every host controller re-declares the whole chain. Declaring it is reasonable, but the failure mode is harsh — left alone here since it looks like a design call for you.

PR #1503, opened 2026-07-25
Move backend user authorization guards to the actual write events

### LINKED ISSUE #1464: RelationController saves related model when only pivot data changes, triggering User authorization in BelongsToMany relations
### Winter CMS Build

1.2.12

### PHP Version

8.3

### Database engine

MySQL/MariaDB

### Plugins installed

_No response_

### Summary
When editing pivot data on a BelongsToMany relation where the related model is `Backend\Models\User`, the RelationController saves the related User model even when only pivot fields change. This triggers the new `beforeSave()` authorization checks added in 1.2.12 and causes an `AuthorizationException` for users without `backend.manage_users` permission, even though they are only updating pivot data, not the User record itself.

### Description
In 1.2.12, `Backend\Models\User` gained a `beforeSave()` hook that enforces authorization rules to prevent privilege escalation. One rule requires` backend.manage_users` when saving another user’s record.
For BelongsToMany relations where the related model is `Backend\Models\User`, editing only pivot data (e.g. `pivot[is_default]`) causes the RelationController to save the related User model. That save triggers `User::beforeSave()`, which then throws an `AuthorizationException` for users `without backend.manage_users`, even though no User attributes are being changed.

### Steps to replicate

1. Create a BelongsToMany relation where the related model is `Backend\Models\User`.
2. Configure a pivot form with at least one pivot field (e.g. `pivot[is_default]`).
3. Log in as a backend user who does not have `backend.manage_users`.
4. Open a record that uses this relation.
5. Add another backend user to the relation.
6. Edit the pivot data for that user (e.g. toggle `is_default`).
7. Submit the pivot form.

### Expected behavior
Only the pivot record should be updated. The related User model should not be saved when only pivot fields change, since no User attributes are modified.

### Actual behavior
The RelationController calls `prepareModelsToSave()` with the related model (User). `FormModelSaver::setModelAttributes()` always adds the main model to `modelsToSave` (line 55), so `User::save()` is called even when only pivot data changes. This triggers `User::beforeSave()`, which throws:
```
You do not have permission to manage other administrators.
```

### Root cause
In `RelationController::onRelationManagePivotUpdate()` (and `onRelationManagePivotCreate()`):
```php
$modelsToSave = $this->prepareModelsToSave($hydratedModel, $saveData);
foreach ($modelsToSave as $modelToSave) {
    $modelToSave->save(null, $this->pivotWidget->getSessionKey());
}
```
`prepareModelsToSave()` uses `FormModelSaver::setModelAttributes()`, which always pushes the main model into `modelsToSave` before processing nested attributes. For BelongsToMany relations, the main model is the related model (User). As a result, the User is saved even when the form only contains pivot data.

### Impact
- Users without `backend.manage_users` cannot edit pivot data for relations that use `Backend\Models\User` as the related model.
- This affects any plugin or module that uses BelongsToMany relations to `Backend\Models\User` with editable pivot data.
- Actually bad workaround: grant `backend.manage_users` to these users, which broadens their permissions beyond what is needed for pivot editing.

### Suggested fix
For BelongsToMany (and similar) relations when handling pivot datas, the RelationController should only save models whose attributes actually changed. In particular:
- If `$saveData` only contains `pivot` (or equivalent pivot keys), only the pivot record should be saved.
- The related model (e.g. User) should not be saved when no non-pivot attributes are present in `$saveData`.

### Additional context
- This worked in 1.2.9 because `User` did not have the `beforeSave()` authorization checks.
- The new security rules in `User::beforeSave()` are appropriate; the issue is that the RelationController saves the related model unnecessarily when only pivot data changes.
- Related files:
  - `modules/backend/behaviors/RelationController.php` (e.g. `onRelationManagePivotUpdate`, `onRelationManagePivotCreate`)
  - `modules/backend/traits/FormModelSaver.php` (`prepareModelsToSave`, `setModelAttributes`)
  - `modules/backend/models/User.php` (`beforeSave`)

LukeTowers:
Fixes #1464

## Problem

`RelationController`'s pivot handlers save every model returned by `prepareModelsToSave()`. That list always includes the related model — `FormModelSaver::setModelAttributes()` pushes it before it even looks at `$saveData` — so a submission containing nothing but pivot data still saves the related record.

For a `belongsToMany` relation to `Backend\Models\User`, that no-op save fired `User::beforeSave()`, which throws for any operator lacking `backend.manage_users`. The relation became unusable for them. This is a regression against 1.2.9, surfaced by the user authorization guards added since.

## Fix

An earlier revision of this PR taught the pivot handlers to skip models with nothing to write. Review showed that approach is inherently brittle — the save pipeline has several independent write channels (attribute updates, deferred binding commits, relation syncs queued by `setSimpleValue()`, purgeable attributes that look dirty but aren't), and any skip heuristic must re-derive all of them. Two gaps were confirmed: relation-widget fields on the related model were silently dropped, and `_`-prefixed helper fields defeated the `isDirty()` check.

This revision moves the guard instead, onto the events that only fire for **actual writes**:

- **`beforeCreate()` / `beforeUpdate()`** guard attribute writes. Eloquent's update event only fires when attributes have genuinely changed — after purgeable attributes are stripped — so a save with nothing to write requires no permission.
- **`model.relation.beforeAttach` / `beforeDetach` / `beforeAdd` / `beforeRemove`** guard relation writes. Direct `attach()`/`detach()`, deferred binding commits (`commitDeferredAfter` → `add()` → `attach()`) and queued `setSimpleValue()` syncs (afterSave → `sync()` → `attach()`/`detach()`) all funnel through these events, so the guard fires at the moment the relation write happens rather than trying to predict it.

The pivot handlers save every prepared model unconditionally again — restoring pre-1.2.10 behavior for plugin `afterSave` hooks and relation-widget fields — and `onRelationManagePivotUpdate()` is wrapped in `Db::transaction()` to match `onRelationManagePivotCreate()`.

### Guard semantics

- Attribute changes: unchanged rules (self non-escalating edits allowed; others' records need `backend.manage_users`; superuser targets need a superuser actor; CLI/queue contexts exempt).
- Relation changes are guarded only for the relations the user model **owns** — `groups`, `avatar`, `throttle`, listed in the public `$permissionGuardedRelations` property. Changing them on **another user's** record requires `backend.manage_users`, now even via direct `attach()`/`detach()` — those paths previously bypassed the save-based guard entirely.
- Plugin-added relations on the user model are deliberately **not** guarded: a plugin writing its own user-side relation (e.g. assigning a user to a support ticket via `$user->tickets()->add()`) keeps working for unprivileged operators, since that relation is the plugin's domain. Plugins with genuinely sensitive relations can append to `$permissionGuardedRelations` or bind their own guard to the same `model.relation.*` events. (Relations that merely reference users from the other side — `$ticket->users()->attach()` — never touched the guard, since the events fire on the parent model.)
- Relation changes to **your own** record are allowed: backend groups carry no permissions out of the box (`getMergedPermissions()` merges only role + user permissions), so group membership is not an escalation vector.

## Tests

`RelationControllerPivotTest` (10 cases) exercises the production pivot save sequence end-to-end: pivot-only saves with and without the permission, related-field edits (applied and refused), relation-widget fields via the `setSimpleValue()` path (applied and refused), and deferred binding commits (applied and refused). `UserAuthorizationTest` grows from 31 to 39 cases: no-op saves need no permission, creates are guarded, group attach/detach/sync are guarded for other users' records, self group changes stay allowed, and a plugin-style relation added to the user model is not guarded.

Negative control: with the `model.relation.*` bindings commented out, all six unauthorized-relation-change tests fail — the coverage bites.

Full core suite passes (backend: 169 tests / 371 assertions; the only failures elsewhere are 3 pre-existing environmental failures in `cms/ControllerTest` caused by a local asset-URL config, present on the unmodified branch too). phpcs clean on all changed files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved many-to-many pivot persistence by ensuring related model prep and saves run atomically.
  - Pivot updates now reliably handle authorization: pivot-only changes succeed without requiring permission to edit the related record, while related record field edits remain protected.
  - Deferred relation bindings are committed during pivot updates when authorized.

- **New Features**
  - Stricter user update authorization, including guarded relation changes (e.g., group membership), with clearer denial behavior.

- **Tests**
  - Added pivot regression coverage and expanded user/group authorization scenarios.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1569,10 +1569,9 @@ class RelationController extends ControllerBehavior
             $hydratedModels = $this->relationObject->whereIn($foreignKeyName, $foreignIds)->get();
 
             foreach ($hydratedModels as $hydratedModel) {
-                $modelsToSave = $this->prepareModelsToSave($hydratedModel, $saveData);
-                foreach ($modelsToSave as $modelToSave) {
-                    $modelToSave->save(null, $this->pivotWidget->getSessionKey());
-                }
+                $this->relationSavePivotModels(
+                    $this->prepareModelsToSave($hydratedModel, $saveData)
+                );
             }
         });
 
@@ -1583,14 +1582,17 @@ class RelationController extends ControllerBehavior
     {
         $this->beforeAjax();
 
-        $foreignKeyName = $this->relationModel->getQualifiedKeyName();
         $hydratedModel = $this->pivotWidget->model;
         $saveData = $this->pivotWidget->getSaveData();
 
-        $modelsToSave = $this->prepareModelsToSave($hydratedModel, $saveData);
-        foreach ($modelsToSave as $modelToSave) {
-            $modelToSave->save(null, $this->pivotWidget->getSessionKey());
-        }
+        /*
+         * If any of the models fail to save, abort the whole update
+         */
+        Db::transaction(function () use ($hydratedModel, $saveData) {
+            $this->relationSavePivotModels(
+                $this->prepareModelsToSave($hydratedModel, $saveData)
+            );
+        });
 
         return ['#'.$this->relationGetId('view') => $this->relationRenderView()];
     }
@@ -1676,6 +1678,29 @@ class RelationController extends ControllerBehavior
     //
     // Helpers
     //
+
+    /**
+     * Saves the models prepared from a pivot form submission.
+     *
+     * Models that already exist and have no changed attributes are skipped. `prepareModelsToSave()`
+     * always queues the related model, so a submission containing nothing but pivot data would
+     * otherwise trigger that model's save events - including authorization guards such as
+     * `Backend\Models\User::beforeSave()`. Their deferred bindings are still committed so that
+     * relation widgets on the pivot form continue to work.
+     */
+    protected function relationSavePivotModels(array $modelsToSave): void
+    {
+        $sessionKey = $this->pivotWidget->getSessionKey();
+
+        foreach ($modelsToSave as $modelToSave) {
+            if ($modelToSave->exists && !$modelToSave->isDirty()) {
+                $modelToSave->commitDeferred($sessionKey);
+                continue;
+            }
+
+            $modelToSave->save(null, $sessionKey);
+        }
+    }
 
     /**
      * Returns the existing record IDs for the relation.
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1569,9 +1569,10 @@ class RelationController extends ControllerBehavior
             $hydratedModels = $this->relationObject->whereIn($foreignKeyName, $foreignIds)->get();
 
             foreach ($hydratedModels as $hydratedModel) {
-                $this->relationSavePivotModels(
-                    $this->prepareModelsToSave($hydratedModel, $saveData)
-                );
+                $modelsToSave = $this->prepareModelsToSave($hydratedModel, $saveData);
+                foreach ($modelsToSave as $modelToSave) {
+                    $modelToSave->save(null, $this->pivotWidget->getSessionKey());
+                }
             }
         });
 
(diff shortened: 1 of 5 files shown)


LukeTowers:
@quendistudio this should fix #1464

PR #1506, opened 2026-08-04
Add fields property to FieldSet class

AIC-BV:
Seems to work with a small fix, even with fieldsets within fieldsets!

> Out of the box the fieldsets rendered empty — legend only, zero fields. Root cause: FieldSet::init() calls fillFromConfig(['fields']), but WidgetBase::fillFromConfig() only fills properties that exist on the class (property_exists check), and FieldSet declares public $form but not public $fields. So the inner form always got fields => null, silently. I added public $fields; to FieldSet.php:27 locally and everything came to life. Claude

image

The dark scheme might need some adjusting
```
@media (prefers-color-scheme: dark) {
    .fieldset {
        background: #000;
    }
}
```
image

### DIFF
--- a/modules/backend/formwidgets/FieldSet.php
+++ b/modules/backend/formwidgets/FieldSet.php
@@ -27,6 +27,11 @@ class FieldSet extends FormWidgetBase
     public $form;
 
     /**
+     * @var array Field configuration
+     */
+    public $fields;
+
+    /**
      * @var bool Determines if this form field should display comments and labels.
      */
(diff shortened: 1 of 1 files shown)


mjauvin:
Thanks, I merged your changes and removed the (now unused) $form property.

Can you test again to make sure?

AIC-BV:
> Thanks, I merged your changes and removed the (now unused) $form property.
> 
> Can you test again to make sure?

Works 👍

PR #1511, opened 2026-08-12
Working toward Octane support: make core request state worker-safe

austinderrick:
## Summary

Working toward Octane support. Requires wintercms/storm#241 for the primitives it calls.

Under PHP-FPM every request gets a fresh process, so state left on a long-lived object never hurts anyone. Under Laravel Octane the same process serves many requests, and anything derived from one request stays visible to the next. Most of what follows was found by running a real application on a worker and watching it misbehave, not by reading the source. Several of these leak one user's data to another.

None of this changes PHP-FPM behavior. What remains in this PR is inert without a worker: it makes core's request-derived state resettable, and nothing in core resets it.

## Split with the Winter.Octane plugin

Following the suggestion in review, the Octane-specific integration now lives in a first-party plugin: [Winter.Octane](https://github.com/austinderrick/Winter.Octane). The plugin registers Octane's service provider (Winter disables Laravel's package discovery, so an installed `laravel/octane` otherwise does nothing), clears request state at the start of every operation ahead of Octane's own listeners, calls `resetWorkerState()` on any plugin that defines it, and carries the worker test suite, which drives Octane's real `ApplicationGateway` against a single application instance. Its `composer.json` requires `laravel/octane`, so this repo no longer suggests the package.

Two consequences of plugin placement are worth weighing in review:

* The plugin registers as `elevated`, so it still loads on the privileged routes and commands where `PluginManager::$noInit` skips normal plugin initialization. Without that, the first request a worker served could decide that the reset never attaches for the worker's lifetime.
* A disabled plugin never registers. Disabling Winter.Octane while workers are still serving would silently stop the reset, so the plugin's README tells operators to switch back to PHP-FPM before disabling it. This is the one safety property that core wiring would have provided and a plugin cannot.

What stays in this PR are the changes a plugin cannot make: core classes exposing their request-derived state in resettable form, and boot-time decisions that must admit a worker.

## Manager state rebuilt per request

Five managers build their contents once and cache the result. The current user's permissions filter some of that content. A copy built for one back-end user then went to whoever came next.

* `Backend\Classes\NavigationManager` caches the main menu and side menus after permission filtering.
* `Backend\Classes\WidgetManager` caches the report widget list.
* `System\Classes\SettingsManager` caches settings items after filtering.
* `System\Classes\MailManager` caches registered mail partials and layouts.
* `Cms\Classes\ComponentManager` caches resolved component details.

Each now separates what is registered once per worker from what is derived per request, and rebuilds only the second part. Registrations survive, which matters because rebuilding them on every request would cost more than the leak.

## Trait declared caches

`Backend\Traits\PreferenceMaker` and the URL maker keep per-request caches in static properties declared on the trait rather than on the classes that use it.

PHP gives every using class its own copy of a trait's statics, so assigning through the trait clears nothing. The reset resolves the classes that use each trait and clears the property on each one. The preference cache is also keyed by user, since the same key previously served whichever user asked first.

## Boot time gates

`Backend`, `System` and `Cms` service providers registered parts of themselves only when `runningInBackend()` was true. Winter evaluates that test once per worker, at boot, when no request exists. Back-end registration therefore never ran at all under Octane.

The gates now also admit an application server, using the detection method added in wintercms/storm#241.

## CMS code cache

`Cms\Classes\CodeParser::rebuild()` wrote a generated class to a deterministic path and loaded it with `require_once`. The class name inside is random for every rebuild, and `require_once` keys on the path.

In a process serving more than one request, the path was normally included already. The load then did nothing at all and left the new class undeclared:

```
Error: Class "Cms<hash>Class" not found in CodeParser.php
```

This also disabled the recovery path in `handleCorruptCache()`, whose whole purpose is to rebuild after finding the cache unusable. It rebuilt, then could not load what it had rebuilt. Using `require` cannot redeclare anything, because the class name is unique to each rebuild.

## Twig include gate

`System\Twig\Engine` opened a gate before loading a template and closed it afterward. A throw during loading skipped the close, so the gate stayed open for every later request the worker served.

The close now happens in a `finally` block.

## Testing

Remaining in this PR:

| File | Covers |
| --- | --- |
| `modules/cms/tests/classes/CodeParserWorkerTest.php` | A rebuilt cache file loads in a reused process |
| `modules/system/tests/twig/EngineTest.php` | The include gate closes after a throw |

The worker tests moved to [Winter.Octane](https://github.com/austinderrick/Winter.Octane), where they exercise the plugin's own wiring rather than core's: request state isolation, listener ordering and priority, cross-user manager state, the reset manifest, and plugin reset discovery, all dispatched through Octane's real gateway.


LukeTowers:
@austinderrick can Octane support be provided by a first party plugin (i.e. Winter.Octane)?

austinderrick:
> @austinderrick can Octane support be provided by a first party plugin (i.e. Winter.Octane)?

Hey @LukeTowers : we could move some of this to a first party plugin, like the registerOctane/resetRequestState orchestrator, etc.

A big chunk of the changes here are more specific to the Winter internals and how they hold onto variables, so we'll probably need to keep some of them. (And the partner https://github.com/wintercms/storm/pull/241 too).

I'll draft something up for it!

austinderrick:
@LukeTowers : I made a repo over here for the plugin and invited you, so we can move it to the Winter org: https://github.com/austinderrick/Winter.Octane

Also on this PR and the related Storm one, I rewired them to work with the plugin instead.
(Both need some changes to fix state operations, but it's much more minimal with the plugin. I like this approach a lot more!)

LukeTowers:
@austinderrick if you make me an admin or an owner on the repo then I can transfer it to the Winter organization and you can continue working on it from the official namespace.

austinderrick:
> @austinderrick if you make me an admin or an owner on the repo then I can transfer it to the Winter organization and you can continue working on it from the official namespace.

@LukeTowers : Yep! Transfer headed your way.

LukeTowers:
@austinderrick repository has been transferred and added to packagist: https://packagist.org/packages/winter/wn-octane-plugin

PR #1512, opened 2026-08-14
Add PHPStan static analysis for the modules

austinderrick:
## Summary

Adds PHPStan, via larastan as storm already uses, at level 0 over the classes and models of the three modules, wired into the Code Quality workflow. This brings the modules the same discipline storm has, starting at the lowest level with a small baseline so the diff stays reviewable.

Two things make the modules analysable that a stock PHPStan setup lacks: the modules are not composer-autoloaded, so the module directories are registered as scan directories, and module classes extend runtime class aliases (`Model` and friends), so a small bootstrap file registers the alias map from `modules/system/aliases.php` before analysis.

## What the first run found

Every error was verified against the code before being fixed or baselined. The fixes:

* `ThemeExport` and `ThemeImport` build their "not intended to be saved" exception message with a broken sprintf placeholder (`"The % model"`), so the model class name never appeared in the message.
* `FilterScope` declares every configurable property except `$default`, which was created dynamically on assignment. Dynamic property creation is deprecated since PHP 8.2.
* `CmsObject::save()` swallowed the parent's bool result while documenting `@return bool`, making a successful save indistinguishable from a vetoed one for any caller of the inherited `update()`; it now returns the parent result.
* Eleven docblocks promised a return value on paths that return nothing, or the wrong type entirely. The notable ones: `UpdateManager::downloadPlugin()` and `downloadTheme()` claim to return `self` but return nothing (no caller chains on them, so this is a documentation fix rather than a behavior fix), and `CombineAssets::getDeepHashFromAssets()` claimed `void` while returning the hash string its only caller concatenates into the cache key.

## No baseline

Every error from the first run is resolved rather than ignored, so there is no baseline file. The two categories that initially looked unresolvable both had proper answers:

* `BundleManager`'s setup and scaffold handlers run bound to the invoking asset command through `Closure::call()`, which PHPStan cannot see from the registration site. `@param-closure-this` on `registerSetupHandler()` and `registerScaffoldHandler()` declares that contract on the API itself, which also gives third-party handler authors a typed `$this`.
* The `new static()` warnings are answered with `@phpstan-consistent-constructor` on the seven classes concerned, which documents the contract those factories already rely on without touching the public API the way `final` constructors would.

## Running it

```bash
composer install
vendor/bin/phpstan analyse
```

The workflow job does the same. From here the level can be raised one step at a time and the paths can widen beyond classes and models as the baseline burns down.


.github/workflows/code-quality.yaml
```diff
@@ -37,6 +37,29 @@ jobs:
       - name: Run code quality checks (on pull request)
         if: github.event_name == 'pull_request'
         run: ./.github/workflows/utilities/phpcs-pr ${{ github.base_ref }}
+  staticAnalysis:
+    runs-on: ubuntu-latest
+    name: PHPStan
+    steps:
+      - name: Checkout changes
+        uses: actions/checkout@v2
```
bennothommo: Needs up-to-date checkout version.

```suggestion
        uses: actions/checkout@v7
```

.github/workflows/code-quality.yaml
```diff
@@ -37,6 +37,29 @@ jobs:
...
+      - name: Install Composer dependencies
+        run: composer install --no-interaction --no-progress --no-scripts
+
+      - name: Reset modules
+        run: |
+          git reset --hard
+          git clean -fd
+
+      - name: Run static analysis
+        run: vendor/bin/phpstan analyse --no-progress
```
bennothommo: You'll most likely hit memory limit issues without this.

```suggestion
        run: vendor/bin/phpstan analyse --no-progress --memory-limit=2G
```

phpstan-baseline.neon
```diff
@@ -0,0 +1,73 @@
+parameters:
```
bennothommo: Can any of these errors be resolved instead of being ignored in the baseline?

phpstan-bootstrap.php
```diff
@@ -0,0 +1,37 @@
...
+                . ($parts === [] ? '' : '/') . $file;
+
+            if (is_file($path)) {
+                require_once $path;
+            }
+
+            return;
+        }
+    }
+});
```
bennothommo: This looks like an AI workaround to load the module classes. It'd be more complete if we used the default module loading functionality inside `System\ServiceProvider` -

```php
        $modules = Config::get('cms.loadModules', []);
        $classLoader = $this->app->make(ClassLoader::class);
        foreach ($modules as $module) {
            if (strtolower(trim($module)) != 'system') {
                $classLoader->autoloadPackage($module . '\\', "modules/" . strtolower($module) . '/');
            }
        }
```

You may need to adjust this by requiring the `Winter\Storm\Support\ClassLoader` class directly and feeding the `load` method of that class into `spl_autoload_register`.

bennothommo (changes requested):
OMG thank you @austinderrick - I've been wanting to do this for forever, but the last time I tried, it was a nightmare. I've suggested a couple of tweaks needed before we'd accept the changes, but great work so far!

.github/workflows/code-quality.yaml
austinderrick: Updated!

.github/workflows/code-quality.yaml
austinderrick: Updated!

phpstan-bootstrap.php
austinderrick: Updated! https://github.com/austinderrick/winter/blob/d37b34657bd22d8c5ef7d24bacd99d2cecb26c5c/phpstan-bootstrap.php

phpstan-baseline.neon
austinderrick: Good call! I pushed a bit further and this is now removed! 👍

austinderrick:
@bennothommo : Updated! :)

bennothommo:
Great stuff @austinderrick.

Just one last change - per [Winter Storm's workflow](https://github.com/wintercms/storm/blob/develop/.github/workflows/code-analysis.yaml), the name of the workflow should be "Tests" so it's grouped with the testing, and job name should be "Code Analysis". It's more of a test rather than code quality in my opinion.

austinderrick:
> Great stuff @austinderrick.
> 
> Just one last change - per [Winter Storm's workflow](https://github.com/wintercms/storm/blob/develop/.github/workflows/code-analysis.yaml), the name of the workflow should be "Tests" so it's grouped with the testing, and job name should be "Code Analysis". It's more of a test rather than code quality in my opinion.

Makes sense, @bennothommo! Updated and pushed. 💯

.github/workflows/code-analysis.yaml
```diff
@@ -0,0 +1,51 @@
+name: Tests
+
+on:
+  pull_request:
+  push:
+    branches:
+      - 1.0
+      - 1.1
+      - 1.2
```
LukeTowers: This functionality isn't present in those branches so there's no point including them.

LukeTowers (changes requested):

.github/workflows/code-analysis.yaml
austinderrick: Updated!

austinderrick:
@LukeTowers / @bennothommo : This is ready for rereview and/or merging whenever you're ready.
(Not rushing you: it just seems to be stuck in a weird "changes requested" status and wanted to make sure it didn't get forgotten there. 😄 )

PR #1518, opened 2026-08-14
Render the page after a non-AJAX postback

austinderrick:
Submitting a backend form without AJAX runs the handler and then returns a blank page. The handler result normalizes to `true`, so the dispatch chain skips the page action entirely.

The CMS controller already falls through to its page render in this case, and the backend controller now does the same. The new regression test fails without the fix, and a second test covers handlers that suppress the view.

### DIFF
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -64,6 +64,11 @@ class Controller extends ControllerBase
     public $suppressView = false;
 
     /**
+     * @var bool Whether the page action has been executed during this request.
+     */
+    protected $actionExecuted = false;
+
+    /**
      * @var array Routed parameters.
      */
     protected $params;
@@ -307,6 +312,8 @@ class Controller extends ControllerBase
          */
         $this->setNavigationContext($action, $params);
 
+        $result = null;
+
         /*
          * Execute AJAX event
          */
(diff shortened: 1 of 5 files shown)


LukeTowers:
@austinderrick can you explain the why of why you need this?

austinderrick:
Hey @LukeTowers : In another draft branch, I'm playing with PHPStan level 1 findings.

This popped up there and I noticed it was more "bug like" versus syntax, so I pulled it out.

It first surfaced there as `$result might not be defined`.
I can go back to that branch and "fix" it with a one-line `$result = null;`, if that's preferred.
That satisfies the analyzer completely, but it just hides the actual bug.

The variable is undefined because the postback branch never reaches the page action, so a bare `$result = null;` turns today's undefined-variable 500 into a silent blank page, which is arguably worse.

This PR fixes the dispatch instead. Plain handlers fall through and render the page like the CMS controller does, and the paths that can't render (widget/discovery, where the action already ran view-suppressed) throw a real error in place of the more cryptic one.

Happy to drop this and do the one-liner in the PHPStan branch if you'd rather not touch the dispatch here.

LukeTowers:
@austinderrick I still don't understand the actual real world problem that is being solved here, your description reads to me like static analysis / AI doesn't like how something is done right now and thinks it's problematic; but I still don't understand what the actual real world problem is that is trying to be fixed by this.

Could you provide an example of a scenario that doesn't work right now that this change is intended to fix? Touching stuff related to the internal routing & especially the AJAX framework routing is extremely risky; I've had to deal with a lot of security reports related to routing over the past several months.

PR #1528, opened 2026-08-24
Fieldset formwidget fields should be translatable

mjauvin:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nested form field-set initialization by removing an unnecessary nested-state setting.

### DIFF
--- a/modules/backend/formwidgets/FieldSet.php
+++ b/modules/backend/formwidgets/FieldSet.php
@@ -54,7 +54,7 @@ class FieldSet extends FormWidgetBase
         $config->alias = $this->alias . $this->defaultAlias;
         // set arrayName from parent form to save fields to the model
         $config->arrayName = $this->getParentForm()->arrayName;
-        $config->isNested = true;
+        $config->isNested = false;
 
         $widget = $this->formWidget = $this->makeWidget(Form::class, $config);
         $widget->previewMode = $this->previewMode;
--- a/modules/backend/formwidgets/FieldSet.php
+++ b/modules/backend/formwidgets/FieldSet.php
@@ -54,7 +54,6 @@ class FieldSet extends FormWidgetBase
         $config->alias = $this->alias . $this->defaultAlias;
         // set arrayName from parent form to save fields to the model
         $config->arrayName = $this->getParentForm()->arrayName;
-        $config->isNested = false;
 
         $widget = $this->formWidget = $this->makeWidget(Form::class, $config);
         $widget->previewMode = $this->previewMode;


AIC-BV (approved):

LukeTowers (changes requested):
@mjauvin this is going to cause issues with code that extends form widgets that checks for isNested to avoid injecting fields on the wrong Form widget instance.

AIC-BV:
Sorry for the AI dump
A possible fix in the code suggestion box below

@LukeTowers you're right — EasyAudit, Winter.Pages and Winter.TailwindUI all guard on `isNested` together with a `model instanceof` / controller check, and a fieldset's inner form shares both the parent's model *and* controller, so dropping the flag has them inject a duplicate Activity Log tab / Preview tab / brand fields inside the fieldset.

The root issue is that `isNested` carries two meanings:

1. "this Form widget is a child of another Form widget" — what extension guards use it for;
2. "this form's data scope differs from the model's attributes" — what the docblock describes ("a good indicator to expect that the form model and dataset values will differ"), and what core and Winter.Translate use it for.

For `repeater` and `nestedform` both hold. For `fieldset` only (1) does: it reuses the parent's model *and* the parent's `arrayName`, and merges its `getSaveData()` straight back up, so its fields are model attributes exactly as if they had been declared on the parent form.

Conflating them breaks two things:

- **Winter.Translate skips the form** (this PR's motivation) — translatable attributes inside a fieldset never get their `ml*` widget.
- **Core's own required-detection is wrong.** `Form::defineFormField()` prefixes the attribute name for nested forms; with the fieldset's inherited `arrayName` that yields `isAttributeRequired('.website_name')` for a parent `arrayName` of `Settings` — a name that can never match a rule — so `required` is silently never auto-detected for fieldset fields.

### Suggestion

Keep `isNested = true` so every existing guard behaves exactly as it does today, and add a narrower flag for the second meaning:

```diff
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -74,6 +74,15 @@ class Form extends WidgetBase
      */
     public $isNested = false;
 
+    /**
+     * @var bool Used to flag that this nested form shares the data scope of its parent
+     * form; ie. its fields resolve against the same model attributes and array name as
+     * if they had been defined on the parent form directly. Only relevant when
+     * $isNested is true. Used by the fieldset form widget, which nests fields visually
+     * without introducing a new data scope.
+     */
+    public $sharesParentScope = false;
+
     //
     // Object properties
     //
@@ -138,6 +147,7 @@ class Form extends WidgetBase
             'context',
             'arrayName',
             'isNested',
+            'sharesParentScope',
         ]);
 
@@ -881,8 +891,9 @@ class Form extends WidgetBase
         if ($field->required === null && $this->model && method_exists($this->model, 'isAttributeRequired')) {
-            // Check nested fields
-            if ($this->isNested) {
+            // Check nested fields, unless the nested form shares its parent's data
+            // scope, in which case the attribute name needs no prefixing
+            if ($this->isNested && !$this->sharesParentScope) {
                 // Get the current attribute level
                 $nameArray = HtmlHelper::nameToArray($this->arrayName);
                 unset($nameArray[0]);
--- a/modules/backend/formwidgets/FieldSet.php
+++ b/modules/backend/formwidgets/FieldSet.php
@@ -54,6 +54,10 @@ class FieldSet extends FormWidgetBase
         // set arrayName from parent form to save fields to the model
         $config->arrayName = $this->getParentForm()->arrayName;
+        // the fields are nested visually only; they resolve against the parent
+        // form's model and array name, exactly as if defined on the parent form
+        $config->isNested = true;
+        $config->sharesParentScope = true;
```

Winter.Translate's guard then becomes:

```php
// EventRegistry::registerModelTranslation()
$sharesParentScope = property_exists($widget, 'sharesParentScope') && $widget->sharesParentScope;

if (!$model->hasTranslatableAttributes() || ($widget->isNested && !$sharesParentScope)) {
    return;
}
```

`property_exists()` keeps that working against older core releases.

### Verified

On a settings model with 22 translatable attributes, all of them inside `type: fieldset` groups:

- with the change: all 22 render as `ml*` widgets inside the fieldset;
- with Translate's original `$widget->isNested` guard restored: 0 — confirming the inner form still reports `isNested === true`, so third-party guards are untouched.

Happy to open the Winter.Translate PR (and a docs note) alongside if you like the direction. Naming is of course open — `isInline`, `inheritsParentScope`, whatever reads best.

### Alternative

If you'd rather not add core surface at all: have Winter.Translate recurse into `type: fieldset` field definitions when it processes the parent form's fields. That fixes translation only, and leaves the required-detection bug — and any other consumer reading `isNested` as "different data scope" — in place.

LukeTowers:
@AIC-BV seems reasonable enough, feel free to submit a PR.

